PR #33, opened 2018-03-20
Limit max db connections

jmattheis:
While load testing mysql compained about too many connections.


jmattheis:
Manually merged d5d19b55b

Issue #71, opened 2018-11-12
Accept POST messages without title


JohnnyLeone:
I'd like to use the requests for a application without always specifying a title for it. The title should not a mandatory field and if no title got specified we should always pick the application name for it. This also reduces the minimum size of a API request.

madding:
@jmattheis I make pull request in my fork, because I don't have permission. Please, try to check it https://github.com/gotify/server/pull/72

jmattheis:
Done in #72

PR #72, opened 2018-11-12
[#71] Make title in message not necessary field

madding:


Gopkg.lock
```diff
@@ -2,201 +2,314 @@
 
 
 [[projects]]
+  digest = "1:5f8ec75d953827c5f7b5faf8bffd5573dfd65dd181f07e769c6681036fb5db75"
```
jmattheis: As this change has nothing todo with dependencies I would prefer if you would remove the gopkg.lock & package-lock.json changes from the PR.

docs/spec.json
```diff
@@ -1523,4 +1522,4 @@
       "in": "query"
     }
   }
-}
\ No newline at end of file
+}
```
jmattheis: The file must exactly match the file created by `make update-swagger` otherwise the build fails. The generated file does not contain a newline at the end.

model/message.go
```diff
@@ -27,9 +27,8 @@ type Message struct {
 	Message string `form:"message" query:"message" json:"message" binding:"required"`
 	// The title of the message.
 	//
-	// required: true
 	// example: Backup
-	Title string `form:"title" query:"title" json:"title" binding:"required"`
```
jmattheis: The application name should be used as default value otherwise the title just would be empty.

madding:
@jmattheis I fixed all notes by review

api/message.go
```diff
@@ -129,7 +129,12 @@ func (a *MessageAPI) DeleteMessage(ctx *gin.Context) {
 func (a *MessageAPI) CreateMessage(ctx *gin.Context) {
 	message := model.Message{}
 	if err := ctx.Bind(&message); err == nil {
-		message.ApplicationID = a.DB.GetApplicationByToken(auth.GetTokenID(ctx)).ID
+		application := a.DB.GetApplicationByToken(auth.GetTokenID(ctx))
+		message.ApplicationID = application.ID
+		// if title is empty set application.name as default value
+		if len(message.Title) == 0 {
```
jmattheis: Just comparing the strings looks a little cleaner IMO.
```suggestion
		if message.Title == "" {
```

api/message.go
```diff
@@ -129,7 +129,12 @@ func (a *MessageAPI) DeleteMessage(ctx *gin.Context) {
 func (a *MessageAPI) CreateMessage(ctx *gin.Context) {
 	message := model.Message{}
 	if err := ctx.Bind(&message); err == nil {
-		message.ApplicationID = a.DB.GetApplicationByToken(auth.GetTokenID(ctx)).ID
+		application := a.DB.GetApplicationByToken(auth.GetTokenID(ctx))
+		message.ApplicationID = application.ID
+		// if title is empty set application.name as default value
```
jmattheis: This comment doesn't really add something to the code, the code speaks for it self, so please remove it.

JohnnyLeone:
As far as i checked the tests. There are more changes needed for the tests to work properly. For example: `Test_CreateMessage_onlyRequired` should also check for the correct title if no title got specified. I think the current tests look good, but they need to be modified for that new behavior.

api/message.go
madding: May be here need to check strings with whitespaces ? Like this:
```strings.TrimSpace(message.Title) == ""```

Comment (jmattheis) [reply api/message.go]: Yeah good idea.

**Comment by madding** [comment]
@jmattheis I fixed tests and done all changes

File: api/message_test.go [hunk 6]
```diff
@@ -366,18 +366,36 @@ func (s *MessageSuite) Test_CreateMessage_failWhenNoMessage() {
 	assert.False(s.T(), s.notified)
 }
 
-func (s *MessageSuite) Test_CreateMessage_failWhenNoTitle() {
+func (s *MessageSuite) Test_CreateMessage_WhenNoTitle() {
```
jmattheis: WhenNoTitle doesn't really make sence
```suggestion
func (s *MessageSuite) Test_CreateMessage_WithoutTitle() {
```

api/message_test.go
```diff
@@ -366,18 +366,36 @@ func (s *MessageSuite) Test_CreateMessage_failWhenNoMessage() {
...
-	assert.Equal(s.T(), 400, s.recorder.Code)
-	assert.False(s.T(), s.notified)
+	msgs := s.db.GetMessagesByApplication(8)
+	assert.Len(s.T(), msgs, 1)
+	assert.Equal(s.T(), "Application name", msgs[0].Title)
+	assert.Equal(s.T(), 200, s.recorder.Code)
+	assert.True(s.T(), s.notified)
+}
+
+func (s *MessageSuite) Test_CreateMessage_WhenBlankTitle() {
```
jmattheis:
```suggestion
func (s *MessageSuite) Test_CreateMessage_WithBlankTitle() {
```

api/message_test.go
```diff
@@ -366,18 +366,36 @@ func (s *MessageSuite) Test_CreateMessage_failWhenNoMessage() {
 	assert.False(s.T(), s.notified)
```
jmattheis: The method `Test_CreateMessage_onlyRequired` in this file should be renamed to `Test_CreateMessage_WithTitle`

madding:
I made fixes and squash all into one commit

PR #73, opened 2018-11-12
Do not require a message title

JohnnyLeone:
This commit makes the mandatory title field optional. If a message is
sent without a title, the application name will be used instead.

I've also started editing the tests for the necessary changes. I think those changes can also be part of this.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -130,6 +130,10 @@ func (a *MessageAPI) CreateMessage(ctx *gin.Context) {
 	message := model.Message{}
 	if err := ctx.Bind(&message); err == nil {
 		message.ApplicationID = a.DB.GetApplicationByToken(auth.GetTokenID(ctx)).ID
+		// If no title was specified always use the application name
+		if len(message.Title) == 0 {
+			message.Title = a.DB.GetApplicationByToken(auth.GetTokenID(ctx)).Name
+		}
 		message.Date = timeNow()
 		a.DB.CreateMessage(&message)
 		a.Notifier.Notify(auth.GetUserID(ctx), &message)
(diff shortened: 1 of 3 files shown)


api/message.go
```diff
@@ -130,6 +130,10 @@ func (a *MessageAPI) CreateMessage(ctx *gin.Context) {
 	message := model.Message{}
 	if err := ctx.Bind(&message); err == nil {
 		message.ApplicationID = a.DB.GetApplicationByToken(auth.GetTokenID(ctx)).ID
```
jmattheis: It would be better if `a.DB.GetApplicationByToken` would be executed only once here, because it makes an call to the database.

api/message.go
```diff
@@ -130,6 +130,10 @@ func (a *MessageAPI) CreateMessage(ctx *gin.Context) {
...
+		// If no title was specified always use the application name
+		if len(message.Title) == 0 {
```
jmattheis: Just comparing the strings looks a little cleaner IMO.
```suggestion
		if message.Title == "" {
```

docs/spec.json
```diff
@@ -1523,4 +1522,4 @@
       "in": "query"
     }
   }
-}
\ No newline at end of file
+}
```
jmattheis: The file must exactly match the file created by make update-swagger otherwise the build fails. The generated file does not contain a newline at the end.

JohnnyLeone:
Sorry i didn't checked my notifications as i created my PR. I'm totally fine if we do the remaining tasks in #72. As far as i checked the changes over there, all changes from here are already part in that PR. 
If it is OK, we can close this one here.

jmattheis:
@JohnnyLeone Okay good, still thanks for your contribution (:

PR #83, opened 2018-11-20
Add API end-point for edit application name and description

madding:


madding:
@jmattheis I fixed all notes.

jmattheis:
Thanks!

Issue #84, opened 2018-11-21
Missing 400 BadRequest response in swagger definition


jmattheis:
All endpoints who create or update something should define a 400 bad request response.

jmattheis:
Done in #86

PR #85, opened 2018-11-22
Update swagger doc. Add 400 BadRequest response in create and update endpoints

daniruizcamacho:
This code is related to this open issue https://github.com/gotify/server/issues/84

I don't see any update endpoint in the documentation and I only need to add the bad request error response code in creation endpoints.


jmattheis:
Fixed in #86

Issue #23, opened 2018-03-04
Make response headers configurable


jmattheis:
Could be used for setting an `Access-Control-Allow-Origin`-header.

Nicolai-:
Any news here? :)

We need this for several different sub-domains, and want just one single gotify instance for all of them - like
- gotify.example.com
- sub1.example.com
- sub2.example.com
- etc

This is currently only available if we run in dev-mode - but... ;)

jmattheis:
@Nicolai- probably this week or next one, why do you need this feature for having different subdomains? You could use a nginx proxy to handle more than one domain.

fMads:
@jmattheis I work with @Nicolai- 
We currently have many subdomains (each of our customers have their own subdomain), and would like all of them to talk to the same gotify instance (instead of having one instance per subdomain).

When testing now, and trying to connect from sub.example.com to gotify.example.com, we get a 403 Access Denied/Not authorized.
Our best guess is, that it's because of CORS issues (the issue linked in this one #67).

jmattheis:
@fMads @Nicolai- What exactly do you mean with 
> trying to connect from sub.example.com to gotify.example.com

Are all (sub) domains are hosted on one server? Do you using the Gotify UI or do you manually call the API of Gotify? 

When you receive a 403 unauthorize and its logged in Gotify then its likely not an issue with missing CORS headers. When the CORS headers are missing there should be only an "OPTIONS" request logged in Gotify and the browser console should show something like this (from chrome):
> Access to XMLHttpRequest at 'http://localhost/version' from origin 'http://localhost:3001' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

fMads:
@jmattheis The subdomains are spread over multiple servers - and all subdomains are running in their own docker-container - gotify is also running in it's own container.

Hmm, I think you're right about the CORS.
What we tried were
- Open Gotify UI
- Break in javascript where it creates the websocet connection (MessageAction.ts:76)
- Copy the url (including the token)
- Run in Chrome Devtools on one of the other domains

The error we get are: "WebSocket connection to 'wss://gotify.example.com/stream?token=CQ.bCn8b6Y.2ZYh' failed: Error during WebSocket handshake: Unexpected response code: 403"

jmattheis:
[...]

PR #94, opened 2018-12-07
#23 fix check same origins

madding:


madding:
@jmattheis I fixed all your comments.

madding:
need to rerun tests @jmattheis

jmattheis:
@madding could you squash your commits?

madding:
yep

Issue #65, opened 2018-11-08
"url" message parameter


Red5d:
It would be useful to have an optional "url" parameter (and maybe a "urltitle" parameter too for customizing the display text for it) when posting a message to display a clickable link to a URL in the notification message.

madding:
@jmattheis Can I add this npm package on client side ?

https://www.npmjs.com/package/markdown

jmattheis:
Closed in favor of #129

PR #100, opened 2018-12-24
[#65] add pathOnClick to message struct, add redirect after notification click

madding:


madding:
@jmattheis I added new field PathOnClick to message on API server. When user click on notifcation on client side if PathOnClick exist user will be redirect to specified path.

jmattheis:
Hey, thanks for your contribution!

I'm still unsure how the feature should be implemented, maybe it should be just link parsing on client side.

madding:
@jmattheis 
I see two solutions for message view: we send html from API or we use markdown library on client side. Both solutions doesn't work for browser notification. My solution works only for browser notification.

jmattheis:
@madding True, but the 'goto path on notification click' functionality wasn't requested anywhere. #65 is only about a clickable link inside the message.

madding:
@jmattheis 
I squashed commits and updated branch, now you can merge or close this pull.

jmattheis:
@madding I'll close this PR for now and will reopen it when it is clear that this functionality is needed.

PR #110, opened 2019-01-09
Plugin Support

### LINKED ISSUE #51: Pluggable extensions
It would be nice to have pluggable extensions for Gotify. This could f.ex. be used for polling the GitHub Notification API.

* [Golang Plugin API](https://golang.org/pkg/plugin/) or some kind of JavaScript engine to enable pluggable code
* *?* Allow adding routes something like `/plugin/:name/...` *(overkill ?!?)*
  * Could be useful for webhooks
* *?* Users can enable plugins
* *?* Key-Value per user for configuration (f.ex authentication token)

### LINKED ISSUE #69: Allow updating app name and description through API/UI
I could not find a way to update an application's name and description through the UI or the API. Since its possible to change the application image after creating an application, I expected to be able to change other application properties as well.

eternal-flame-AD:
This fixes #51 
This fixes #69

I will implement the proposal we discussed in #51 progressively. Here's a checklist:

- [x] Backend Database Storage for Plugins
- [x] Global Configuration Item
- [x] Plugin base interface
- [x] Plugin loader infrastructure && base plugin APIs
- [x] Webhooker
- [x] Configor
- [x] MessageSender
- [x] WebUI

Before merge:
- [ ] Documentation [gotify/website](https://github.com/gotify/website) -> https://github.com/gotify/website/pull/2


database/database.go
```diff
@@ -35,12 +35,11 @@ func New(dialect, connection, defaultUser, defaultPass string, strength int, cre
 		db.DB().SetMaxOpenConns(1)
 	}
 
-	if !db.HasTable(new(model.User)) && !db.HasTable(new(model.Message)) &&
-		!db.HasTable(new(model.Client)) && !db.HasTable(new(model.Application)) {
-		db.AutoMigrate(new(model.User), new(model.Application), new(model.Message), new(model.Client))
-		if createDefaultUser {
-			db.Create(&model.User{Name: defaultUser, Pass: password.CreatePassword(defaultPass, strength), Admin: true})
-		}
+	db.AutoMigrate(new(model.User), new(model.Application), new(model.Message), new(model.Client), new(model.PluginConf))
```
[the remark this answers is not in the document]
jmattheis: Alright, yeah that was also on my todo list, back when I created the function I kinda missunderstood what AutoMigrate actually did.

database/migration_test.go
```diff
@@ -0,0 +1,64 @@
+package database
+
+import (
+	"os"
+	"testing"
+
+	"github.com/gotify/server/model"
```
jmattheis: The imports from github.com should be grouped without empty lines between them.
```
import (
	"os"
	"testing"

	"github.com/gotify/server/model"
	"github.com/jinzhu/gorm"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/suite"
)
```

database/plugin.go
```diff
@@ -0,0 +1,36 @@
+package database
+
+import (
+	"github.com/gotify/server/model"
+)
+
+// GetPluginConfByUserID gets plugin configurations from a user
+func (d *GormDatabase) GetPluginConfByUserID(userid uint) []*model.PluginConf {
```
jmattheis: To stay consistent with the other database methods, name the methods like this:
```
func (d *GormDatabase) GetPluginConfByUser(userID uint) []*model.PluginConf {
func (d *GormDatabase) GetPluginConfByUserAndPath(userID uint, path string) *model.PluginConf {
func (d *GormDatabase) GetPluginConfByUserAndPath(userID uint, path string) *model.PluginConf {
```

model/pluginconf.go
```diff
@@ -0,0 +1,38 @@
...
+// Holds information about a plugin instance for one user.
+//
+// swagger:model PluginConf
+type PluginConfExternal struct {
+	// The user id.
+	//
+	// read only: true
+	// required: true
+	// example: 25
+	UserID uint `json:"user_id"`
```
jmattheis: The external representation of the plugin config doesn't need the user id. Each user can only see his own configs so the userID will always be the same.

model/pluginconf.go
```diff
@@ -0,0 +1,38 @@
...
+	// The user name. For login.
+	//
+	// required: true
+	// example: P1234
+	Token string `binding:"required" json:"token" query:"token" form:"token"`
+	// The file name of the plugin.
+	//
+	// example: rss.so
+	FileName string `json:"file_name" form:"admin" query:"admin"`
```
jmattheis: camelCase is used throughout Gotify so lets not introduce snake_case (:. (Also form and query has weird values)
```suggestion
	FileName string `json:"fileName" form:"fileName" query:"fileName"`
```

model/pluginconf.go
```diff
@@ -0,0 +1,38 @@
...
+	// required: true
+	// example: 25
+	UserID uint `json:"user_id"`
+	// The user name. For login.
+	//
+	// required: true
+	// example: P1234
+	Token string `binding:"required" json:"token" query:"token" form:"token"`
+	// The file name of the plugin.
+	//
```
jmattheis: Should the filename be editable, I see no usage why a user should be able to do this.
So add this here:
```
// read only: true
// required: true
```

model/pluginconf.go
```diff
@@ -0,0 +1,34 @@
...
+	// example: rss.so
+	// read only: true
+	// required: true
+	FileName string `json:"fileName" form:"fileName" query:"fileName"`
+	// Whether the plugin instance is enabled.
+	//
+	// example: true
+	Enabled bool `json:"enabled"`
```
jmattheis: This should also be required as it is never unset because it is a primitive.

eternal-flame-AD: Yeah right, thanks for your reminder. I'll read more about swagger docs about these. :)

eternal-flame-AD:
The WebUI part is not completed yet, we can review on the go part first

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+// consumes: [application/json]
+// produces: [application/json]
+// security: [pluginTokenHeader: [], pluginTokenQuery: []]
+// responses:
+//   200:
+//     description: Ok
+//     schema:
+//       type: array
+//       items:
+//         $ref: "#/definitions/Application"
```
jmattheis: Disable doesn't return an application

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+		app := c.db.GetApplicationByToken(c.conf.ApplicationToken)
+		if app != nil {
+			// managed application exist
+			c.managedApp = app
+			return
+		}
+	}
+
+	app := &model.Application{
+		Token:       auth.GenerateApplicationToken(),
```
jmattheis: Add a flag `Internal`-Flag to the Application if true the application should not be deletable or editable by the user. 

The internal application should not need an token as the app should only be used internally.

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+		Title:         msg.Title,
+		Priority:      msg.Priority,
+		Date:          timeNow(),
+	}
+	c.db.CreateMessage(&message)
+	c.notifier.Notify(c.conf.UserID, &message)
+	return nil
+}
+
+func (c *PluginAPI) performPluginInstanceInitialization(instance loader.Instance, conf model.PluginConf) error {
```
jmattheis: Can we extract methods like initPluginForAllUsers, DestroyPluginsForUser, enable/disable to a different struct (inside the plugin package) something like plugin.Manager. I'd like that the api file only contains the swagger docuementation and params parsing (and simple implementations of routes)

api/plugin.go
eternal-flame-AD: The app should be editable(eg: for renaming, uploading images, etc.), but not deletable.

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+					capabilities = append(capabilities, "v1.webhooker")
+				}
+				if _, ok := inst.(papiv1.Storager); ok {
+					capabilities = append(capabilities, "v1.storager")
+				}
+			}
+			result = append(result, model.PluginConfExternal{
+				ID:           conf.ID,
+				Name:         name,
+				Token:        conf.Token,
```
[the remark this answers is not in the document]
eternal-flame-AD: To calculate the custom route endpoint.

router/router.go
```diff
@@ -33,13 +41,59 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
...
+	userChangeNotifier.OnUserAdded(pluginHandler.InitPluginsForUser)
 
 	g := gin.New()
 
 	g.Use(gin.Logger(), gin.Recovery(), error.Handler(), location.Default())
 	g.NoRoute(error.NotFound())
 
+	{
+		g.GET("/plugin", authentication.RequireClient(), pluginHandler.GetPlugins)
+		pluginRoute := g.Group("/plugin/", authentication.RequirePlugin())
```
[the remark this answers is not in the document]
eternal-flame-AD: Yeah, the endpoint design might need some changes. Actually, I hit a problem while designing custom routers for plugins. Since you need to provide a `RouterGroup` to the plugin, it is inevitable to register specific routes such as `/plugin/Pabcd/custom/`. However I am not sure whether it is a [limitation](https://github.com/gin-gonic/gin/issues/1301) or the `gin` team [disproves](https://github.com/gin-gonic/gin/issues/1681) it explicitly, a wildcard route and a specific route cannot co-exist even though they don't actually overlap(eg: `/plugin/:id/enable` and `/plugin/Pabcd/custom/`).

router/router.go
```diff
@@ -33,13 +41,59 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
...
+			pluginRoute.POST("/config", pluginHandler.UpdateConfig)
+			pluginRoute.GET("/config/schema", pluginHandler.GetConfigSchema)
+			pluginRoute.GET("/display", pluginHandler.GetDisplay)
+			pluginRoute.POST("/enable", pluginHandler.EnablePlugin)
+			pluginRoute.POST("/disable", pluginHandler.DisablePlugin)
+		}
+		pluginHandler.CustomRouteMux = g.Group("/plugin/custom/")
+	}
+
+	if pluginsDir := conf.PluginsDir; pluginsDir != "" {
```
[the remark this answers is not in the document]
eternal-flame-AD: I will resolve this in PluginManager

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+	}
+	instance, err := c.Holder.GetInstance(token)
+	if err != nil {
+		ctx.AbortWithError(500, errors.New("plugin instance not found"))
+	}
+	switch instance := instance.(type) {
+	case papiv1.Displayer:
+		ctx.JSON(200, instance.GetDisplay())
+	default:
+		ctx.JSON(501, errors.New("plugin does not implement Displayer"))
```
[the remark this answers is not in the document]
eternal-flame-AD: I will resolve this in PluginManager

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+	"github.com/gotify/server/auth"
+	"github.com/gotify/server/model"
+	"github.com/gotify/server/plugins/holder"
+	"github.com/gotify/server/plugins/loader"
+
+	papiv1 "github.com/gotify/plugin-api"
+)
+
+// The PluginDatabase interface for encapsulating database access.
+type PluginDatabase interface {
```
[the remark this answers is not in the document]
eternal-flame-AD: Guess it's the same proposal as you mentioned later as `plugin.Manager`?

plugins/loader/interface.go
```diff
@@ -0,0 +1,36 @@
+package loader
+
+// Instance is an abstraction of plugin instance
+type Instance interface{}
+
+// P is an abstraction of plugin handler
+type P interface {
```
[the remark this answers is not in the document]
eternal-flame-AD: I will provide a unified abstraction layer at `plugin.Manager`

api/plugin.go
jmattheis: right.

api/plugin.go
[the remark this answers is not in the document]
jmattheis: Do we need that? the plugins will probably add an own path to it, so it will be different for every plugin?

api/plugin.go
```diff
@@ -0,0 +1,602 @@
...
+	plugins := c.DB.GetPluginConfByUser(userID)
+	result := make([]model.PluginConfExternal, 0)
+	for _, conf := range plugins {
+		if inst, err := c.Holder.GetInstance(conf.Token); err == nil {
+			p, err := c.Holder.PluginHolder.Get(conf.ModulePath)
+			name := "[UNNAMED]"
+			if err == nil && p.GetPluginInfo().Name != "" {
+				name = p.GetPluginInfo().Name
+			}
+			capabilities := make([]string, 0)
```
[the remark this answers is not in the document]
jmattheis: Haven't thought about that, so keep it.

api/plugin.go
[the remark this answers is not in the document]
jmattheis: I'd still prefer using the id as identifier in the url and then also have the PluginID in the url for the custom route. This makes it more conistent with the other apis as they all use integer ids.

> user need to know two pieces of information to calculate the endpoint

I guess not? The plugin can set custom paths so it is not really useful to do this. The plugin should output the full path via the Displayer interface. Only the plugin will know on which path it has registered the webhook url.

> two authentication methods would coexist in /plugin/:id/*

I don't see this as a problem. Things like this do also already exist `POST /message` requires an authentication token and `GET /message` a client token. 
The custom plugin path could also be served on an other path something like `domain.com/custom/plugin/:plugintoken` but I'd prefer not to because it belongs to `/plugin`.

api/plugin.go
eternal-flame-AD: I have moved the actual initialization code to the loader, but I think it is necessary to leave some methods there(the biggest problem is creating custom webhooks). Moving all initialization code to loader would require routes to be registered in loader, which is also not elegant in my opinion.

api/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+		err := c.setPluginEnableStatus(id, false)
+		if err != nil {
+			ctx.AbortWithError(500, err)
+		}
+	})
+}
+
+func (c *PluginAPI) setPluginEnableStatus(id uint, enabled bool) error {
+	c.mutex.Lock()
```
jmattheis: why do we need a mutex here?

api/plugin.go
```diff
@@ -0,0 +1,474 @@
...
+			return
+		}
+		err := c.setPluginEnableStatus(id, false)
+		if err != nil {
+			ctx.AbortWithError(500, err)
+		}
+	})
+}
+
+func (c *PluginAPI) setPluginEnableStatus(id uint, enabled bool) error {
```
jmattheis:
```go
func (c *PluginAPI) setPluginEnableStatus(id uint, enabled bool) error {
	c.mutex.Lock()
	defer c.mutex.Unlock()
	instance, err := c.Holder.GetInstance(id)
	if err != nil {
		return errors.New("instance not found")
	}
	conf := c.DB.GetPluginConfByID(id)

	if conf.Enabled == enabled {
		return errors.New("config is already enabled/disabled")
	}
	
	if enabled {
		err = instance.Enable()
	} else {
		err = instance.Disable()
	}
	if err != nil {
		return err
	}
	return c.DB.UpdatePluginConf(conf)
}
```

auth/authentication.go
```diff
@@ -41,6 +42,20 @@ func (a *Auth) RequireAdmin() gin.HandlerFunc {
 	})
 }
 
+// RequirePlugin returns a gin middleware which requires a plugin token or basic authentication header to be supplied
+// with the request.
+func (a *Auth) RequirePlugin() gin.HandlerFunc {
```
jmattheis: We do not need this anymore right?

jmattheis:
@eternal-flame-AD I could continue working on this PR if you want/dont have time for it.

eternal-flame-AD:
It's Chinese Spring Festival so actually I am on holiday^_^ That's OK if you want to finish it:)

api/plugin.go
eternal-flame-AD: To prevent disable and enable being called simultaneously creating an inconsistency with the enabled state of the plugin instance and the record in the database

plugins/wrapper/wrapper.go
```diff
@@ -0,0 +1,73 @@
+package wrapper
+
+import (
+	"github.com/gin-gonic/gin"
+)
+
+// PluginInstance is a wrapper to a real underlying plugin instance.
+type PluginInstance interface {
```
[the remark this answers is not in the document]
eternal-flame-AD: Then I think there would be no way to determine whether a plugin has implemented a feature trait without using an error? Since we have an API that only detects the existence of plugin feature traits but does not execute them.

eternal-flame-AD:
Thank you very much for your patience and wonderful reviewing^_^. I will come back to normal productivity shortly.

api/plugin.go
eternal-flame-AD: I will have the API handler modified together with the WebUI change.

plugins/wrapper/wrapper.go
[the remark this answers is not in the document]
jmattheis: The PluginInstance could have an method `Capabilities` which returns a string array.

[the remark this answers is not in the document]
eternal-flame-AD: I'm sorry but I don't see a benefit in this. 
I guess you proposed this in order to simplify the process to get rid of the rather complicated wrapping logic in `github.com/gotify/server/plugin/loader`, but some signatures in the plugin API require a dedicated transformation, such as the interface for `Messenger` requires a `interface { SetMessageHandler(h papiv1.MessageHandler) }` and papiv1.MessageHandler is `interface { SendMessage(m Message) error }`, which requires an intermediate struct for transforming this API to the wrapper API.
Plus, I really love to use the wonderful type system in go to ensure the availability of methods instead of relying on an error or `strings.Contains`.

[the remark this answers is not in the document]
jmattheis: oke

eternal-flame-AD:
Up till now, I think we have a functional APi and WebUI. There are several things on the TODO list:
[ ] Tests
[ ] More friendly WebUI
[ ] config schema verification

api/plugin.go
jmattheis: I dunno too, I'll try to implement what I've on my mind on a different branch. (probably this weekend)

router/router.go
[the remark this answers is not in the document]
jmattheis: This is resolved too right?

jmattheis:
Do you know why travis ci isn't building this PR?

From what I see only https://github.com/gotify/server/pull/110#discussion_r251137577 is left (while doing it, I'll surely have some remarks again as the PR is pretty big).

eternal-flame-AD:
I have no idea either:(
Seems that one of the push failed to trigger a build at Travis while GitHub was waiting forever for the results on that push.

api/plugin.go
eternal-flame-AD: Alright, so I'll wait for you then.

api/plugin.go
```diff
@@ -0,0 +1,470 @@
...
+		if con := instance.GetConfigurer(); con != nil {
+			newConf := con.EmptyConfig()
+			conf, err := ioutil.ReadAll(ctx.Request.Body)
+			if err != nil {
+				ctx.AbortWithError(500, err)
+			}
+			if err := yaml.Unmarshal(conf, newConf); err != nil {
+				ctx.AbortWithError(400, err)
+			}
+			if err := con.ValidateAndSetConfig(newConf); err != nil {
```
jmattheis: When this is successful, the config should be saved inside the pluginConf. 

Also, it would be good to also include this functionality to the plugin.manager (then all access of the plugin conf would be encapsulated by the plugin.Manager).
```go
func (m *Manager) SaveConfig(pluginID uint, config string) (statusCode int, err error)
```

config/config.go
```diff
@@ -39,6 +39,7 @@ type Configuration struct {
 	}
 	PassStrength      int    `default:"10"`
 	UploadedImagesDir string `default:"data/images"`
+	PluginsDir        string `default:""`
```
jmattheis: Could you change the default to `data/plugins`

router/router.go
```diff
@@ -60,7 +71,36 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
...
+		pluginHandler.LoadPlugins(pluginsDir)
+	}
+	if err := pluginHandler.AddAllUsers(); err != nil {
+		panic(err)
+	}
+
+	if mode.IsDev() {
+		g.OPTIONS("/*any", func(ctx *gin.Context) {
```
jmattheis: This was already done in line 63, why is it needed?

eternal-flame-AD: The default handler to OPTIONS does not allow all HTTP Methods used in the API

config/config.go
eternal-flame-AD: Okay, I set it to empty at first because I want plugin functionality to be disabled by default, now it seems that an empty directory works as well.

api/plugin.go
eternal-flame-AD: The DB update has been added in https://github.com/gotify/server/pull/110/commits/f86e57b316c1e18a116d1d7f1c2dd3544b22e675

I don't really like the idea of handling the unmarshaling work to the manager, it does not make things less coupled because the returned `statusCode` is another form of coupling.

plugin/manager.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	return nil
+}
+
+func (m *Manager) loadPlugins(directory string) error {
+	if directory == "" {
+		return nil
+	}
+
+	if err := os.MkdirAll(directory, 0777); err != nil {
```
eternal-flame-AD: Firstly, just don't create directories with 0777 permission. Use 0755

Secondly, maybe consider moving the directory creation proc to `app.go`, because the image dir is created there(also use 0644 or 0666 for the image dir), or move them both to an appropriate location.

api/plugin.go
```diff
@@ -0,0 +1,355 @@
...
+		conf.Config = newconfBytes
+		c.DB.UpdatePluginConf(conf)
+	})
+}
+
+func supportOrAbort(ctx *gin.Context, instance compat.PluginInstance, module compat.Module) (aborted bool) {
+	if compat.HasSupport(instance, module) {
+		return false
+	}
+	ctx.AbortWithError(400, fmt.Errorf("plugin does not support %s", module))
```
eternal-flame-AD: Consider use a `501 Not Implemented` here

api/user.go
```diff
@@ -19,11 +19,45 @@ type UserDatabase interface {
 	CreateUser(user *model.User) error
 }
 
+// UserChangeNotifier notifies listeners for user changes.
+type UserChangeNotifier struct {
+	userDeletedCallbacks []func(uid uint) error
```
eternal-flame-AD: I doubt whether we need an `error` returned here.

An `error` is designed to propagate possibly recoverable errors to the upstream caller. In this case, the use of this callback is just to deliver a message that user info has changed, and the upstream caller can do nothing different on whether the callback returned a successful result, and the error is never handled in your implementation as well. I also cannot foresee a case that this error could be utilized in the future.

eternal-flame-AD:
I'm sorry but I am running out of ideas on how to integrate [yaml-plugin](https://github.com/pengx17/monaco-yaml) into the monaco editor. The documentation did not mention a way to integrate it in an environment with no control over `webpack.config.js`(`react-scripts-ts`). Do you have any idea on how to implement this? Thanks

plugin/compat/instance.go
```diff
@@ -0,0 +1,88 @@
+package compat
+
+import "github.com/gin-gonic/gin"
+
+// Module is a capability the plugin provides
+type Module string
```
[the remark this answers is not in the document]
jmattheis: Oke, lets name it `Capability`.

jmattheis:
I'll try out the yaml plugin later this week.

api/plugin.go
jmattheis: We provide a capabilities api, in there it is specified which apis are supported for a given plugin. Therefore I chose 400 because the client made a mistake.

Also, the rfc states that 'not implemented' means that it is unsupported for any resource (in this case plugins) and this is not true:
> The 501 (Not Implemented) status code indicates that the server does
   not support the functionality required to fulfill the request.  This
   is the appropriate response when the server does not recognize the
   request method and is not capable of supporting it for any resource.

https://tools.ietf.org/html/rfc7231#section-6.6.2

plugin/manager.go
jmattheis: Yeah, true. Feel free to move it (:.

router/router.go
jmattheis: Really? I've removed this lines and made a quick curl request and the headers still were returned
```bash
λ curl -I -X OPTIONS http://localhost/
HTTP/1.1 200 OK
Access-Control-Allow-Headers: X-Gotify-Key,Authorization,Content-Type,Upgrade,Origin,Connection,Accept-Encoding,Accept-Language,Host
Access-Control-Allow-Methods: GET,POST,DELETE,OPTIONS,PUT
Access-Control-Allow-Origin: *
Date: Tue, 29 Jan 2019 19:47:28 GMT
Content-Length: 0
```

Do you have a config.yml? In there headers set in line 63 can be overriden.

eternal-flame-AD: Okay, I got this. I copied the configuration from the example and that override the default behavior. I'll revert this. Thanks for your reminder:)

api/plugin.go
eternal-flame-AD: Okay, I will update the swagger docs for this

api/user.go
eternal-flame-AD: https://github.com/gotify/server/blob/d89e6e5b81622b78637cd3218749382efc8802cb/api/user.go#L161-L165

So the user is actually created(on the database), but a 500 will be returned if an error is present. This is a little ambigous from my point.

My opinion is that, if you think we should stumble across the creation/deletion process even one of the callbacks failed, just use `log` to log an error on the server side but still returns an `200 OK`, which is my initial implementation. If we should abort, the record should not be written into the database at all.

eternal-flame-AD:
@jmattheis Are you very concerned about keeping the perfect 100% test coverage? I have tried to cover as many lines as possible while not making the test even more complicated than the code itself. There are a few lines in `plugin/manager.go` which is not covered. Covering them would require writing a dedicated plugin just to emulate an initialization failure, etc. Its okay if you still insist on covering them, just tell me and I will have them covered.

Since travis is no longer building, I uploaded the current coverage as of 7117bf9. The current calculated total coverage is 97.97%

[coverage.txt](https://github.com/gotify/server/files/2811578/coverage.txt)

To generate an html representation:
```bash
$ go tool cover -html=coverage.txt -o=coverage.html
```

Makefile
```diff
@@ -22,6 +22,7 @@ test-coverage:
 			rm profile.out ; \
 		fi \
 	done
+	awk 'BEGIN{print "mode: atomic"} /^(.+):([0-9]+).([0-9]+),([0-9]+).([0-9]+) ([0-9]+) ([0-9]+)$$/ {print}' coverage.txt > coverage_new.txt && mv coverage_new.txt coverage.txt
```
eternal-flame-AD: I added this to remove the extra `mode: atomic` and empty lines caused by multiple `go test` invocations. It seems that codecov accepts the original form but the official `go tool cover` refuses them.

jmattheis:
I'll review this PR on the weekend. I'm okay with not having 100% codecov (but I'd still prefer it).

Makefile
eternal-flame-AD: Actually I think a `go test ./...` works here. The `vendor` directory is by default excluded from the test.

eternal-flame-AD:
25b39b0 and ad72981 are some optimizations on tests to make them run faster (more than 1 minute on my machine to ~10s). If you don't like them I will remove them, or I can cherry-pick them to a new PR after this one is merged.

eternal-flame-AD:
Since we are entering a rather latter stage in this PR, I am deploying this branch to my own server and I will polish it as I use this feature in reality:)

plugin/manager.go
```diff
@@ -0,0 +1,318 @@
...
+	if compat.HasSupport(instance, compat.Messenger) {
+		instance.SetMessageHandler(redirectToChannel{
+			ApplicationID: pluginConf.ApplicationID,
+			UserID:        pluginConf.UserID,
+			Messages:      m.messages})
+	}
+	if compat.HasSupport(instance, compat.Storager) {
+		instance.SetStorageHandler(dbStorageHandler{pluginConf.ID, m.db})
+	}
+	if compat.HasSupport(instance, compat.Configurer) {
```
eternal-flame-AD: Should we in some way provide a way for plugin to update its config scheme seamlessly? I think here we have several options:

- Leave it as is, the plugin updates configuration scheme by adding more fields and deprecating existing fields in its `Config` struct. The user will have to update configuration scheme manually. If `ValidateAndSetConfig` returns an error during initialization, the configuration is automatically replaced by the new `DefaultConfig`
```golang
const pluginVersion = 1
type Conf struct {
    SomeKey string
}
```
```golang
const pluginVersion = 2
type Conf struct {
    SomeKey        string
    SomeFeature bool
}
```
- Record a config scheme number along with the config itself for the plugin, and the plugin implement an interface like `ConfigurerUpgrader`. The interface implementation would look like this:
```golang
func (c *Plugin) ConfigSchema() uint {
    return 8
}

func (c *Plugin) UpgradeConfig(schema uint, yamlConf []byte) interface{} {
    if canUpgradeFrom(schema) {
        return upgradeConfig(schema, yamlConf)
    }
    return c.DefaultConfig()
}
```

plugin/manager.go
```diff
@@ -0,0 +1,318 @@
...
+		if len(pluginConf.Config) == 0 {
+			// The Configurer is newly implemented
+			// Use the default config
+			pluginConf.Config, _ = yaml.Marshal(instance.DefaultConfig())
+			m.db.UpdatePluginConf(pluginConf)
+		}
+		c := instance.EmptyConfig()
```
eternal-flame-AD: Why not use `DefaultConfig()` to get a pointer to the new Configuration? This not only removes a poilerplate method(`func (c *Plugin) EmptyConfig() interface{} { return new(Config) }`) but also enables plugin to provide default values to some unset keys, eg:
```golang
type Config struct {
    // default: true
    VerifyJWTSignature bool
}

func (c *Plugin) DefaultConfig() interface{} {
     return &Config{true}
}
```
In this case, if this key is missing in the real configuration, it will be set to true by default.

plugin/manager.go
```diff
@@ -0,0 +1,318 @@
...
+		yaml.Unmarshal(pluginConf.Config, c)
+		instance.ValidateAndSetConfig(c)
+	}
+	if compat.HasSupport(instance, compat.Webhooker) {
+		instance.RegisterWebhook(*m.mux.Group(pluginConf.Token))
```
eternal-flame-AD: I have to admit that I made a mistake here while writing the plugin API. `gin.RouterGroup` is a struct that should be stored as pointers and I mistook it for an type alias. By now I did not see a problem with this but maybe we should fix this in case some day a change in the `gin` implementation breaks this.

Makefile
jmattheis: Yeah, before go1.10 it wasn't possible to test multiple packages with coverage enabled.

router/router.go
```diff
@@ -31,12 +37,22 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 		DB:       db,
 		ImageDir: conf.UploadedImagesDir,
 	}
-	userHandler := api.UserAPI{DB: db, PasswordStrength: conf.PassStrength, NotifyDeleted: streamHandler.NotifyDeletedUser}
+	userChangeNotifier := new(api.UserChangeNotifier)
+	userHandler := api.UserAPI{DB: db, PasswordStrength: conf.PassStrength, UserChangeNotifier: userChangeNotifier}
 
-	g := gin.New()
+	pluginManager, err := plugin.NewManager(db, conf.PluginsDir, g.Group("/plugin/:id/custom/"), streamHandler)
```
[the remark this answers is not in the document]
jmattheis: What do you think it the best? 

I'd be okay with using a different root path but I'd prefer using `/custom/P000000000` because the plugins can do pretty much everything, it doesn't have to be a hook. 
The last option seems also good.

plugin/manager.go
jmattheis: Oke, then lets change it.

plugin/manager.go
jmattheis: Good Idea!

plugin/messagehandler.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// MessageWithUserID encapsulates a message with a given user ID
+type MessageWithUserID struct {
+	Message model.Message
+	UserID  uint
+}
+
+// SendMessage sends a message to the underlying message channel
+func (c redirectToChannel) SendMessage(msg compat.Message) error {
```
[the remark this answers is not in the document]
jmattheis: We should definitily prevent it, the user can delete the application when the application is disabled, when the Message then gets added to the database it has an application id which doesn't exist.

api/plugin.go
```diff
@@ -0,0 +1,390 @@
...
+		newconfBytes, err := ioutil.ReadAll(ctx.Request.Body)
+		if err != nil {
+			ctx.AbortWithError(500, err)
+			return
+		}
+		if err := yaml.Unmarshal(newconfBytes, newConf); err != nil {
+			ctx.AbortWithError(400, err)
+			return
+		}
+		if err := instance.ValidateAndSetConfig(newConf); err != nil {
```
jmattheis: Can we validate newConf with the schema before it will be set on the plugin?

plugin/manager.go
jmattheis: Thats not an easy question, I think that it will be overkill if we add config migration.

How about we do it like this:
* admin updates plugin which has a different config object.
* on startup gotify/server will disable all plugins with an invalid config
* users then need to fix the config before they are able to enable the plugin again

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+                            });
+                    }}>
+                    Save
+                </Button>
+            </div>
+        );
+    }
+}
+
+class DisplayerPanel extends Component<{pluginInfo: IPlugin; displayText: string}> {
```
jmattheis: React.Component's with no state can be converted to a StatelessFunctionalComponent, they are a little easier to read IMO.
```typescript
const DisplayerPanel: React.SFC<{displayText: string}> = ({displayText}) => {
    return <ReactMarkDown source={displayText} />;
};
```

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+                            displayText={this.state.displayText || 'Loading...'}
+                        />
+                    </PanelWrapper>
+                ) : null}
+            </DefaultPage>
+        ) : null;
+    }
+}
+
+class PanelWrapper extends Component<{
```
jmattheis: This can also be converted to a stateless functional compoenent.

If the props have more than one entry, then I'd prefer an separate interface for the props. F.ex.:
```typescript
interface PanelWrapperProps {
    name: string;
    description?: string;
    refresh?: () => Promise<void>;
    icon?: React.ComponentType;
}
const PanelWrapper: React.SFC<PanelWrapperProps> = ({name, description, refresh, icon}) => ...
```

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+    displayText: string | null;
+    currentConfig: string | null;
+}
+
+export default class PluginDetailView extends Component<IProps, IState> {
+    public state: IState = {pluginInfo: null, displayText: null, currentConfig: null};
+
+    public componentDidMount() {
+        const {match} = this.props;
+        axios.get<IPlugin[]>(`${config.get('url')}plugin`).then((response) => {
```
jmattheis: The plugins should already be loaded inside the PluginStore, could you inject the pluginStore into the PluginDetailView and then use the already existing values there? The pluginStore does also already have a getByID method which can be used here.

ui/src/tests/application.test.ts
```diff
@@ -17,14 +17,55 @@ enum Col {
...
+    id: number,
+    data: {name?: string; description?: string}
+): (() => Promise<void>) => {
+    return async () => {
+        await page.click($table.cell(id, Col.EditUpdate, '.edit'));
+        await page.waitForSelector($dialog.selector());
+        if (data.name) {
+            const nameSelector = $dialog.input('.name');
+            await page.focus(nameSelector);
+            const curLength: number = await page.$eval(
```
jmattheis: Use utils.ts#clearField instead of manually hitting backspace.

api/plugin.go
```diff
@@ -0,0 +1,390 @@
...
+		if conf == nil {
+			ctx.AbortWithError(404, errors.New("unknown plugin"))
+			return
+		}
+		if aborted := pluginOwnerOrAbort(ctx, conf); aborted {
+			return
+		}
+		_, err := c.Manager.Instance(id)
+		if err != nil {
+			ctx.AbortWithError(404, errors.New("plugin instance not found"))
```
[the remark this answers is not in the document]
eternal-flame-AD: Maybe the user removed the plugin file but the config is still stored so that it can be restored after the plugin file was added again.

plugin/manager.go
```diff
@@ -0,0 +1,318 @@
...
+	return nil, errors.New("instance not found")
+}
+
+// HasInstance returns whether the given plugin ID has a corresponding instance
+func (m *Manager) HasInstance(pluginID uint) bool {
+	instance, err := m.Instance(pluginID)
+	return err == nil && instance != nil
+}
+
+// RemoveUser disabled all plugins of a user when the user is disabled
```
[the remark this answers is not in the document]
eternal-flame-AD: The user ID will never be the same right? I think as long as the user ID is different the config will not be returned.

Still I will resolve this as keeping the obsolete config is useless.

eternal-flame-AD:
Thanks for your review:)
I will push changes after updating the [API scheme](https://github.com/gotify/plugin-api/pull/2) so things don't get messed up.

api/plugin_test.go
```diff
@@ -0,0 +1,702 @@
...
+
+	{
+		test.WithUser(s.ctx, 1)
+
+		s.ctx.Request = httptest.NewRequest("GET", fmt.Sprintf("/plugin/%d/display", conf.ID), nil)
+		s.ctx.Params = gin.Params{{Key: "id", Value: fmt.Sprint(conf.ID)}}
+		s.a.GetDisplay(s.ctx)
+
+		assert.Equal(s.T(), 200, s.recorder.Code)
+		test.JSONEquals(s.T(), mockInst.DisplayString, s.recorder.Body.String())
```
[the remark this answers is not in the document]
eternal-flame-AD: I think since it is still an API, returning JSON for everything is good for keeping things consistent as escaping some unwanted values which might cause issues during transfer(maybe?). `axios` will parse JSON automatically by its `content-type`

plugin/manager.go
eternal-flame-AD: I think that's a good idea. My expected behavior is:
1. Hit an error during Unmarshaling or ValidateAndSetConfig in initialization
2. Obtain a new DefaultConfig
3. Comment out existing configs and append to DefaultConfig
4. Use that as the new config
5. Disable the plugin

plugin/messagehandler.go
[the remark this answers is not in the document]
eternal-flame-AD: The user can't delete applications when plugins are disabled. The user can only delete them when the plugin file is missing.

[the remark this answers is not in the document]
jmattheis: Okay, I don't see this as a real problem as well, if the plugin author is malicious then sending messages is one of the smaller problems (:.

api/plugin.go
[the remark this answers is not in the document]
jmattheis: Right, didn't thought about that.

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    private async refreshFeatures() {
+        await this.refreshDisplayer();
+        await this.refreshConfigurer();
+    }
+
+    private async refreshConfigurer() {
+        if (this.state.pluginInfo!.capabilities.indexOf('configurer') !== -1) {
+            const response = await axios.get(
```
jmattheis: I'd prefer if the url calls are done inside the plugin store. (Please do the same for the other usages of `!`)
```typescript
class PluginStore {
    public requestConfig = (id: number): Promise<String> => {
         return axios.get(`${config.get('url')}plugin/${id}/config`).then(response => response.data);
    };
}
```

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+        });
+    }
+
+    private async refreshFeatures() {
+        await this.refreshDisplayer();
+        await this.refreshConfigurer();
+    }
+
+    private async refreshConfigurer() {
+        if (this.state.pluginInfo!.capabilities.indexOf('configurer') !== -1) {
```
jmattheis: This can produce an undefined error. So check if the variable is undefined:
```suggestion
        if (this.state.pluginInfo && this.state.pluginInfo.capabilities.indexOf('configurer') !== -1) {
```

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+        axios.get<IPlugin[]>(`${config.get('url')}plugin`).then((response) => {
+            const pluginInfo: IPlugin = response.data.filter(
+                (p) => p.id.toString() === match.params.id
+            )[0];
+            this.setState({pluginInfo});
+            this.refreshFeatures();
+        });
+    }
+
+    private async refreshFeatures() {
```
jmattheis: Now both request calls are executed in sequence, but they could been done in parrallel:
```
 private async refreshFeatures(): Promise<void> {
        return Promise.all([this.refreshConfigurer(), this.refreshDisplayer()]).then(() => {});
    }
```

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+                </h1>
+                {this.props.description ? (
+                    <h4 style={{color: 'grey'}}>{this.props.description}</h4>
+                ) : null}
+                <hr />
+                <div
+                    className={this.props.name
+                        .toLowerCase()
+                        .trim()
+                        .replace(/ /g, '-')}>
```
jmattheis: Can be done without regex.
```suggestion
                        .replace(' ', '-')}>
```

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    variant="contained"
+                    color="primary"
+                    fullWidth={true}
+                    disabled={
+                        !this.state.unsavedChanges ||
+                        this.state.unsavedChanges === this.props.initialConfig
+                    }
+                    className="config-save"
+                    onClick={() => {
+                        let newConfig = ((this.refs.monaco as any).editor as any) // tslint:disable-line
```
jmattheis: Can't you use the state variable `unsavedChanges` here? Instead of getting the value from the monaco ref?

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+                            .getModel()
+                            .getValue();
+                        this.props
+                            .save(newConfig)
+                            .then(() => {
+                                this.setState({unsavedChanges: ''});
+                            })
+                            .catch((err) => {
```
jmattheis: The errors should already be logged into the console, so this should not be needed.

ui/src/tests/plugin.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    return await innerText(page, `.plugin-info .${className} > span`);
+};
+
+const getDisplayer = async () => {
+    return await innerText(page, '.displayer');
+};
+
+const inDetailPage = async (id: number, callback: (() => Promise<void>)) => {
+    await page.click($table.cell(id, Col.Details));
+    await page.waitForFunction(`window.location.toString().endsWith("${id}")`);
```
jmattheis: Use waitForExists instead and check for the plugin name: (plugin name should be given via method parameters)
```suggestion
    await waitForExists(page, selector.heading(), 'test plugin')
```

ui/src/tests/plugin.test.ts
```diff
@@ -0,0 +1,161 @@
...
+        });
+    });
+    if (!pluginSupported) {
+        return;
+    }
+    describe('functionality test', () => {
+        describe('initial status', () => {
+            it('has echo plugin', async () => {
+                expect(await count(page, $table.rows())).toBe(1);
+                expect(await innerText(page, $table.cell(1, Col.Name))).toBeTruthy();
```
jmattheis: Could you equal the actual name of the echo plugin?

api/plugin.go
eternal-flame-AD: The schema is generated based on the struct so I don't see a possibility that the unmarshaling succeeded but the schema is incompatible.

ui/src/tests/setup.ts
```diff
@@ -19,14 +19,23 @@ const windowsPrefix = process.platform === 'win32' ? '.exe' : '';
 const appDotGo = path.join(__dirname, '..', '..', '..', 'app.go');
 const testBuildPath = path.join(__dirname, 'build');
 
-export const newTest = async (): Promise<GotifyTest> => {
+export const newTest = async (plugins?: string[]): Promise<GotifyTest> => {
```
jmattheis: Could we only pass the plugin dir here, and build the plugins in a separate method that is only called in the plugin tests?

plugin/messagehandler.go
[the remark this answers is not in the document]
eternal-flame-AD: Yeah, loading plugins are actually executing real executables so they would the same platform rights as gotify does, so here we are assuming that users have checked the plugin for sanity before loading them. The question we need to consider is how to help plugin writers write plugin easily and less likely to make unwise decisions(such as why not to expose the Content to Displayer API).  So adding this restriction could in some way promote this but I am wondering whether it is necessary.

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+                {this.props.pluginInfo.capabilities.indexOf('webhooker') !== -1 ? (
+                    <p>
+                        Custom Route Prefix:{' '}
+                        {((url) => (
+                            <a
+                                href={url}
+                                target="_blank"
+                                style={{color: 'white'}}
+                                className="custom-route">
+                                {url}
```
jmattheis: Can you change this line in the index.tsx
```
const url = slashes.concat(port ? hostname.concat(':', port) : '');
```
Otherwise, the url looks like this: `http://142.1.2.3:/plugin/1/custom/PJSZDRWRvIfTAJC/`

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,259 @@
...
+    public render() {
+        return <ReactMarkDown source={this.props.displayText} />;
+    }
+}
+class PluginInfo extends Component<{pluginInfo: IPlugin}> {
+    public render() {
+        return (
+            <div>
+                {this.props.pluginInfo.name ? (
+                    <p className="name">
```
jmattheis: Wrap all texts inside Typography, then the font will be applied to it:
```
<Typography variant="p" className="name">
    Name: {this.props.pluginInfo.name}
</Typography>
```

plugin/compat/v1.go
```diff
@@ -0,0 +1,183 @@
...
+func (c *PluginV1Instance) SetMessageHandler(h MessageHandler) {
+	if c.messenger != nil {
+		c.messenger.SetMessageHandler(&PluginV1MessageHandler{WrapperHandler: h})
+	}
+}
+
+// RegisterWebhook see papiv1.Webhooker
+func (c *PluginV1Instance) RegisterWebhook(basePath string, mux *gin.RouterGroup) {
+	if c.webhooker != nil {
+		c.webhooker.RegisterWebhook(basePath, mux)
```
jmattheis: I kinda don't like, that the webhook will still be registed even if the plugin is disabled.

jmattheis: We could add a middleware to the plugin group which returns 400 bad request if the plugin is disabled.

api/plugin_test.go
```diff
@@ -0,0 +1,702 @@
...
+	conf := s.getDanglingConf(1)
+
+	{
+		test.WithUser(s.ctx, 1)
+
+		s.ctx.Request = httptest.NewRequest("GET", fmt.Sprintf("/plugin/%d/config", conf.ID), nil)
+		s.ctx.Params = gin.Params{{Key: "id", Value: fmt.Sprint(conf.ID)}}
+		s.a.GetConfig(s.ctx)
+
+		assert.Equal(s.T(), 404, s.recorder.Code)
```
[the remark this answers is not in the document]
eternal-flame-AD: See https://github.com/gotify/server/pull/110#discussion_r253242196

plugin/manager.go
```diff
@@ -0,0 +1,318 @@
...
+		UserID:     userID,
+		ModulePath: info.ModulePath,
+		Token:      auth.GeneratePluginToken(),
+	}
+	if compat.HasSupport(instance, compat.Configurer) {
+		pluginConf.Config, _ = yaml.Marshal(instance.DefaultConfig())
+	}
+	if compat.HasSupport(instance, compat.Messenger) {
+		app := &model.Application{
+			Token:       auth.GenerateApplicationToken(),
```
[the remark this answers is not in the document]
eternal-flame-AD: I would say it would be more "human-friendly" but it does not bring any real benefit both to users and developers, and requires extra code both on the server and client side. Let's leave this for later.

ui/src/application/UpdateApplicationDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+                <DialogContent>
+                    <DialogContentText>
+                        An application is allowed to send messages.
+                    </DialogContentText>
+                    <TextField
+                        autoFocus
+                        margin="dense"
+                        className="name"
+                        label="Name *"
+                        type="email"
```
jmattheis:
```suggestion
                        type="text"
```

plugin/compat/wrap_test.go
```diff
@@ -0,0 +1,152 @@
...
+	})
+
+	assert.NotEqual(s.T(), "", inst.GetDisplay())
+}
+
+func TestCompatSuite(t *testing.T) {
+	suite.Run(t, new(CompatSuite))
+}
+
+func TestWrapIncompatiblePlugins(t *testing.T) {
```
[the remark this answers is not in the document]
eternal-flame-AD: Opening the plugin file is handled in `Manager`, I will add tests there.

api/plugin_test.go
```diff
@@ -0,0 +1,702 @@
...
+func (s *PluginSuite) Test_GetDisplay_NotImplemented_expectEmptyString() {
+	conf := s.db.GetPluginConfByUserAndPath(1, mock.ModulePath)
+	inst, err := s.manager.Instance(conf.ID)
+	assert.Nil(s.T(), err)
+	mockInst := inst.(*mock.PluginInstance)
+
+	mockInst.SetCapability(compat.Displayer, false)
+	defer mockInst.SetCapability(compat.Displayer, true)
+
+	{
```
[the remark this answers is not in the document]
eternal-flame-AD: I find adding curly brackets here makes the tests easier to read. It separated the tests into two phases, the first it preparation works such as generating datas and retrieve initialization data from `SetupSuit`, the second phase is the real request phase, which is very similar beween tests, it also makes future updating the tests easier and you can easily spot where to update each test if a change in the requests phase is induced. If you still don't like them I will remove them.

ui/src/application/UpdateApplicationDialog.tsx
eternal-flame-AD: I think I copied it from an existing code:(. Maybe update them both?

https://github.com/gotify/server/blob/master/ui/src/application/AddApplicationDialog.tsx#L43-L49

plugin/compat/v1.go
eternal-flame-AD: Then there would be too much burden on the database if the webhooks are called at a really high frequency. Maybe use the [callback](http://gorm.io/docs/write_plugins.html) function on the `GormDatabse` to provide notifications? (actually I was also wondering why we need to implement notifications for UserAdded and UserRemoved manually).

Since this PR is really getting uncontrollably bigger, lets take these unnessesery changes later if you choose to migrate to gorm callbacks.

ui/src/plugin/PluginDetailView.tsx
eternal-flame-AD: Then only the first whitespace will be replaced.

plugin/messagehandler.go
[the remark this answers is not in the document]
jmattheis: Hmm, lets only output a error message when a plugin tries to send a message.

api/plugin.go
jmattheis: In should be possible to add properties like minLength to the json schema. It can be parsed successfully but doesn't have to be valid.

api/plugin_test.go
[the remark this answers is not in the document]
jmattheis: hmm k.

plugin/manager.go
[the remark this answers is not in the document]
jmattheis: Oke.

ui/src/application/UpdateApplicationDialog.tsx
jmattheis: Yeah, please fix both (:.

plugin/compat/v1.go
jmattheis: I don't think the check for enabled is a problem for the database, on every other request the client/apptoken will get queried for authentication.

Yeah lets not use callbacks for now.

jmattheis:
I've added a plugin template here https://github.com/gotify/plugin-template this automatically builds the shared objects.

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,267 @@
...
+    public state = {unsavedChanges: ''};
+
+    public render() {
+        return (
+            <div>
+                <MonacoEditor
+                    language="yaml"
+                    theme="vs-dark"
+                    height="15em"
+                    ref="monaco"
```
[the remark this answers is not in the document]
eternal-flame-AD: I think it might be needed in the condition schema validation so I did not delete it for now.

plugin/manager_test.go
```diff
@@ -0,0 +1,415 @@
+// +build linux darwin
+// +build !race
```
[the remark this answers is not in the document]
eternal-flame-AD: If the main is built with race detector, the test pugin needs to be built with race detector too. I can create a test file built only in race mode to indicate that race detector is enabled.

plugin/example/clock/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/gotify/plugin-api"
+	"github.com/robfig/cron"
+)
+
+// GetGotifyPluginInfo returns gotify plugin info
+func GetGotifyPluginInfo() plugin.Info {
+	return plugin.Info{
+		Name:        "clock",
+		Description: "Sends an hourly reminder",
```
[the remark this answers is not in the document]
eternal-flame-AD: Are you sure? From `man crontab`

```
       In  the  POSIX locale, the user or application shall ensure that a crontab entry is a text file consisting of lines of six fields each.  The fields shall be separated by <blank> characters. The first
       five fields shall be integer patterns that specify the following:

        1. Minute [0,59]

        2. Hour [0,23]

        3. Day of the month [1,31]

        4. Month of the year [1,12]

        5. Day of the week ([0,6] with 0=Sunday)
```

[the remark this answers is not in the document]
jmattheis: At first, I thought so too. But the golang lib for cron uses a format with seconds. https://godoc.org/github.com/robfig/cron#hdr-CRON_Expression_Format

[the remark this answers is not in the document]
eternal-flame-AD: Didn't expect that. Thanks for your reminder:)

ui/src/plugin/PluginDetailView.tsx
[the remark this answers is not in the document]
jmattheis: I think we wont have schema validation in monaco because the plugin I mentioned (https://github.com/pengx17/monaco-yaml) doesn't work in browser.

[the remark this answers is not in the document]
eternal-flame-AD: So this is not an essential feature so maybe we can leave this for a later PR :).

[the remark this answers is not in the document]
jmattheis: Yes.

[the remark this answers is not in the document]
eternal-flame-AD: But I am not getting syntax highlighting for now on my browser, do you experience the same problem, or do you have any idea where could have gone wrong?

ui/src/plugin/PluginDetailView.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    value={
+                        this.state.unsavedChanges
```
jmattheis: The unsaved changes may be empty (when deleting all content) so maybe set it to false and explicitly check for it. Otherwise it is not possible to `ctrl-a backspace`.

eternal-flame-AD:
I am wondering whether we should remove the `/plugin/:id/config/schema` API for now, as it is neither complete in the WebUI and in the server side(there was some problems handling fields with no struct tags). I think it is not neccesary for now and could be implemented later.

jmattheis:
@eternal-flame-AD Yeah. Remove it for now.

ui/src/plugin/PluginDetailView.tsx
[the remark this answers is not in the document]
eternal-flame-AD: Okay, I used `react-codemirror2` as it seemed still active.

api/user.go
eternal-flame-AD: I think we have gone another way a little far, I did not see a problem in this so I think its OK to keep it for now.

ui/package.json
```diff
@@ -6,13 +6,15 @@
     "@material-ui/core": "^1.5.1",
```
jmattheis: Could you also remove the react-monaco-editor dependency?

jmattheis:
Okay, from my side this seems mergable, would you squash your commit?

eternal-flame-AD:
Okay, I would squash them now. Let's wait for a few days for the documentation to be updated and I will start to write more plugins to try out this function more deeply to spot potential underlying bugs.

plugin/manager.go
```diff
@@ -325,6 +319,39 @@ func (m *Manager) initializeSingleUserPlugin(userCtx compat.UserContext, p compa
...
 
+func (m *Manager) initializeConfigurerForSingleUserPlugin(instance compat.PluginInstance, pluginConf *model.PluginConf) {
+	if len(pluginConf.Config) == 0 {
+		// The Configurer is newly implemented
+		// Use the default config
+		pluginConf.Config, _ = yaml.Marshal(instance.DefaultConfig())
+		m.db.UpdatePluginConf(pluginConf)
+	}
+	c := instance.DefaultConfig()
+	if yaml.Unmarshal(pluginConf.Config, c) != nil || instance.ValidateAndSetConfig(c) != nil {
```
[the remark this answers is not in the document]
jmattheis: Okay could you document this in gotify/website? (what should be validated in this method and that the default config must be valid)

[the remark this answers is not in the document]
eternal-flame-AD: Okay

eternal-flame-AD:
Okay I have squashed the commits. You can close this if you want to but I would recommend waiting for a few days for some new insights, since it is not ready to go to release because the documantation is not ready yet.

eternal-flame-AD:
Maybe take a look at https://github.com/eternal-flame-AD/gotify-netlify, I have set up a working building envrionment. Though I cannot get the local build environment exactly the same as what it is on travis, so I would get an error loading the plugin built on travis environment with a local built gotify server, but I think the problem will go away after the first official release with plugin is out.

jmattheis:
Hopefully this will work (:, otherwise plugins will not be very useful, or rather only useful for users with knowledge on building go programs.

jmattheis:
@eternal-flame-AD Do you by chance have experience with https://github.com/yuin/gopher-lua#libraries-for-gopherlua it would be great to have an alternative to go plugin system, I think the plugin code is good encapsulated so it should be fairly easy to support both (If we do this, it should be done in a separate PR).

eternal-flame-AD:
Okay I think this is ready. You can merge this together with the documentation PR.

jmattheis:
Yeah, will do it after some final testing which I'll probably do on the weekend.

eternal-flame-AD:
Okay, do you want me to squash the commits again or the current state is OK?

jmattheis:
The current state is ok. 

Could you have a look at https://github.com/gotify/build? I've added docker images for building gotify, this should unify the build environment and make it possible to locally build gotify/server and still use prebuild plugins and vice versa. I've updated https://github.com/gotify/plugin-template.

api/plugin.go
```diff
@@ -0,0 +1,395 @@
...
+				Enabled:      conf.Enabled,
+				Capabilities: inst.Supports().Strings(),
+			})
+		}
+	}
+	ctx.JSON(200, result)
+}
+
+// EnablePlugin enables a plugin.
+// swagger:operation POST /plugin/:id/enable plugin enablePlugin
```
jmattheis: `:id` has to be `{id}`. Otherwise requests can't be made from the swagger-ui.

eternal-flame-AD:
Okay, can you merge this into master or review/plugin so we can sort out the whole workflow?

As for gotify/build I think we should not specify the command for building plugins, instead tell the builder to refer to the documentation(see my documentation PR) or the template on deploying plugins, or a builder who only read the documentation of gotify/build might build the plugin without overlaying the corresponding version of server go.mod

As for the template, it will only be possible to build plugins against the master branch of server. I think we should support building against a tag, or even a commit. You can see the .travis.yml on my netlify repo for an example implementation.

jmattheis:
Yeah right, having a single reference how it building should be done is better.

eternal-flame-AD:
Yep.

Feel free to reach to me if you want me to collaborate on the build environment or the template:)

PR #111, opened 2019-01-12
migrate to go.mod

eternal-flame-AD:
This PR migrated dependency management system from dep to go.mod


go.mod
```diff
@@ -0,0 +1,36 @@
+module gotify
```
jmattheis: shouldn't this be `github.com/gotify/server`?

eternal-flame-AD: I named the module gotify so that `go build` will automatically produce an executable called `gotify`(rather than the default `server`). I did observe some similar practices(naming the main module differenct from its import path) on other repos. I read some other repo's practices and found that is a rare case. I will have this changed. Thanks.

jmattheis:
The swagger spec generation sadly doesn't support go modules. We have to workaround this problem.

In the Makefile execute the swagger executable without go modules:
```
update-swagger:
	GO111MODULE=off swagger generate spec --scan-models -o docs/spec.json
```
After this the swagger executable will still not find the dependencies because they aren't on the gopath. So we somehow need to get the go mod dependencies there. The easiest solution I've found so far is to fill the vendor folder. So executing `go mod vendor` inside .travis.yml(install step) should work.

go.sum
```diff
@@ -0,0 +1,425 @@
...
+github.com/gorilla/securecookie v1.1.1/go.mod h1:ra0sb63/xPlUeL+yeDciTfxMRAA+MP+HVt/4epWDjd4=
+github.com/gorilla/sessions v1.1.2/go.mod h1:8KCfur6+4Mqcc6S0FEfKuN15Vl5MgXW92AE8ovaJD0w=
+github.com/gorilla/sessions v1.1.3/go.mod h1:8KCfur6+4Mqcc6S0FEfKuN15Vl5MgXW92AE8ovaJD0w=
+github.com/gorilla/websocket v1.4.0 h1:WDFjx/TMzVgy9VdMMQi2K2Emtwi2QcUQsztZ/zLaH/Q=
+github.com/gorilla/websocket v1.4.0/go.mod h1:E7qHFY5m1UJ88s3WnNqhKjPHQ0heANvMoAMk2YaljkQ=
+github.com/gotify/configor v1.0.2-0.20190112111140-7d9c7c7e6233 h1:ZFYA2/LqyzFP2TunWDdU2XhmmQOSG2JL5aPvLOo6IPQ=
+github.com/gotify/configor v1.0.2-0.20190112111140-7d9c7c7e6233/go.mod h1:Sclq5yNfX/DJHu0TR2k0+Hi34YxsuKTacgrY3z83HoU=
+github.com/gotify/location v0.0.0-20170722210143-03bc4ad20437 h1:4qMhogAexRcnvdoY9O1RoCuuuNEhDF25jtbGIWPtcms=
+github.com/gotify/location v0.0.0-20170722210143-03bc4ad20437/go.mod h1:5JgfyQg+71Ck3uXX/4FBHc4YxdKZ9shU8gs2AUj7Nj0=
+github.com/gotify/server v1.2.1 h1:FfququCa1fDPFKwvsc5EC9CBwVNpMKYh61g4Ipo/n9g=
```
jmattheis: This shouldn't be here, could you run `go mod tidy`?

PR #124, opened 2019-02-05
use gzip middleware

eternal-flame-AD:
Currently, the built js bundle for WebUI is nearly 1 megabyte, which takes quite significant of time to transfer on a unstable network. With the plugin functionality the size nearly doubled (~1.8MB) due to the packed code editor and extra logos. Using gzip can significantly decrease the size transferred (60%+ bandwidth saved).
image

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -7,7 +7,6 @@ dmitri.shuralyov.com/html/belt v0.0.0-20180602232347-f7d459c86be0/go.mod h1:JLBr
 dmitri.shuralyov.com/service/change v0.0.0-20181023043359-a85b471d5412/go.mod h1:a1inKt/atXimZ4Mv927x+r7UpyzRUf4emIoiiSC2TN4=
 dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c/go.mod h1:0PRwlb0D6DFvNNtx+9ybjezNCa8XF0xaYcETyp6rHWU=
 git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999/go.mod h1:fPE2ZNJGynbRyZ4dJvy6G277gSllfV2HJqblrnkyeyg=
-github.com/BurntSushi/toml v0.3.0 h1:e1/Ivsx3Z0FVTV0NSOv/aVgbUWyQuzj7DDnFblkRvsY=
 github.com/BurntSushi/toml v0.3.0/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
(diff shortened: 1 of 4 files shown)


jmattheis:
Could you update this pr, it now has merge conflicts after merging #110 .

eternal-flame-AD:
Okay, I have ran `go mod tidy` against the merge and the confliction should be resolved.

go.sum
```diff
@@ -410,6 +410,7 @@ github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXf
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/tarm/serial v0.0.0-20180830185346-98f6abe2eb07/go.mod h1:kDXzergiv9cbyO7IOYJZWg1U88JhDg3PB6klq9Hg2pA=
+github.com/ugorji/go/codec v0.0.0-20181022190402-e5e69e061d4f/go.mod h1:VFNgLljTbGfSG7qAOspJ7OScBnGdDN/yBr0sguwnwf0=
 github.com/ugorji/go/codec v0.0.0-20181204163529-d75b2dcb6bc8/go.mod h1:VFNgLljTbGfSG7qAOspJ7OScBnGdDN/yBr0sguwnwf0=
 github.com/ugorji/go/codec v0.0.0-20181209151446-772ced7fd4c2 h1:EICbibRW4JNKMcY+LsWmuwob+CRS1BmdRdjphAm9mH4=
 github.com/ugorji/go/codec v0.0.0-20181209151446-772ced7fd4c2/go.mod h1:VFNgLljTbGfSG7qAOspJ7OScBnGdDN/yBr0sguwnwf0=
```
jmattheis: Do you know, why there are diffierent versions for go codec listed here? Shouldn't only one be used?

jmattheis:
@eternal-flame-AD I'm planning on adding a text chat for the Gotify project, which platform would you prefer? gitter, Slack, Discord or something else? (I'm asking you because I think you would probably use it) (I'd probably prefer Discord but I don't really care.)

eternal-flame-AD:
I would prefer slack, I am a little more active on slack than Discord:), but Discord would be great too because a lot of project uses that.

go.sum
eternal-flame-AD: From `go help modules`
```
The go command maintains, in the main module's root directory alongside
go.mod, a file named go.sum containing the expected cryptographic checksums
of the content of specific module versions. Each time a dependency is
used, its checksum is added to go.sum if missing or else required to match
the existing entry in go.sum.
```

So according to my acknoledge, go will add all known versions of a used package in the module cache to `go.sum`, so the cryptographic signature of the package would be more "distributed" across every project. Running `go mod tidy` will not remove them even if I deleted the whole `go.sum` and generate it again.

eternal-flame-AD: I think deleting my module cache and tidying again might solve the problem but it would still appear again in future invocations of `go mod tidy` by me or other contributors.

eternal-flame-AD: I have another theory that the gzip package might require another version of `github.com/ugorji/go/codec`, but I since `go mod tidy` said we need this line let's just keep it

eternal-flame-AD: https://github.com/gin-contrib/gzip/blob/master/go.sum#L22

I found this line, it was indeed overlaied from the gzip package. So we do need this line.

jmattheis: Understood!

PR #125, opened 2019-02-09
Building with docker

jmattheis:
The arm7 docker image wasn't really used and the base image gotify/server used got deprecated.

@eternal-flame-AD Could you have a look at this PR and gotify/website#6?


jmattheis:
> As for the template, it will only be possible to build plugins against the master branch of server. I think we should support building against a tag, or even a commit. You can see the .travis.yml on my netlify repo for an example implementation.

https://github.com/gotify/server/pull/110#issuecomment-462035957

Do you mean [this](https://github.com/eternal-flame-AD/gotify-netlify/blob/master/.travis.yml#L34)? Currently this has only impact on the golang version. What exactly do you mean with building agains a tag/commit. That the gomod-cap check will be executed with the go.mod files from these commits?

eternal-flame-AD:
> Do you mean this? Currently this has only impact on the golang version. What exactly do you mean with building agains a tag/commit. That the gomod-cap check will be executed with the go.mod files from these commits?

Yes.

> Could you have a look at this PR and gotify/website#6?

Okay, I will take a close look at that and leave my review tomorrow :)

jmattheis:
Thanks! Would you mind if I add you as a outside collaborator? Then I could add you as reviewer.

With the plugin-template something like this should already be possible:
```bash
$ make GOTIFY_VERSION=v1.2.1 check
```
Is that what you meant?`

eternal-flame-AD:
> Thanks! Would you mind if I add you as a outside collaborator? Then I could add you as reviewer.

That sounds great:) Thanks

With the plugin-template something like this should already be possible:

Yeah it does have a check-go-mod that checks the dependency against a server version, but it cannot update go.mod automatically to fit another version of server, so practically using only prepared commands in Makefile you cannot update go.mod to fit another server version in the build script other than what was committed in the VCS. In my netlify repo . 
travis.yml loops through a list of predefined targets and updates go.mod for each of them before building for that target. The workflow looks like this:

```bash
make GOTIFY_VERSION=v1.3.0 update-go-mod
make build # build for 1.3.0
make GOTIFY_VERSION=v1.4.0 update-go-mod
make build # build for 1.4.0
```

jmattheis:
In https://github.com/gotify/build/commit/43beb35a22fb67876e5a15cf7cbf473ee56de6e5 I've added automatic building of gotiy/build docker images. They now are tagged like this `gotify/build:{GO_VERSION}-{GOOS}-{GOARCH}`

PR #127, opened 2019-02-11
support reverse proxy with path rewrite

### LINKED ISSUE #122: Possible to run from behind reverse proxy to directory
I have gotify running in docker on port 40001 and hoped to be able to use `https://mypublic.domain/gotify/...` to access it.

e.g. with this nginx config:

    location ^~ /gotify/ {
      proxy_set_header X-Forwarded-Proto $scheme;                                                                                           
      proxy_set_header Host $host;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_pass http://127.0.0.1:40001/;
    }

but the webui does not work because the html references root-relative urls like `/static/css/main.c1fdf91b.css`

Is it possible to either configure it to use relative not root-relative paths, or to include a path prefix?

eternal-flame-AD:
This fixes #122.

Changed the home path so that the page loads static assets relative to the page URL.

Also changed how API path is parsed so that the base path is considered.

### DIFF
--- a/docs/swagger.go
+++ b/docs/swagger.go
@@ -10,11 +10,14 @@ import (
 
 // Serve serves the documentation.
 func Serve(ctx *gin.Context) {
-	url := location.Get(ctx)
-	ctx.Writer.WriteString(get(url.Host))
+	base := location.Get(ctx).Host
+	if bQ := ctx.Query("base"); bQ != "" {
+		base = bQ
+	}
+	ctx.Writer.WriteString(get(base))
 }
 
-func get(host string) string {
+func get(base string) string {
 	box := packr.NewBox("./")
-	return strings.Replace(box.String("spec.json"), "localhost", host, 1)
+	return strings.Replace(box.String("spec.json"), "localhost", base, 1)
 }
--- a/docs/swagger.go
+++ b/docs/swagger.go
@@ -11,8 +11,8 @@ import (
 // Serve serves the documentation.
 func Serve(ctx *gin.Context) {
 	base := location.Get(ctx).Host
-	if bQ := ctx.Query("base"); bQ != "" {
-		base = bQ
+	if basePathFromQuery := ctx.Query("base"); basePathFromQuery != "" {
+		base = basePathFromQuery
 	}
 	ctx.Writer.WriteString(get(base))
 }
--- a/docs/ui.go
+++ b/docs/ui.go
@@ -39,10 +39,15 @@ var ui = `
     <script src="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/3.20.5/swagger-ui-bundle.js"> </script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/3.20.5/swagger-ui-standalone-preset.js"> </script>
     <script>
+    function getBaseURL() {
+      var path = window.location.pathname
+      path = path.substr(0, path.lastIndexOf('/')+1)
+      return window.location.host + path;
+    }
     window.onload = function() {
       // Begin Swagger UI call region
       const ui = SwaggerUIBundle({
-        url: "../swagger",
+        url: "swagger?base="+encodeURIComponent(getBaseURL()),
         dom_id: '#swagger-ui',
         deepLinking: true,
         presets: [
--- a/ui/src/application/Applications.tsx
+++ b/ui/src/application/Applications.tsx
@@ -18,6 +18,7 @@ import AddApplicationDialog from './AddApplicationDialog';
 import {observer} from 'mobx-react';
 import {observable} from 'mobx';
 import {inject, Stores} from '../inject';
+import * as config from '../config';
 import UpdateDialog from './UpdateApplicationDialog';
 
 @observer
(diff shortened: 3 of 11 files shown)


eternal-flame-AD:
The nginx configuration I used to test this:
```nginx
events {}
http {
    server {
        listen 127.0.0.1:10001;
        location /gotify/ {
            rewrite ^/gotify/(.*) /$1  break;
            proxy_pass http://127.0.0.1:10000$uri$is_args$args;
        }
        location /gotify/stream {
            rewrite ^/gotify/(.*) /$1  break;
            proxy_pass http://127.0.0.1:10000$uri$is_args$args;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "Upgrade";
        }
    }
}
```

eternal-flame-AD:
> When accessing the UI under localhost:8080/gotify (note no slash at the end) then it doesn't work.
ngxin won't match the location if you did not add the trailing slash. Maybe caddy behaves differently here so I would fix it:)


> Loading the Swagger UI on the proxy can't load the swagger definition.

ui/src/index.tsx
```diff
@@ -21,9 +21,10 @@ const defaultDevConfig = {
     url: 'http://localhost:80/',
 };
 
-const {port, hostname, protocol} = window.location;
+const {port, hostname, protocol, pathname} = window.location;
 const slashes = protocol.concat('//');
-const url = slashes.concat(port ? hostname.concat(':', port) : hostname);
+const path = pathname.endsWith('/') ? pathname : pathname.substring(0, pathname.lastIndexOf('/'));
```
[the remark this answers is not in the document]
eternal-flame-AD: I considered this again and I don't think this is an issue. An URL without a trailing slash should not silently return `index.html`. Maybe add a 301 redirect to add the trailing slash automatically in the proxy configuration.

eternal-flame-AD:
I induced an API change that application image URLs are no longer resolved to absolute ones. Resolving URIs relative to the gotify path should be handled by the front-end, not back-end.

For the trailing slash problem I changed my `nginx.conf`, and now automatic redirects works:
```nginx
events {}
http {
    server {
        listen 127.0.0.1:10001;
        location = /gotify {
            return 301 /gotify/;
        }
        location /gotify/ {
            rewrite ^/gotify/(.*) /$1  break;
            proxy_pass http://127.0.0.1:10000$uri$is_args$args;
            proxy_redirect ~^/(.*)$ /gotify/$1;
        }
        location /gotify/stream {
            rewrite ^/gotify/(.*) /$1  break;
            proxy_pass http://127.0.0.1:10000$uri$is_args$args;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "Upgrade";
        }
    }
}
```

For caddy config I guess it should be something like this(I am not very familiar with that, I just translated it from the nginx config and I havn't tested this extensively)
I did not find a way to let caddy rewrite 301/302 URLs from proxy upstream. I can only remove the response `Location` header to prevent the browser from recording incorrect 301 redirects.
```Caddyfile
localhost:10001

redir /gotify /gotify/

proxy /gotify/ localhost:10000 {
   without /gotify/
   transparent
   header_downstream -Location
}

proxy /gotify/stream localhost:10000 {
	without /gotify
	websocket
	transparent
}
```

docs/ui.go
```diff
@@ -39,10 +39,15 @@ var ui = `
...
     <script>
+    function getBaseURL() {
+      var path = window.location.pathname
+      path = path.substr(0, path.lastIndexOf('/')+1)
+      return window.location.host + path;
+    }
     window.onload = function() {
       // Begin Swagger UI call region
       const ui = SwaggerUIBundle({
-        url: "../swagger",
```
eternal-flame-AD: The relative URL from `/docs` to `/swagger` should be `swagger`, not `../swagger`. This is fixed automatically by the browser if it is served at root but that was not the case with a path prefix.

ui/src/index.tsx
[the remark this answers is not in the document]
jmattheis: Oke

ui/src/application/Applications.tsx
```diff
@@ -152,7 +153,7 @@ const Row: SFC<IRowProps> = observer(
         <TableRow>
             <TableCell padding="checkbox">
                 <div style={{display: 'flex'}}>
-                    <Avatar src={image} />
+                    <Avatar src={config.get('url') + image} />
```
jmattheis: The message.tsx also has a component which displays the image. This need to be changed to.

docs/ui.go
```diff
@@ -39,10 +39,15 @@ var ui = `
...
+        url: "swagger?base="+encodeURIComponent(getBaseURL()),
```
jmattheis: This is a little hacky IMO, but I cant really think of a better way :(.

docs/swagger.go
```diff
@@ -10,11 +10,14 @@ import (
 
 // Serve serves the documentation.
 func Serve(ctx *gin.Context) {
-	url := location.Get(ctx)
-	ctx.Writer.WriteString(get(url.Host))
+	base := location.Get(ctx).Host
+	if bQ := ctx.Query("base"); bQ != "" {
```
jmattheis: could you name this variable something meaningful?

docs/ui.go
eternal-flame-AD: Yeah if we want the try option on swagger to work I think this is the only way.

pm-pm:
I think this is exactly the solution i am looking for. :-) I hope it will run with apache too.
Could you please compile/pre-release a version with this feature, because i am (in the moment) not able to do it by myself.
Thanks

PR #128, opened 2019-02-12
Save theme settings to localStorage

### LINKED ISSUE #57: save dark/light mode in localstorage

tryzniak:
Let's just stick with a theme preference :)

Closes #57

### DIFF
--- a/ui/src/layout/Layout.tsx
+++ b/ui/src/layout/Layout.tsx
@@ -21,17 +21,6 @@ import {observer} from 'mobx-react';
 import {observable} from 'mobx';
 import {inject, Stores} from '../inject';
 
-const lightTheme = createMuiTheme({
-    palette: {
-        type: 'light',
-    },
-});
-const darkTheme = createMuiTheme({
-    palette: {
-        type: 'dark',
-    },
-});
-
 const styles = (theme: Theme) => ({
     content: {
         margin: '0 auto',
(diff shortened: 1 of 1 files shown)


ui/src/layout/Layout.tsx
```diff
@@ -51,13 +51,21 @@ class Layout extends React.Component<WithStyles<'content'> & Stores<'currentUser
...
+    private localStorageThemeKey = 'gotifyTheme';
 
     public componentDidMount() {
         if (this.version === Layout.defaultVersion) {
             axios.get(config.get('url') + 'version').then((resp: AxiosResponse<IVersion>) => {
                 this.version = resp.data.version;
             });
         }
+
+        const theme = this.darkThemeVisible ? 'dark' : 'light';
```
eternal-flame-AD: Maybe consider refactoring `this.darkThemeVisible` so that we don't have to make the conversion between boolean and `"dark"|"light"` every time? :) f.ex. https://github.com/eternal-flame-AD/gotify-server/tree/review/persisttheme

tryzniak:
Thank you for the expounded PR review. I've made some changes! Check them out.

ui/src/layout/Layout.tsx
```diff
@@ -41,12 +30,26 @@ const styles = (theme: Theme) => ({
     },
 });
 
+const localStorageThemeKey = 'gotify-theme';
+type ThemeKey = 'dark' | 'light';
+const themeMap: {[key in ThemeKey]: Theme} = {
```
jmattheis: TypeScript has a predefined type for this "map" type.
```suggestion
const themeMap: Record<ThemeKey, Theme> = {
```

ui/src/layout/Layout.tsx
```diff
@@ -58,10 +61,17 @@ class Layout extends React.Component<WithStyles<'content'> & Stores<'currentUser
                 this.version = resp.data.version;
             });
         }
+
+        const localStorageTheme = window.localStorage.getItem(localStorageThemeKey);
+        if (localStorageTheme) {
+            this.currentTheme = localStorageTheme as ThemeKey;
```
jmattheis: This type assertion could produce problems in the future. For example if the 'light' theme would be removed from gotify then some users may still have "light" in their local storage. This would result in a `undefined` theme, because 'light' isn't present anymore:
https://github.com/gotify/server/blob/1f0f7d471517d527677009beacc6875e2c1bebc2/ui/src/layout/Layout.tsx#L84

Anyway, this could be fixed with a [User-Defined Type Guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards). It could look like this:
```typescript
const isThemeKey = (value: any): value is ThemeKey => {
    return value === 'light' || value === 'dark';
}
```
and be used like this:
```typescript
const localStorageTheme = window.localStorage.getItem(localStorageThemeKey);
if (isThemeKey(localStorageTheme)) {
    this.currentTheme = localStorageTheme;
} else {
    window.localStorage.setItem(localStorageThemeKey, this.currentTheme);
}
```

tryzniak:
Thank you, folks, for helping me out! I've made some changes. Definitely learned a thing or two about typescript.

PR #136, opened 2019-03-02
Remove -mod=vendor and mount pkg/mod instead

jmattheis:
See gotify/plugin-template#4

Tested build here https://travis-ci.org/gotify/server/builds/500778885#L678

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,12 @@
 LICENSE_DIR=./licenses/
-BUILD_DIR=./build/
+BUILD_DIR=./build
 DOCKER_DIR=./docker/
 SHELL := /bin/bash
 GO_VERSION=`cat GO_VERSION`
+DOCKER_BUILD_IMAGE=gotify/build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,8 +1,12 @@
 LICENSE_DIR=./licenses/
-BUILD_DIR=./build/
+BUILD_DIR=./build
 DOCKER_DIR=./docker/
 SHELL := /bin/bash
 GO_VERSION=`cat GO_VERSION`
+DOCKER_BUILD_IMAGE=gotify/build
+DOCKER_WORKDIR=/proj
+DOCKER_RUN=docker run --rm -v "$$PWD/.:${DOCKER_WORKDIR}" -v "$$GOPATH/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
```
eternal-flame-AD: Not all systems has $GOPATH set. Use `go env GOPATH` instead.

PR #140, opened 2019-03-08
listen on custom address

### LINKED ISSUE #79: Bind to IP address
I would love to have the option to bind the gotify server to an IP address (in my case 127.0.0.1).

I run it behind an nginx reverse proxy under its own subdomain. Nginx takes care of HTTPS, so I only have HTTP enabled in gotify. I would like to only allow the localhost to connect to gotify.

I could easily do this with iptables (and I probably will until this is implemented) but I think a solution in the server would be more elegant.

eternal-flame-AD:
This fixes #79

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -2,11 +2,13 @@
 # Save it to `config.yml` when edited
 
 server:
+  listenaddr: "" # the address to bind on, leave empty to bind on all addresses
   port: 80 # the port the HTTP server will listen on
 
   ssl:
     enabled: false # if https should be enabled
     redirecttohttps: true # redirect to https if site is accessed by http
+    listenaddr: "" # the address to bind on, leave empty to bind on all addresses
     port: 443 # the https port
     certfile: # the cert file (leave empty when using letsencrypt)
     certkey: # the cert key (leave empty when using letsencrypt)
--- a/config/config.go
+++ b/config/config.go
@@ -10,10 +10,12 @@ import (
 // Configuration is stuff that can be configured externally per env variables or config file (config.yml).
 type Configuration struct {
 	Server struct {
-		Port int `default:"80"`
-		SSL  struct {
+		ListenAddr string `default:""`
+		Port       int    `default:"80"`
+		SSL        struct {
 			Enabled         *bool  `default:"false"`
 			RedirectToHTTPS *bool  `default:"true"`
+			ListenAddr      string `default:""`
 			Port            int    `default:"443"`
 			CertFile        string `default:""`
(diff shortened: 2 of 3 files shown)


runner/runner.go
```diff
@@ -21,8 +21,9 @@ func Run(engine *gin.Engine, conf *config.Configuration) {
 			httpHandler = redirectToHTTPS(string(conf.Server.SSL.Port))
 		}
 
+		addr := fmt.Sprintf("%s:%d", conf.Server.ListenAddr, conf.Server.SSL.Port)
```
jmattheis:
```suggestion
		addr := fmt.Sprintf("%s:%d", conf.Server.SSL.ListenAddr, conf.Server.SSL.Port)
```

config/config.go
```diff
@@ -10,10 +10,12 @@ import (
 // Configuration is stuff that can be configured externally per env variables or config file (config.yml).
 type Configuration struct {
 	Server struct {
-		Port int `default:"80"`
-		SSL  struct {
+		ListenAddr string `default:""`
+		Port       int    `default:"80"`
+		SSL        struct {
```
jmattheis: I'm thinking about removing this functionality, most people add tls via a reverse proxy. What do you think?

config.example.yml
```diff
@@ -2,11 +2,13 @@
 # Save it to `config.yml` when edited
 
 server:
+  listenaddr: "" # the address to bind on, leave empty to bind on all addresses
   port: 80 # the port the HTTP server will listen on
 
   ssl:
     enabled: false # if https should be enabled
     redirecttohttps: true # redirect to https if site is accessed by http
+    listenaddr: "" # the address to bind on, leave empty to bind on all addresses
```
jmattheis: Do we need the listen addr two times for http and https or is one enough?

eternal-flame-AD:
> Just for confirmation, you are @eternal_flame:matrix.org right?

Yes.

config.example.yml
eternal-flame-AD: I also don't see a case where we would be listening on different addresses, but the current config structure makes HTTP and HTTPS config somewhat mangled(http listen options is in .server while HTTPS listen options is in .server.ssl) so I kept both to prevent confusion.

config/config.go
eternal-flame-AD: Maybe we can consider this later in another issue.

As for me, I am currently using this feature, as it is much easier to configure for modern TLS1.3 connections, but I am also OK with configuring SSL through reverse proxy.

PR #146, opened 2019-03-10
Use js utf8-in-16 polyfill

### LINKED ISSUE #145: Some characters do not work (password)
Hi,

Using a password manager which generated this password: "3F7ßx[Zy[0ô_<5/MÃd&l"
I couldn't login after I've set it (both on the web ui and the android app)
Fortunately I was still logged in so I changed it back to something with less weird characters and I could log back in again.

If you just don't want to bother supporting those kind of characters, maybe you should tell the user when the password doesn't meet the correct charset

eternal-flame-AD:
This fixes #145. 

I used a solution from MDN to fix the string encoding problem in javascript base64 implementation.

@oxodao I would be very grateful if you could verify the fix and provide a feedback here:)


oxodao:
I confirm this is now working correctly

PR #150, opened 2019-03-10
Fix websocket allowed origin

### LINKED ISSUE #149: 2.0.2 Docker : Websocket: request origin not allowed by Upgrader.CheckOrigin
After starting a clean 2.0.2 docker 8030:80 container I get this error in the terminal when I login.  I didn't experience this error in 1.2.0

```
 Websocket: request origin not allowed by Upgrader.CheckOrigin
```

eternal-flame-AD:
This fixes #149.

See the original issue for description. The main problem for the original implementation is comparing hostname and a `host:port`

### DIFF
--- a/api/stream/stream.go
+++ b/api/stream/stream.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


eternal-flame-AD:
I took a look at my server config and found that I hit that problem too:( I thought it was intended to only allow origins in the config file. When I hit the problem I simply added my origin to the allowedorigins list and thought it was the intended behavior :(

Hard to believe this was not discovered until now which basically prevented zero-conf usage on port other than 80 and 443.

jmattheis:
Are cors headers ignored for web socket requests?, if not then we probably can remove this setting.

eternal-flame-AD:
You mean to remove the whole `allowedorigins` setting?

jmattheis:
Jup, I mean if it is already secured through cors then we don't need a second check.

eternal-flame-AD:
I think the setting was here originally to provide users with options to explicitly allow websocket from other origins. Please allow me some time to investigate on why we need to check websocket origins again.

eternal-flame-AD:
I checked the MDN for websockets and it seems that WebSockets indeed is not bound by CORS restrictions. The question here is do we need to keep the option here to allow users to configure their own allowed origins? IMO I prefer keeping this option.

jmattheis:
It would be great when you could create a test which shows that the problem is fixed.

PR #161, opened 2019-03-12
use crypto/rand for token generation

### LINKED ISSUE #160: Use cryptographic random source for token generation
[math/rand#Seed](https://godoc.org/math/rand#Seed), as is indicated on godoc, only provides at most 31 bit of randomness, which is really not a good choice for generating tokens.

I also want to propose newly generated tokens to be longer, now it has about 78 bits, which is not ideally secure from my point of view.

eternal-flame-AD:
This fixes #160.


auth/token.go
```diff
@@ -10,8 +11,28 @@ var (
 	applicationPrefix = "A"
 	clientPrefix      = "C"
 	pluginPrefix      = "P"
+
+	randReader = rand.Reader
 )
 
+func randIntn(n int) int {
```
jmattheis: I'd prefer a implementation similar to this https://github.com/traggo/server/blob/master/auth/rand/random.go it is easier to read IMO.

auth/mockrandsource.go
```diff
@@ -0,0 +1,12 @@
+package auth
+
+import (
+	"log"
+	"math/rand"
+)
+
+// UseMathRand changes the random generator to math/rand, used only in testing environment
+func UseMathRand() {
```
jmattheis: Could we have a test function like this instead:
```go
func Tokens(tokens... string) func() string {
	return func() string {
		if len(tokens) <= 0 {
			panic("no tokens left")
		}
		token := tokens[0]
		tokens = tokens[1:]
		return token
	}
}
```
and then add a variable to packages that need a token and then explicitly set tokens?
```go
func someTest() {
    tokenCreator = test.Tokens("firstToken", "secondToken")
    // revert tokenCreator with defer
    // execute testing method
}
```

auth/token.go
eternal-flame-AD: I don't really like generation algorithms that make some characters to occur more likely than others:(

Guess I will make this more readable instead.

eternal-flame-AD: For every token character generated, the algorithm is losing ~0.2 bits of entrophy. So for all 14 characters generated we are losing aroung 2.8 bits of entrophy in total. This is not very significant to pose direct risks, but not so negligible that we can ignore, though., in my opinion,

api/application_test.go
```diff
@@ -49,6 +51,7 @@ func (s *ApplicationSuite) BeforeTest(suiteName, testName string) {
 }
 
 func (s *ApplicationSuite) AfterTest(suiteName, testName string) {
+	auth.UseCryptoRand()
```
jmattheis: I'd prefer if this is not exported in the public api, the public api should only provide the crypto rand.

Having something only in the package has serveral benefits:
* less public api which is only used for tests
* no possible wrong usage of the public api (when only crypto is available)

application.go
```go
var authApplicationToken = authGenerateApplicationToken

func (a *ApplicationAPI) CreateApplication(ctx *gin.Context) {
	app := model.Application{}
	if err := ctx.Bind(&app); err == nil {
		app.Token = auth.GenerateNotExistingToken(authGenerateApplicationToken, a.applicationExists)
		app.UserID = auth.GetUserID(ctx)
		app.Internal = false
		a.DB.CreateApplication(&app)
		ctx.JSON(200, withResolvedImage(&app))
	}
}
```
application_test.go
```go
func someTest() {
    old := authApplicationToken
    authApplicationToken = test.Tokens("Abbbbb", "...")
    defer func() {
         authApplicationToken = old
    }
    // do something with the tokens
}
```

eternal-flame-AD: The problem is the mechanism to switch random sources from default crypto to math Rand must call a API or change a global variable in the auth package, no matter from api or test package. Exporting an API instead of changing global variable enables users and developers to see warnings on misuse of this exported API.

jmattheis: I agree but this solution has the benefit that all things are package local so there is no public api which can be used to change the rand. I'd differentiate here changing behavior globally and resetting a private global variable. Also, it is also done this way in other parts of Gotify f.ex. here https://github.com/gotify/server/blob/master/api/stream/client.go#L14

eternal-flame-AD: So we would need to make a copy of Generate(Image|Application|Client)Token in the API package. I could do that if you really want to, but I think the benefit it brings us is really just about as much as the more code to maintain it brings us:(

Also, maybe I think we are going in the wrong direction, instead of making complicated guesses about the generated token, why not just pull it out from test database and verify if the token is valid?

jmattheis: Yeah that would be ok as well. Another option would be to add it to the *Api struct.

eternal-flame-AD: I would update this pr tomorrow.

jmattheis: Okay, thanks.

api/tokens.go
```diff
@@ -0,0 +1,17 @@
+package api
+
+import (
+	"github.com/gotify/server/auth"
+)
+
+var generateApplicationToken = func() string {
```
jmattheis: This could be just:
```
var generateApplicationToken = auth.GenerateApplicationToken
```

PR #164, opened 2019-03-15
Endpoint and UI improvements to update client name

### LINKED ISSUE #101: Make client name changeable
https://github.com/gotify/server/issues/69#issuecomment-447151793 

Like the user and application api the client api should also have an update interface.

GianOrtiz:
Closes #101

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -15,6 +15,7 @@ type ClientDatabase interface {
 	GetClientByID(id uint) *model.Client
 	GetClientsByUser(userID uint) []*model.Client
 	DeleteClientByID(id uint) error
+	UpdateClient(client *model.Client) error
 }
 
 // The ClientAPI provides handlers for managing clients and applications.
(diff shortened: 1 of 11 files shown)


ui/src/client/ClientStore.ts
```diff
@@ -19,6 +19,13 @@ export class ClientStore extends BaseStore<IClient> {
             .then(() => this.snack('Client deleted'));
     }
 
+    @action
+    public update(id: number, name: string): Promise<void> {
+        return axios
+            .put(`${config.get('url')}client/${id}`, {name})
+            .then(() => this.snack('Client updated'));
```
eternal-flame-AD: https://github.com/gotify/server/blob/178c76f410b69f7daaa8442ea051c105856cac74/ui/src/application/AppStore.ts#L38-L43

Please do a refresh after the update. Otherwise, the client name indicated on the clients list will not be updated on the list until the next page refresh.

GianOrtiz:
Waiting for the review @eternal-flame-AD :smiley:

ui/src/client/UpdateClientDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+                            </Button>
+                        </div>
+                    </Tooltip>
+                </DialogActions>
+            </Dialog>
+        );
+    }
+
+    private handleChange(propertyName: string, event: React.ChangeEvent<HTMLInputElement>) {
+        const state = this.state;
```
eternal-flame-AD:
```suggestion
        const state = {};
```
State is merged to the previous state automatically, so I think an empty object is enough. I don't have much react experience though, so feel free to appeal this if you think I am missing something:)

ui/src/tests/client.test.ts
```diff
@@ -56,6 +77,12 @@ describe('Client', () => {
         expect(await innerText(page, $table.cell(2, Col.Name))).toBe('phone');
         expect(await innerText(page, $table.cell(3, Col.Name))).toBe('desktop app');
     });
+    it('updates client', async () => {
+        await updateClient(1, {name: 'firefox'});
```
eternal-flame-AD:
```suggestion
        await updateClient(1, {name: 'firefox'})();
```
Here as well.

ui/src/tests/client.test.ts
```diff
@@ -56,6 +77,12 @@ describe('Client', () => {
...
+    });
+    it('has updated client name', async () => {
+        await hasClient('firefox', 1);
```
eternal-flame-AD:
```suggestion
        await hasClient('firefox', 1)();
```
`hasClient` has type `(name: string, row: number): (() => Promise<void>)` so hasClient has type of `() => Promise<void>`, which is never called, so the test did not do anything. This is why it did not detect that the client name is not changed.

api/client_test.go
```diff
@@ -165,6 +165,40 @@ func (s *ClientSuite) Test_DeleteClient() {
 	assert.True(s.T(), s.notified)
 }
 
+func (s *ClientSuite) Test_UpdateClient_expectSucess() {
```
jmattheis:
```suggestion
func (s *ClientSuite) Test_UpdateClient_expectSuccess() {
```

api/client.go
```diff
@@ -24,6 +25,65 @@ type ClientAPI struct {
...
+// Update a client.
+//
+// ---
+// consumes: [application/json]
+// produces: [application/json]
+// security: [clientTokenHeader: [], clientTokenQuery: [], basicAuth: []]
+// parameters:
+// - name: body
+//   in: body
+//   description: the client name to update
```
jmattheis: While it's true that only the name can be updated, this variable refers to an object which contains the name.
```suggestion
//   description: the client to update
```

api/client.go
```diff
@@ -24,6 +25,65 @@ type ClientAPI struct {
 	NotifyDeleted func(uint, string)
 }
 
+// UpdateClient updates a client name by its id.
```
jmattheis:
```suggestion
// UpdateClient updates a client by its id.
```

ui/src/client/UpdateClientDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+        return (
+            <Dialog
+                open={true}
+                onClose={fClose}
+                aria-labelledby="form-dialog-title"
+                id="client-dialog">
+                <DialogTitle id="form-dialog-title">Update a Client</DialogTitle>
+                <DialogContent>
+                    <DialogContentText>
+                        A client receives message and manages creation of a new tokens and delete
```
jmattheis: How about something like:
```
A client manages messages, clients, applications and users (with admin permissions).
```

PR #173, opened 2019-03-19
use text for string storage in db

### LINKED ISSUE #97: character limits with sqlite3
I just got around to kicking the tires with this project.  Due to a copy/paste mistake, I discovered that there are virtually no character limits in most (maybe all) fields when using sqlite.  For the most part, both the web interface and the mobile app deal with 1600 characters for application name and description along with message title and message.  I have a feeling that is not the intended use case, as varchar(255) is used throughout the sqlite database.

Doing some digging, I learned that sqlite3 will allow the of use VARCHAR during creation, but it will only treat it as TEXT without the limit.  See the sqlite section of [this page][0].  The sqlite documentation on [limits][1] state that sqlite3_limit needs to be used at run time to set SQLITE_LIMIT_LENGTH or it needs to be set at compile time.  Looks like the run time option is possible with [go-sqlite3][2].

FYI, my testing was done with the gotify/server docker container and sqlite3 3.26.0.  Here is as high as my testing got:

```
$ sqlite3 gotify.db 'select length(message) from messages where id=17;'
46846
```

[0]: https://www.vertabelo.com/blog/technical-articles/comparing-char-and-varchar-data-types-in-different-database-engines
[1]: https://sqlite.org/limits.html
[2]: https://github.com/mattn/go-sqlite3/pull/440

eternal-flame-AD:
This fixes #97

### DIFF
--- a/model/application.go
+++ b/model/application.go
@@ -23,12 +23,12 @@ type Application struct {
 	//
 	// required: true
 	// example: Backup Server
-	Name string `form:"name" query:"name" json:"name" binding:"required"`
+	Name string `gorm:"type:text" form:"name" query:"name" json:"name" binding:"required"`
 	// The description of the application.
 	//
 	// required: true
 	// example: Backup server for the interwebs
-	Description string `form:"description" query:"description" json:"description"`
+	Description string `gorm:"type:text" form:"description" query:"description" json:"description"`
 	// Whether the application is an internal application. Internal applications should not be deleted.
 	//
 	// read only: true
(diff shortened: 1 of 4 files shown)


eternal-flame-AD:
Yes, `unique_index`'s can't use text type, as far as I know.

jmattheis:
Okay, I think it is better to let the author of the PR merge it to master, so feel free todo it (:.

eternal-flame-AD:
@jmattheis I can't merge to protected branches.

jmattheis:
Could you retry it?

PR #178, opened 2019-03-22
Do not delete token on network error

jmattheis:


ui/src/message/WebSocketStore.ts
```diff
@@ -42,6 +42,8 @@ export class WebSocketStore {
                 .catch((error: AxiosError) => {
                     if (error && error.response && error.response.status === 401) {
                         this.snack('Could not authenticate with client token, logging out.');
+                    } else {
+                        this.snack('Lost network connection, please refresh the page.');
```
eternal-flame-AD: I think this only displays once, right? So the user would have no idea that it is no longer connected to the server once he missed the initial notification. Can we schedule a reconnect here? Or at least add a persistent/repeating notification so that users can't miss it.

jmattheis: I've added a banner with a retry button.

PR #180, opened 2019-03-28
Allow markdown in messages

### LINKED ISSUE #129: Add markdown support to messages
gotify/android should also support markdown.

jmattheis:
Fixes #129 

I dunno if we should support html, there could be malicious stuff in it. From my tests `<script>` tags wont work, only tested in firefox so far.

### DIFF
--- a/docs/spec.json
+++ b/docs/spec.json
@@ -2021,10 +2021,10 @@
           "example": 25
         },
         "message": {
-          "description": "The actual message.",
+          "description": "The message. Markdown (excluding html) is allowed.",
           "type": "string",
           "x-go-name": "Message",
-          "example": "Backup was successfully finished."
+          "example": "**Backup** was successfully finished."
         },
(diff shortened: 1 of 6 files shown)


jmattheis:
Example:
image

eternal-flame-AD:
Yeah, I tried some XSS and almost every other method except for a direct `<script>` tag works, I prefer to have them mitigated.

Plus, could we sanitize the notification text by only sending `innerText` to the notification? Currently, all formatting characters and HTML tags are displayed in the notification, which is not very beautiful:(

jmattheis:
I'm unsure if this markdown stripping should be done on clientside, we could also do it on server side(with an extra property on model.Message), what do you think?

eternal-flame-AD:
I think we should just do it clientside. Otherwise, the size of the message model would nearly double, which is not really useful from my opinion.

PR #191, opened 2019-05-19
return database error

### LINKED ISSUE #190: Error is not returned on database error
**Can the issue be reproduced with the latest available release? (y/n)** Y

- [ ] Docker
- [x] Linux machine
- [ ] Windows machine

- [ ] None
- [x] Nginx
- [ ] Apache
- [ ] Caddy

- [x] WebUI
- [x] gotify-cli
- [x] Android Client 

**Bug Details**
When i send messages containing emoji then those messages will be pushed but not stored on database. They disappear after refresh.

eternal-flame-AD:
This fixes #190 

TODO:
- [X] Return error for write operations
- [x] Return error for read operations

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -16,9 +16,9 @@ import (
 // The ApplicationDatabase interface for encapsulating database access.
 type ApplicationDatabase interface {
 	CreateApplication(application *model.Application) error
-	GetApplicationByToken(token string) *model.Application
-	GetApplicationByID(id uint) *model.Application
-	GetApplicationsByUser(userID uint) []*model.Application
+	GetApplicationByToken(token string) (*model.Application, error)
+	GetApplicationByID(id uint) (*model.Application, error)
+	GetApplicationsByUser(userID uint) ([]*model.Application, error)
 	DeleteApplicationByID(id uint) error
 	UpdateApplication(application *model.Application) error
 }
@@ -100,11 +100,13 @@ func (a *ApplicationAPI) CreateApplication(ctx *gin.Context) {
 //         $ref: "#/definitions/Error"
 func (a *ApplicationAPI) GetApplications(ctx *gin.Context) {
 	userID := auth.GetUserID(ctx)
-	apps := a.DB.GetApplicationsByUser(userID)
-	for _, app := range apps {
-		withResolvedImage(app)
+	apps, err := a.DB.GetApplicationsByUser(userID)
+	if success := checkErrorOrAbort(ctx, 500, err); success {
+		for _, app := range apps {
+			withResolvedImage(app)
+		}
+		ctx.JSON(200, apps)
 	}
-	ctx.JSON(200, apps)
 }
 
 // DeleteApplication deletes an application by its id.
@@ -143,18 +145,21 @@ func (a *ApplicationAPI) GetApplications(ctx *gin.Context) {
 //         $ref: "#/definitions/Error"
 func (a *ApplicationAPI) DeleteApplication(ctx *gin.Context) {
 	withID(ctx, "id", func(id uint) {
-		if app := a.DB.GetApplicationByID(id); app != nil && app.UserID == auth.GetUserID(ctx) {
-			if app.Internal {
-				ctx.AbortWithError(400, errors.New("cannot delete internal application"))
-				return
-			}
-			if success := checkErrorOrAbort(ctx, 500, a.DB.DeleteApplicationByID(id)); success {
-				if app.Image != "" {
-					os.Remove(a.ImageDir + app.Image)
+		app, err := a.DB.GetApplicationByID(id)
+		if success := checkErrorOrAbort(ctx, 500, err); success {
+			if app != nil && app.UserID == auth.GetUserID(ctx) {
+				if app.Internal {
+					ctx.AbortWithError(400, errors.New("cannot delete internal application"))
+					return
 				}
+				if success := checkErrorOrAbort(ctx, 500, a.DB.DeleteApplicationByID(id)); success {
+					if app.Image != "" {
+						os.Remove(a.ImageDir + app.Image)
+					}
+				}
+			} else {
+				ctx.AbortWithError(404, fmt.Errorf("app with id %d doesn't exists", id))
 			}
-		} else {
-			ctx.AbortWithError(404, fmt.Errorf("app with id %d doesn't exists", id))
 		}
 	})
 }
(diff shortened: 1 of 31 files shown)


api/application.go
```diff
@@ -142,17 +145,21 @@ func (a *ApplicationAPI) GetApplications(ctx *gin.Context) {
 //         $ref: "#/definitions/Error"
 func (a *ApplicationAPI) DeleteApplication(ctx *gin.Context) {
 	withID(ctx, "id", func(id uint) {
-		if app := a.DB.GetApplicationByID(id); app != nil && app.UserID == auth.GetUserID(ctx) {
-			if app.Internal {
-				ctx.AbortWithError(400, errors.New("cannot delete internal application"))
-				return
+		app, err := a.DB.GetApplicationByID(id)
+		if success := checkErrorOrAbort(ctx, 500, err); success {
```
jmattheis: What do you think about returning early in case of an error via a db request.
This function could look like this:
```go
func (a *ApplicationAPI) DeleteApplication(ctx *gin.Context) {
	withID(ctx, "id", func(id uint) {
		app, err := a.DB.GetApplicationByID(id)
		if success := checkErrorOrAbort(ctx, 500, err); !success {
			return
		}
		if app == nil || app.UserID != auth.GetUserID(ctx) {
			ctx.AbortWithError(404, fmt.Errorf("app with id %d doesn't exists", id))
			return
		}

		if app.Internal {
			ctx.AbortWithError(400, errors.New("cannot delete internal application"))
			return
		}
		
		if success := checkErrorOrAbort(ctx, 500, a.DB.DeleteApplicationByID(id)); !success {
			return
		}
		
		if app.Image != "" {
			os.Remove(a.ImageDir + app.Image)
		}
	})
}
```
I'd say it is much more readable. It would be great if you could do this to the newly added ifs with chechErrorOrAbort, so the old code which doesn't follow return early can stay. (I could do it to, as it is not the most exiting thing todo (: )

api/application_test.go
```diff
@@ -155,7 +161,9 @@ func (s *ApplicationSuite) Test_CreateApplication_withExistingToken() {
 
 	expected := &model.Application{ID: 2, Token: secondApplicationToken, Name: "custom_name", UserID: 5}
 	assert.Equal(s.T(), 200, s.recorder.Code)
-	assert.Contains(s.T(), s.db.GetApplicationsByUser(5), expected)
+	if app, err := s.db.GetApplicationsByUser(5); assert.NoError(s.T(), err) {
```
jmattheis: The additional ifs let gocyclo fail right? It could be written like this, then it would have a similar effect but no ifs.
```go
app, err := s.db.GetApplicationsByUser(5)
require.NoError(s.T(), err)
assert.Contains(s.T(), app, expected)
```
Tho, I think we do not need this check in most of the tests. In the tests the returned value is equal checked against an expected value and this check would already fail, if the db threw some error. What do you think?

jmattheis:
Good Work! Sucks that I ignored these errors while creating Gotify, now we have this huge change :/.

Sadly it doesn't have 100% test coverage anymore, but letting the db fail wouldn't be that easy and good looking in the tests, so I'm okay with it.

I'll fully review this PR after you've addressed my two remarks.

api/application.go
eternal-flame-AD: I wrote it that way so codecov is not going to complain about the coverage of that early return statement. If you are okay with not having 100% coverage I will use early return.

jmattheis: Yeah, code quality / clarity > codecov (:.

api/application_test.go
eternal-flame-AD: If we do not stop the further assertion, a panic might happen and the detailed failure information might be lost.

jmattheis: See `require` https://godoc.org/github.com/stretchr/testify/require
> The `require` package have same global functions as in the `assert` package, but instead of returning a boolean result they call `t.FailNow()`.

So after a failure no further assertions will be done.

api/application.go
```diff
@@ -299,11 +319,13 @@ func (a *ApplicationAPI) UploadApplicationImage(ctx *gin.Context) {
 				os.Remove(a.ImageDir + app.Image)
 			}
 
-			app.Image = name + ext
-			a.DB.UpdateApplication(app)
+			app.Image = name
```
jmattheis: Now the image contains the image dir, like this it can't be resolved anymore. Example URL: `http://localhost:8080/image/data/images/LZZCuJ5E5nb_5qGgDo_lwqEne.PNG`

plugin/manager.go
```diff
@@ -172,7 +183,7 @@ func (m *Manager) HasInstance(pluginID uint) bool {
 // RemoveUser disabled all plugins of a user when the user is disabled
 func (m *Manager) RemoveUser(userID uint) error {
 	for _, p := range m.plugins {
-		pluginConf := m.db.GetPluginConfByUserAndPath(userID, p.PluginInfo().ModulePath)
+		pluginConf, _ := m.db.GetPluginConfByUserAndPath(userID, p.PluginInfo().ModulePath)
```
jmattheis: return error here

plugin/pluginenabled.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func requirePluginEnabled(id uint, db Database) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		if conf := db.GetPluginConfByID(id); conf == nil || !conf.Enabled {
+		if conf, _ := db.GetPluginConfByID(id); conf == nil || !conf.Enabled {
```
jmattheis: use checkErrorOrAbort here.

auth/authentication.go
```diff
@@ -35,8 +35,9 @@ func (a *Auth) RequireAdmin() gin.HandlerFunc {
 		if user != nil {
 			return true, user.Admin, user.ID
 		}
-		if token := a.DB.GetClientByToken(tokenID); token != nil {
-			return true, a.DB.GetUserByID(token.UserID).Admin, token.UserID
+		if token, _ := a.DB.GetClientByToken(tokenID); token != nil {
```
jmattheis: This error should also be returned. Same for the other ones in this file.

api/errorHandling.go
```diff
@@ -0,0 +1,10 @@
+package api
+
+import "github.com/gin-gonic/gin"
+
+func checkErrorOrAbort(ctx *gin.Context, code int, err error) (success bool) {
```
jmattheis: Could we name name it something like isSuccessfulOrAbort or hasNoErrorOrAbort, currently it is not so clear IMO.

auth/authentication.go
eternal-flame-AD: I don't think we should return error during user authentication process, as it might contain sensitive information. Or maybe we can return a general 500 error?

jmattheis: yeah, a general error should be enough

auth/authentication.go
```diff
@@ -26,47 +26,56 @@ type Auth struct {
...
+	return a.requireToken(func(tokenID string, user *model.User) (bool, bool, uint, error) {
 		if user != nil {
-			return true, user.Admin, user.ID
+			return true, user.Admin, user.ID, nil
 		}
-		if token, _ := a.DB.GetClientByToken(tokenID); token != nil {
-			user, _ := a.DB.GetUserByID(token.UserID)
-			return true, user.Admin, token.UserID
+		if token, err := a.DB.GetClientByToken(tokenID); err != nil {
+			return true, false, 0, err
```
jmattheis: This should be `false, false, 0, err` first flag is authenticated which should be false here. Please check the other returns, in case of an error the user isn't authenticated.

api/user.go
```diff
@@ -354,14 +380,22 @@ func (a *UserAPI) UpdateUserByID(ctx *gin.Context) {
...
+					return
+				}
+				if !user.Admin && oldUser.Admin && adminCount == 1 {
 					ctx.AbortWithError(400, errors.New("cannot delete last admin"))
 					return
 				}
 				internal := a.toInternalUser(user, oldUser.Pass)
 				internal.ID = id
-				a.DB.UpdateUser(internal)
+				successOrAbort(ctx, 500, a.DB.UpdateUser(internal))
```
jmattheis: return if error

eternal-flame-AD: The current branch head is already the correct one I think? Maybe you are not on the tip of the branch?

PR #196, opened 2019-06-08
support longer plugin storage and message extras

### LINKED ISSUE #195: Plugin Configuration is stored as varbinary(255) on mariaDB
Yes
- [ ] Docker
- [X] Linux machine
- [ ] Windows machine

```
server:
  listenaddr: "127.0.0.1"
  port: 10000 # the port for the http server
  ssl:
    port: 443 # the https port
    certfile: # the cert file (leave empty when using letsencrypt)
    certkey: # the cert key (leave empty when using letsencrypt)
    letsencrypt:
      enabled: false # if the certificate should be requested from letsencrypt
      accepttos: false # if you accept the tos from letsencrypt
      cache: data/certs # the directory of the cache from letsencrypt
      hosts: # the hosts for which letsencrypt should request certificates
      - mydomain.tld
      - myotherdomain.tld
  responseheaders: # response headers are added to every response (default: none)
  stream:
    allowedorigins: # allowed origins for websocket connections (same origin is always allowed)
      - ".+.example.com"
      - "otherdomain.com"
database: # for database see (configure database section)
  dialect: mysql
  connection: root:root@/gotify?charset=utf8&parseTime=True&loc=Local
defaultuser: # on database creation, gotify creates an admin user
  name: admin # the username of the default user
  pass: admin # the password of the default user
pluginsdir: data/plugins
```
</details>

- [X] None
- [ ] Nginx
- [ ] Apache
- [ ] Caddy

```

```
</details>

- [X] WebUI
- [X] gotify-cli
- [X] Android Client 
- [X] 3rd-party API call (Please include your code)

**What did you do?**

Ran a plugin with configuration storage.

Configuration is stored correctly.

A database error occured
```
(Error 1406: Data too long for column 'config' at row 1) 
[2019-06-08 12:04:21]  
panic: Error 1406: Data too long for column 'config' at row 1
```

The `plugin_confs` table is initialized with the column `config` of type `varbinary(255)`, which is too small.
```
MariaDB [gotify]> describe plugin_confs;
+----------------+------------------+------+-----+---------+----------------+
| Field          | Type             | Null | Key | Default | Extra          |
+----------------+------------------+------+-----+---------+----------------+
| id             | int(10) unsigned | NO   | PRI | NULL    | auto_increment |
| user_id        | int(10) unsigned | YES  |     | NULL    |                |
| module_path    | text             | YES  |     | NULL    |                |
| token          | varchar(180)     | YES  | UNI | NULL    |                |
| application_id | int(10) unsigned | YES  |     | NULL    |                |
| enabled        | tinyint(1)       | YES  |     | NULL    |                |
| config         | varbinary(255)   | YES  |     | NULL    |                |
| storage        | varbinary(255)   | YES  |     | NULL    |                |
+----------------+------------------+------+-----+---------+----------------+
8 rows in set (0.000 sec)
```

eternal-flame-AD:
This fixes #195

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -1,8 +1,10 @@
 package database
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/gotify/server/auth/password"
 	"github.com/gotify/server/model"
@@ -35,7 +37,14 @@ func New(dialect, connection, defaultUser, defaultPass string, strength int, cre
 		db.DB().SetMaxOpenConns(1)
 	}
 
-	db.AutoMigrate(new(model.User), new(model.Application), new(model.Message), new(model.Client), new(model.PluginConf))
+	if err := db.AutoMigrate(new(model.User), new(model.Application), new(model.Message), new(model.Client), new(model.PluginConf)).Error; err != nil {
+		return nil, err
+	}
+
+	if err := prepareBlobColumn(dialect, db); err != nil {
+		return nil, err
+	}
+
 	userCount := 0
 	db.Find(new(model.User)).Count(&userCount)
 	if createDefaultUserIfNotExist && userCount == 0 {
@@ -43,6 +52,31 @@ func New(dialect, connection, defaultUser, defaultPass string, strength int, cre
 	}
 
 	return &GormDatabase{DB: db}, nil
+}
+
+func prepareBlobColumn(dialect string, db *gorm.DB) error {
+	alterBlobSQLStmtTpl := "alter table `%s` modify column %s %s"
+	blobType := ""
+	switch dialect {
+	case "mysql":
+		blobType = "longblob"
+	case "postgres":
+		blobType = "bytea"
+	}
+	if blobType != "" {
+		for _, target := range []string{
+			"messages:extras",
+			"plugin_confs:config",
+			"plugin_confs:storage",
+		} {
+			targetSpl := strings.Split(target, ":")
+			targetTbl, targetCol := targetSpl[0], targetSpl[1]
+			if err := db.Exec(fmt.Sprintf(alterBlobSQLStmtTpl, targetTbl, targetCol, blobType)).Error; err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 func createDirectoryIfSqlite(dialect string, connection string) {
--- a/database/database.go
+++ b/database/database.go
@@ -1,10 +1,8 @@
 package database
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/gotify/server/auth/password"
 	"github.com/gotify/server/model"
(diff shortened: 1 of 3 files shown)


model/message.go
```diff
@@ -11,7 +11,7 @@ type Message struct {
 	Message       string `gorm:"type:text"`
 	Title         string `gorm:"type:text"`
 	Priority      int
-	Extras        []byte
+	Extras        []byte `gorm:"type:blob"`
```
jmattheis: Seems like postgres doesn't like `blob`:
```
(pq: type "blob" does not exist) 
[2019-06-08 08:59:57]  
[GIN-debug] [WARNING] Running in "debug" mode. Switch to "release" mode in production.
 - using env:	export GIN_MODE=release
 - using code:	gin.SetMode(gin.ReleaseMode)

panic: pq: relation "plugin_confs" does not exist

goroutine 1 [running]:
github.com/gotify/server/router.Create(0xc0000b2240, 0xc00009e240, 0xc0000ee000, 0x52, 0xc0000b05a0)
	/proj/router/router.go:46 +0x311f
main.main()
	/proj/app.go:51 +0x382
```

eternal-flame-AD: Can you try varbinary and see if that works?

jmattheis: varbinary gives me the same error. `bytea` works for postgres but not for mysql.

Maybe we need some custom alter table if dialect == "mysql" but that would only be a workaround.

jmattheis: wo could also convert the byte array to a string and then save it as text which is supported in both databases.

eternal-flame-AD: So we should not specify the column type in the struct tag but to use an ALTER manually, right?

eternal-flame-AD: I think we could get into encoding issues by storing binary data as text :(

jmattheis: > So we should not specify the column type in the struct tag but to use an ALTER manually, right?

Yup

> I think we could get into encoding issues by storing binary data as text :(

The binary data could be converted to base64 or hex before storing it in the database.

eternal-flame-AD: Can you try if this one works? I tried sqlite3 and mariaDB

database/database.go
```diff
@@ -45,6 +54,31 @@ func New(dialect, connection, defaultUser, defaultPass string, strength int, cre
 	return &GormDatabase{DB: db}, nil
 }
 
+func prepareBlobColumn(dialect string, db *gorm.DB) error {
+	alterBlobSQLStmtTpl := "alter table `%s` modify column %s %s"
```
jmattheis: what does tpl mean? Would it be possible to use modify column https://gorm.io/docs/migration.html#ModifyColumn it would look better:
```
db.Model(&model.Message{}).ModifyColumn("extras", blobType)
```

eternal-flame-AD: Template. Okay I will use `ModifyColumn`.

eternal-flame-AD:
Okay. We can implement it with travis build env. Let's leave it for a later PR.

PR #197, opened 2019-06-11
Fixing some typos in comments

noamran:
Thank you for writing such an awesome service!
I noticed some typos in the comments while going over it, went ahead and fixed them. 
Hope that is OK!

### DIFF
--- a/docs/spec.json
+++ b/docs/spec.json
@@ -1852,13 +1852,19 @@
       "title": "Application Model",
       "required": [
         "id",
-        "token",
+        "Token",
         "name",
         "description",
         "internal",
         "image"
       ],
       "properties": {
+        "Token": {
+          "description": "The application token. Can be used as `appToken`. See Authentication.",
+          "type": "string",
+          "readOnly": true,
+          "example": "AWH0wZ5r0Mbac.r"
+        },
         "description": {
           "description": "The description of the application.",
           "type": "string",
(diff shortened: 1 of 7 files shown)


eternal-flame-AD:
@noamran Would you please run `make update-swagger` and commit the result to have your changes applied to the swagger docs as well? Thanks.

docs/spec.json
```diff
@@ -1852,13 +1852,19 @@
...
         "id",
-        "token",
+        "Token",
         "name",
         "description",
         "internal",
         "image"
       ],
       "properties": {
+        "Token": {
```
jmattheis: The travis build still says that the swagger spec is not up-to-date.

Did you run `make update-swagger` to update this file? If yes, are you using the most recent version of go-swagger? You can update it with `make download-tools`. If it's still yes, then there could be a bug in the generation of the swagger definition.

This is the only change that I get when running `make update-swagger`
```diff
diff --git a/docs/spec.json b/docs/spec.json
index 382659a..639c0ba 100644
--- a/docs/spec.json
+++ b/docs/spec.json
@@ -2067,7 +2067,7 @@
       "x-go-package": "github.com/gotify/server/model"
     },
     "Paging": {
-      "description": "The Paging holds holds information about the limit and making requests to the next page.",
+      "description": "The Paging holds information about the limit and making requests to the next page.",
       "type": "object",
       "title": "Paging Model",
       "required": [
```

noamran: I did run these commands, yet that was the only update it created. Is there any change I should be doing manually?

eternal-flame-AD: https://paste.ubuntu.com/p/T8SjW4zGQX/

I generated a patch file you can apply with here.

Do you have any error message shown when you execute `make update-swagger`?

PR #207, opened 2019-08-05
Authentication plugin

### LINKED ISSUE #203: Authentication Plugin
Hello!

First I want to thank you for this awesome project!

I'd like to propose an extension to the current plugin system that may also relate to #78 and #20. Both issues asked for a different authentication method and I really understand that implementing and maintaining different authentication mechanisms is a bit overkill. However, I think we could extend the current plugin system in a way that the community would be able to contribute authentication/identity-provider plugins. Gotify will still need to maintain a database for applications/clients and user information. The actual authentication process however may be "outsourced" to a plugin.

This would allow a lot of different use-cases, like:

- System Authentication
- OAuth2 Authentication (thanks to `Webhooker` we can already setup the OAuth endpoint)
- LDAP integration into Active Directory/FreeIPA/OpenLDAP (and thus bigger installations)
- Single-Sign-On (may require access to HTTP headers...)
- ...

Extending the current plugin system to support authentication providers would require some kind of system-plugin that is not bound to a specific user. 

If you find this idea interesting I will try to come up with a more detailed proposal on the implementation part.

Best regards
Patrick

eternal-flame-AD:
Closes #203.


auth/authentication.go
```diff
@@ -17,119 +16,163 @@ type Database interface {
...
+	return user, designatedAuthenticator, err
+}
+
+func (a *Auth) getUserInfoFromAuth(ctx *gin.Context) (user *model.User, authenticator string, err error) {
+	if designatedAuthenticator := ctx.GetHeader("X-Gotify-Authenticator"); designatedAuthenticator != "" {
+		return a.useDesignatedAuthenticator(ctx, designatedAuthenticator)
+	}
+
+	internalAuthProvider := a.authenticators[""]
+	if user, err := internalAuthProvider.Authenticate(ctx.Request); user != nil {
```
jmattheis: can the internal and external auth providers be merged? That, we don't differentiate between plugin auth and the external auth?

eternal-flame-AD: I wrote it separately to make sure that the internal provider must be the first to check the request.

auth/authentication.go
```diff
@@ -17,119 +16,163 @@ type Database interface {
...
+
+// AuthenticationProvider provides authentication methods
+type AuthenticationProvider interface {
+	Authenticate(req *http.Request) (user *model.User, err error)
 }
 
 // Auth is the provider for authentication middleware
 type Auth struct {
-	DB Database
+	authenticators map[string]AuthenticationProvider
```
ppacher: Shouldn't we use a list instead of a map here so the order in which external authenticators are used is always the same and configurable?

auth/authentication.go
```diff
@@ -17,119 +16,163 @@ type Database interface {
...
+	DB             Database
 }
 
-type authenticate func(tokenID string, user *model.User) (authenticated bool, success bool, userId uint, err error)
+// RegisterAuthenticationProvider registers a new authentication provider
+// use empty string as key to register internal credential manager
+func (a *Auth) RegisterAuthenticationProvider(key string, auth AuthenticationProvider) {
```
ppacher: Can RegisterAuthenticationProvider be called during runtime or must it be called at startup? In the first case we should protect `authenticators` with a mutex

auth/error.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/gin-gonic/gin"
+)
+
+// AuthenticationError is an interface for authentication-related errors with custom HTTP status code
+type AuthenticationError interface {
+	error
+	Code() int
+}
+
+func abortContextWithAuthenticaionError(ctx *gin.Context, err error) {
```
ppacher: Just a small typo

auth/authentication.go
eternal-flame-AD: I think all providers should be registered during initialization so no concurrent write here :)

auth/authentication.go
eternal-flame-AD: Makes sense. Good idea :+1:

eternal-flame-AD:
Sorry I am experiencing some personal health issues these days, I think I'll come back to keep on with the work here this weekend.

auth/basicauthenticator/basicauth.go
```diff
@@ -0,0 +1,35 @@
+package basicauthenticator
```
jmattheis: can we name this auth/internal/authenticator.go? While it does use basic auth, other providers can use it too, for obtaining credentials.

auth/authentication.go
```diff
@@ -17,119 +16,142 @@ type Database interface {
...
-type authenticate func(tokenID string, user *model.User) (authenticated bool, success bool, userId uint, err error)
+// RegisterAuthenticationProvider registers a new authentication provider
+// the provider is used in the order it i registered
+func (a *Auth) RegisterAuthenticationProvider(auth AuthenticationProvider) {
+	a.authenticators = append(a.authenticators, auth)
+}
 
-// RequireAdmin returns a gin middleware which requires a client token or basic authentication header to be supplied
+func (a *Auth) getUserInfoFromAuth(ctx *gin.Context) (user *model.User, authenticator string, err error) {
+	designatedAuthenticator := ctx.GetHeader("X-Gotify-Authenticator")
```
jmattheis: Is it now required to specify this header? This creates a security risk because a user that exists in our internal auth, could be authenticated via a different origin.

I thought we go through the authenticators one by one and try to authenticate the user.
* if wrong password -> abort request
* if unknown user -> goto next authenticator

model/user.go
```diff
@@ -2,13 +2,14 @@ package model
...
-	ID           uint   `gorm:"primary_key;unique_index;AUTO_INCREMENT"`
-	Name         string `gorm:"type:varchar(180);unique_index"`
-	Pass         []byte
-	Admin        bool
-	Applications []Application
-	Clients      []Client
-	Plugins      []PluginConf
+	ID            uint   `gorm:"primary_key;unique_index;AUTO_INCREMENT"`
+	Name          string `gorm:"type:varchar(180);unique_index"`
+	Authenticator string `gorm:"type:varchar(100);index"`
```
jmattheis: I'd prefer the name Origin. Also, is the index needed?

auth/authentication.go
```diff
@@ -17,119 +16,142 @@ type Database interface {
...
+		}
+		return user, nil
+	}
+
+	// 2. try user authentication
+	user, authenticator, err := a.getUserInfoFromAuth(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if user.ID == 0 {
```
jmattheis: If the user does exist, we need to verify that `user.Authenticator == authenticator`.

auth/error.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Error implements AuthenticationError
+func (e TokenRequiredError) Error() string {
+	return fmt.Sprintf("%s token is required to access this api", e.TokenType)
+}
+
+// Code implements AuthenticationError
+func (e TokenRequiredError) Code() int {
+	return 400
```
jmattheis: 403 Forbidden because the user provided valid authentication, but is not authorized.

auth/util_test.go
```diff
@@ -33,14 +33,14 @@ func (s *UtilSuite) Test_getID() {
 
 func (s *UtilSuite) Test_getToken() {
 	ctx, _ := gin.CreateTestContext(httptest.NewRecorder())
-	RegisterAuthentication(ctx, nil, 1, "asdasda")
+	RegisterAuthentication(ctx, nil, "asdasda")
```
jmattheis: nil, as user isn't allowed anymore, there are a lot occurrences of this in tests.

auth/authentication.go
eternal-flame-AD: I think we are on the same track.

> if wrong password -> abort request
Authenticators can return an error, this would abort the authentication process.

> if unknown user -> goto next authenticator
On line 48, if `designatedAuthenticator == ""`, all authenticator would be tried sequentially, otherwise, only authenticators with matching name will be used.

jmattheis: Ahh, I missed the condition where user and err is nil, can we make this more explicit? Could be by returning a named error like ErrUserNotFound = errors.new("user not found"), or by adding another else branch
```
		} else if err != nil {
			return nil, providerName, err
		} else {
			// continue with next provider
		}
```

eternal-flame-AD: I think we can't let external plugins import our auth package as it would introduce an import loop. The way around, as I see, would be to introduce another method to get whether this is a fatal authentication error in our error interface like maybe `Abort() bool` and default this to `false` if this method is missing.

jmattheis: The error constant could be inside the plugin api. But I'm okay with both ways.

ppacher: I think an error constant inside the plugin API is the better way because it's easier to extend the API afterwards without breaking existing code.

eternal-flame-AD: I don't think it is necessary to introduce another common api package just for authenticators as we did in user plugins, as we don't have as many data types to take care of.

auth/util_test.go
eternal-flame-AD: Sorry could you please explain this, which should be nil?

auth/basicauthenticator/basicauth.go
eternal-flame-AD: `/internal/` in the import path would prevent other packages to import this. I am not quite sure on what you want to achieve here. Could you please explain further? Thanks

jmattheis:
Feel free to reopen this pr, if its ready for another review.

ppacher:
@eternal-flame-AD can I help you on this?

eternal-flame-AD:
Yes! I'm having a tight schedule these days so I expect not being able to work extensively on this project for ~2-3 mo. I will be very grateful if u can fork this and make this happen! :) I'll be available for discussions and reviews tho.

PR #208, opened 2019-08-11
Add health api

jmattheis:


api/health.go
```diff
@@ -0,0 +1,42 @@
...
+// ---
+// produces: [application/json]
+// responses:
+//   200:
+//     description: Ok
+//     schema:
+//         $ref: "#/definitions/Health"
+func (a *HealthAPI) Health(ctx *gin.Context) {
+	if err := a.DB.Ping(); err != nil {
+		ctx.JSON(200, model.Health{
```
eternal-flame-AD: Maybe `503 Service Unavailable` for this?

jmattheis: I'll go with 500, because then it is consistent with the other apis when the db is not available

PR #227, opened 2019-10-11
Allow to send message to specifics clients in an app

Zenteno:
Now, You can send messages like:

curl -X POST "https://push.example.de/message?token=<app_token>" -F "title=my title" -F "message=my message" -F "priority=5" -F "clients[]=<clienttoken_1>"  -F "clients[]=<clienttoken_2>"  -F "clients[]=<clienttoken_n>"

And the messages will only be sent to those clients.

If you do not specify the tokens, the message will be sent to all the clients, like the normal behavior

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -6,7 +6,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
 	"github.com/gotify/location"
@@ -388,6 +387,9 @@ func toInternalMessage(msg *model.MessageExternal) *model.Message {
 	if msg.Extras != nil {
 		res.Extras, _ = json.Marshal(msg.Extras)
 	}
+	if msg.Clients!=nil{
+		res.Clients,_ =  json.Marshal(msg.Clients)
+	}
 	return res
 }
 
@@ -403,6 +405,10 @@ func toExternalMessage(msg *model.Message) *model.MessageExternal {
 	if len(msg.Extras) != 0 {
 		res.Extras = make(map[string]interface{})
 		json.Unmarshal(msg.Extras, &res.Extras)
+	}
+	if len(msg.Clients) != 0 {
+		res.Clients = []string{}
+		json.Unmarshal(msg.Clients, &res.Clients)
 	}
 	return res
 }
--- a/api/stream/stream.go
+++ b/api/stream/stream.go
@@ -62,7 +62,7 @@ func (a *API) NotifyDeletedClient(userID uint, token string) {
 				clients = append(clients[:i], clients[i+1:]...)
 			}
 		}*/
-		if cliente,okey := clients[token]; okey{
+		if client, ok := clients[token]; ok {
 			cliente.Close()
 			delete(clients,token)
 		}
--- a/api/stream/stream.go
+++ b/api/stream/stream.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
 	"github.com/gin-gonic/gin"
 	"github.com/gorilla/websocket"
 	"github.com/gotify/server/auth"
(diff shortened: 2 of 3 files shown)


api/stream/stream.go
```diff
@@ -71,29 +75,42 @@ func (a *API) Notify(userID uint, msg *model.MessageExternal) {
...
+				delete(a.clients[remove.userID],key)
+				//a.clients[remove.userID] = append(userIDClients[:i], userIDClients[i+1:]...)
 				break
 			}
 		}
 	}
 }
 
-func (a *API) register(client *client) {
+func (a *API) register(cliente *client) {
```
eternal-flame-AD: Why rename this variable?

api/stream/stream.go
```diff
@@ -71,29 +75,42 @@ func (a *API) Notify(userID uint, msg *model.MessageExternal) {
...
 func (a *API) remove(remove *client) {
 	a.lock.Lock()
 	defer a.lock.Unlock()
 	if userIDClients, ok := a.clients[remove.userID]; ok {
-		for i, client := range userIDClients {
+		for key, client := range userIDClients {
 			if client == remove {
-				a.clients[remove.userID] = append(userIDClients[:i], userIDClients[i+1:]...)
+				delete(a.clients[remove.userID],key)
+				//a.clients[remove.userID] = append(userIDClients[:i], userIDClients[i+1:]...)
```
eternal-flame-AD: cleanup commented out code pls (same for other occurrences in this PR)

api/message.go
```diff
@@ -388,6 +387,9 @@ func toInternalMessage(msg *model.MessageExternal) *model.Message {
 	if msg.Extras != nil {
 		res.Extras, _ = json.Marshal(msg.Extras)
 	}
+	if msg.Clients!=nil{
```
eternal-flame-AD: Please format your code with a linter (run `make format` and/or `gofmt`).

api/stream/stream.go
```diff
@@ -55,12 +54,17 @@ func (a *API) NotifyDeletedClient(userID uint, token string) {
...
-		for i := len(clients) - 1; i >= 0; i-- {
+		
+		/*for i := len(clients) - 1; i >= 0; i-- {
 			client := clients[i]
 			if client.token == token {
 				client.Close()
 				clients = append(clients[:i], clients[i+1:]...)
 			}
+		}*/
+		if cliente,okey := clients[token]; okey{
```
eternal-flame-AD:
```suggestion
		if client, ok := clients[token]; ok {
```

api/stream/stream.go
```diff
@@ -30,7 +29,7 @@ type API struct {
 // pong command.
 func New(pingPeriod, pongTimeout time.Duration, allowedWebSocketOrigins []string) *API {
 	return &API{
-		clients:     make(map[uint][]*client),
+		clients:     make(map[uint]map[string]*client),
```
eternal-flame-AD: could more than one client be connected with the same token? Your implementation would only allow one to be addressable.

model/message.go
```diff
@@ -62,4 +63,11 @@ type MessageExternal struct {
 	// required: true
 	// example: 2018-02-27T19:36:10.5045044+01:00
 	Date time.Time `json:"date"`
+	// The tokens of the clients.
+	//
+	// read only: true
+	// required: false
+	// example: ["token1","token2","token3"]
+	// if empty, the message is sent to all clients
+	Clients[] string `form:"clients[]" query:"-" json:"clients"`
```
eternal-flame-AD: I don't think its a good idea to refer to the clients by their token as applications are not supposed to know this formation, and it would be inappropriate to send other (possibly admin) client tokens to all clients. I would say use client id instead.

api/message.go
```diff
@@ -404,6 +406,10 @@ func toExternalMessage(msg *model.Message) *model.MessageExternal {
 		res.Extras = make(map[string]interface{})
 		json.Unmarshal(msg.Extras, &res.Extras)
 	}
+	if len(msg.Clients) != 0 {
+		res.Clients = []string{}
+		json.Unmarshal(msg.Clients, &res.Clients)
```
eternal-flame-AD: I don't think we should provide this information to all clients

api/stream/stream.go
Zenteno: because when i use the variable as client, the line 111 sent me an error, maybe the compiler got confused with the structure and the variable

Zenteno:
,

PR #228, opened 2019-10-12
Auto reconnect on connection loss

### LINKED ISSUE #226: Automatically reconnect
Currently, when connection to backend is lost, be it due to an update of the container, network outage or something else, the frontend will wait for the user to manually click the retry button.

It would make sense to automatically retry at an interval without any user interaction.

foxbit19:
I've used react hooks to add a side effect into `NetworkLostBanner` component.

The function sets a timer (in `componentDidMount`) that every 3 seconds tries to call the `retry` function. When the component is unmounted, the `clearInterval` is called (in `componentWillUnmount`) to cancel the timed action.

This fix improve server to manage connection loss status (see #226).

Fixes #226

### DIFF
--- a/ui/src/CurrentUser.ts
+++ b/ui/src/CurrentUser.ts
@@ -127,7 +127,7 @@ export class CurrentUser {
             .then(() => this.snack('Password changed'));
     };
 
-    public tryReconnect = (quiet: boolean = false) => {
+    public tryReconnect = (quiet = false) => {
         this.tryAuthenticate().catch(() => {
             if (!quiet) {
                 this.snack('Reconnect failed');
--- a/ui/src/CurrentUser.ts
+++ b/ui/src/CurrentUser.ts
@@ -127,7 +127,7 @@ export class CurrentUser {
             .then(() => this.snack('Password changed'));
     };
 
-    public tryReconnect = (quiet = false) => {
+    public tryReconnect = (quiet: boolean = false) => {
         this.tryAuthenticate().catch(() => {
             if (!quiet) {
                 this.snack('Reconnect failed');
--- a/ui/src/CurrentUser.ts
+++ b/ui/src/CurrentUser.ts
@@ -10,6 +10,8 @@ const tokenKey = 'gotify-login-key';
 
 export class CurrentUser {
     private tokenCache: string | null = null;
+    private reconnectTimeoutId: number | null = null;
+    private readonly reconnectTime: number = 3000;
     @observable
     public loggedIn = false;
     @observable
@@ -90,7 +92,7 @@ export class CurrentUser {
             })
             .catch((error: AxiosError) => {
                 if (!error || !error.response) {
-                    this.hasNetwork = false;
+                    this.lostNetwork();
                     return Promise.reject(error);
                 }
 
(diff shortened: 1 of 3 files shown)


eternal-flame-AD:
@jmattheis Would u pls review this? thx

ui/src/common/NetworkLostBanner.tsx
```diff
@@ -8,6 +8,14 @@ interface NetworkLostBannerProps {
 }
 
 export const NetworkLostBanner = ({height, retry}: NetworkLostBannerProps) => {
+    useEffect(() => {
```
jmattheis: For state and logic we use mobx stores in our ui. I'd prefer having the reconnect logic there to.

It could be like this:
`CurrentUser.ts`
```typescript
private reconnectTimeoutId: number | null = null;

// other methods

public tryReconnect = (quiet = false) => {
    this.tryAuthenticate().catch(() => {
        if (!quiet) {
            this.snack('Reconnect failed');
        }
    });
};

private lostNetwork = () => {
    this.hasNetwork = false;
    if (this.reconnectTimeoutId !== null) {
        window.clearTimeout(this.reconnectTimeoutId);
    }
    this.reconnectTimeoutId = window.setInterval(() => this.tryReconnect(true), 3000);
};
```
Then replace `this.hasNetwork = false` with `this.lostNetwork()`.

Like this, logic stays inside the stores and our react components stay dumb.

In `Layout.tsx`, the `doReconnect`-method and the `reconnecting`-property should be deleted. And the `CurrentUser#tryReconnect` method should be forwarded to the `NetworkLostBanner`.

foxbit19: Thanks for you hexaustive feedback!
I'm agree with the suggested logic: leave reconnect inside the stores and make dumb components.
I've made all the changes and I've added another `private` variable to `CurrentUser.ts`:

```ts
private readonly reconnectTime: number = 3000;
````

This constant (create with `readonly` ts keyword) could be useful to calibrate the amount of time of our timeout object.

Let me know if there are other suggestions!

File: ui/src/layout/Layout.tsx [hunk 2]
```diff
@@ -112,9 +98,7 @@ class Layout extends React.Component<
             <MuiThemeProvider theme={theme}>
                 <HashRouter>
                     <div>
-                        {hasNetwork ? null : (
-                            <NetworkLostBanner height={64} retry={this.doReconnect} />
-                        )}
+                        {hasNetwork ? null : <NetworkLostBanner height={64} retry={tryReconnect} />}
```
jmattheis: No notification will be displayed when the retry button on the NetworkLostBanner is clicked. It seems like the default value for quiet doesn't work (I dunno why). I got it working by explicitly passing `false` to the tryAgain method.
```suggestion
                        {hasNetwork ? null : <NetworkLostBanner height={64} retry={() => tryReconnect(false)} />}
```

ui/src/common/NetworkLostBanner.tsx
jmattheis: Looks pretty good! I only have one small remark left: https://github.com/gotify/server/pull/228#discussion_r334276230

ui/src/layout/Layout.tsx
foxbit19: Not really necessary to explicit the argument on `tryReconnect` function. I've added argument type and called the function without arguments. Now it seems to work.

foxbit19: And I fixed the formatting with `yarn format` :)

ui/src/CurrentUser.ts
```diff
@@ -124,4 +126,23 @@ export class CurrentUser {
             .post(config.get('url') + 'current/user/password', {pass})
             .then(() => this.snack('Password changed'));
     };
+
+    public tryReconnect = (quiet: boolean = false) => {
```
jmattheis: The boolean type is not needed. tslint complains about that https://travis-ci.org/gotify/server/builds/597248630#L1798

foxbit19: Yeah, I've made a mistake about that.

PR #231, opened 2019-10-16
Fix automatic reconnect, add exponential backoff

quthla:
This should fix automatic reconnect when reverse proxy returns 503.
Additionally, it adds exponential backoff with a maximum of 2 mins.

@jmattheis Please let me know if this works.

### DIFF
--- a/ui/src/CurrentUser.ts
+++ b/ui/src/CurrentUser.ts
@@ -97,7 +97,7 @@ export class CurrentUser {
                     return Promise.reject(error);
                 }
 
-                if (error.response.status === 503) {
+                if (error.response.status >= 500) {
                     this.connectionError(`${error.response.statusText} (code: ${error.response.status}).`);
                     return Promise.reject(error);
                 }
--- a/ui/src/CurrentUser.ts
+++ b/ui/src/CurrentUser.ts
@@ -93,7 +93,7 @@ export class CurrentUser {
             })
             .catch((error: AxiosError) => {
                 if (!error || !error.response) {
-                    this.connectionError('No network connection.');
+                    this.connectionError('No network connection or server unavailable.');
                     return Promise.reject(error);
                 }
 
--- a/ui/src/CurrentUser.ts
+++ b/ui/src/CurrentUser.ts
@@ -11,7 +11,7 @@ const tokenKey = 'gotify-login-key';
 export class CurrentUser {
     private tokenCache: string | null = null;
     private reconnectTimeoutId: number | null = null;
-    private readonly reconnectTime: number = 3000;
+    private reconnectTime = 7500;
     @observable
     public loggedIn = false;
     @observable
@@ -19,7 +19,7 @@ export class CurrentUser {
     @observable
     public user: IUser = {name: 'unknown', admin: false, id: -1};
     @observable
-    public hasNetwork = true;
+    public connectionErrorMessage: string | null = null;
 
     public constructor(private readonly snack: SnackReporter) {}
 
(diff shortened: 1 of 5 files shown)


quthla:
```ERROR: /home/travis/gopath/src/github.com/gotify/server/ui/src/CurrentUser.ts:14:28 - Type number trivially inferred from a number literal, remove type annotation```

How is this erroring now? Type was annotated already before

File: ui/src/CurrentUser.ts [hunk 1]
```diff
@@ -144,5 +145,6 @@ export class CurrentUser {
             () => this.tryReconnect(true),
             this.reconnectTime
         );
+        this.reconnectTime = Math.min(this.reconnectTime * 2, 120000);
```
jmattheis: I think we do not need this, there is no real cost in sending the requests more frequent..

ui/src/CurrentUser.ts
```diff
@@ -88,10 +88,11 @@ export class CurrentUser {
                 this.user = passThrough.data;
                 this.loggedIn = true;
                 this.hasNetwork = true;
+                this.reconnectTime = 7500;
                 return passThrough;
             })
             .catch((error: AxiosError) => {
-                if (!error || !error.response) {
+                if (!error || !error.response || error.response.status === 503) {
```
jmattheis: 503 doesn't mean that there is no network connection. So `this.lostNetwork()` is a little misleading here.

I'd say, we rename the variable `CurrentUserStore#hasNetwork` to `connectionError: null | string`. 
`null` means no error -> no banner and if an error is set then the banner will be shown with the message.
503 could be something like `The server couldn't successfully handle our request, (statusCode: 503).`.

Both this.lostNetwork and the NetworkLostBanner need a new name as it is not exclusive to  network lost errors anymore.

quthla:
Yeah sounds good. Feel free to adjust the PR according to your proposal. I've enabled edits by maintainers

ui/src/CurrentUser.ts
quthla: It's just a common pattern for stuff like this but I'm open to your proposal. Considering gotify is offline for a while that'd mean 1200 req/h for nothing

quthla:
@jmattheis have you decided yet how you want to proceed with this?

jmattheis:
If you don't want todo it, then I'll maybe do it next week. I'm currently a little busy.

quthla:
As I said I'm not really familiar with ts and don't have a dev environment but I can try

quthla:
Seems something is broken but apart from that, this should be what you had in mind, right?

ui/src/CurrentUser.ts
```diff
@@ -87,16 +87,22 @@ export class CurrentUser {
...
             })
             .catch((error: AxiosError) => {
                 if (!error || !error.response) {
-                    this.lostNetwork();
+                    this.connectionError('No network connection.');
                     return Promise.reject(error);
                 }
 
-                this.hasNetwork = true;
+                if (error.response.status === 503) {
```
jmattheis: 503 is only service unavailable, we should include all 500 errors.
```suggestion
                if (error.response.status >= 500) {
```

ui/src/user/Login.tsx
```diff
@@ -44,7 +44,7 @@ class Login extends Component<Stores<'currentUser'>> {
                                 size="large"
                                 className="login"
                                 color="primary"
-                                disabled={!this.props.currentUser.hasNetwork}
+                                disabled={this.props.currentUser.connectionErrorMessage}
```
jmattheis: disable expects a boolean, you can double negate the string to get the boolean value
```suggestion
                                disabled={!!this.props.currentUser.connectionErrorMessage}
```
(this is the reason why the build fails)

quthla:
> Looks good, three little remarks.
> 
> The build fails because of formatting you've to run
> 
> ```
> $ yarn format
> ```
> 
> to fix it.

As mentioned before I unfortunately don't have a dev environment set up but tried as good as possible so maybe you can fix these last small issues if you find time.

jmattheis:
@quthla Alright, will probably do it tomorrow or on thursday.

quthla:
Can you make a new release maybe?

jmattheis:
@quthla sure -> https://github.com/gotify/server/releases/tag/v2.0.11

quthla:
Thanks

Issue #171, opened 2019-03-17
Add refresh button in WebUI message list


pm-pm:
If you delete a message on/via phone it does not disappear on the ui automatically.

First this, works well:
```
1. Open web ui
2. Send message via CLI
3. Message is pushed to mobile phone and instantly visible in web ui. Works as aspected.
```

Now the non working part:
```
1. Leave web ui open!
2. Delete message on/via phone
3. Message deleted on phone, but NOT disappear in the web ui.
```
If you click in web ui between different "APPS" or switch "CLIENTS <-> APPS" in menu the already deleted message is still visible in web ui.
You only can "clear" (make disappear) the message from the web ui if you use the browser reload button or logout and login.

Is it possible to make the message disappear automatically on web ui if deleted on/via phone?

eternal-flame-AD:
Yes, currently WebSocket only notifies clients of new messages, but not message deletion, read/unread( #21 related), etc. But the current WebSocket message scheme does not include event types, thus it would break the API to fix the problem.

pm-pm:
Ok, i understand that is a bigger thing.
I do NOT need this funktion very often, more for testing, so i can do without this function.

Two ideas for a workaround (I prefer the second one):
- Reload the messages when you click/switch on the left side from one App to an other App
or
- Add a reload button on top. See picture.

![rms](https://user-images.githubusercontent.com/47995430/54506101-24c06f00-493b-11e9-8125-59e1fd362958.png)

Using the browser reload button is not so good.

eternal-flame-AD:
I would say let's add a refresh button in the messages list. What do you think? @jmattheis

jmattheis:
A refresh button is okay.

foxbit19:
Hi, what about this issue?
I had a look at the code. I suggest to do some refactor tasks on `DefaultPage` component to support more than one button at the top fo the page.

What do you think?

jmattheis:
Yeah seems good, the default page could accept a react node then it would be more customizable
```
<DefaultPage title="" maxWidth={250} rightControl={<Button onClick={...}>Cool button</Button>}>
```
Another option would be a [Fab](https://material-ui.com/components/buttons/#floating-action-buttons) button, sadly we already have one which scrolls to the top.

foxbit19:
`DefaultPage` + react node seems good! I would discard `Fab` button for now.

pm-pm:
Top! :-) Great work.

Thanks.

blabasodfiu:
[...]

PR #233, opened 2019-10-18
WIP: Update DefaultPage to be more general and add refresh button in web UI

foxbit19:
I've updated `DefaultPage` component to be more general following @jmattheis suggestions (see [here](https://github.com/gotify/server/issues/171#issuecomment-541792737)).

Now the component accepts a new argument `rightControl` useful to specify a react node placed on the top right corner of the page content.
Old specific purpose arguments `buttonTitle`, `fButton`, `buttonDisabled`, `hideButton` and `buttonId` were removed.

Using the improved `DefaultPage`, we can easely solve the issue #171.

### DIFF
--- a/ui/src/common/DefaultPage.tsx
+++ b/ui/src/common/DefaultPage.tsx
@@ -15,7 +15,7 @@ const DefaultPage: SFC<IProps> = ({title, rightControl, maxWidth = 700, children
                 <Typography variant="h4" style={{flex: 1}}>
                     {title}
                 </Typography>
-                {rightControl ? rightControl : null}
+                {rightControl}
             </Grid>
             {children}
         </Grid>
--- a/ui/src/common/DefaultPage.tsx
+++ b/ui/src/common/DefaultPage.tsx
@@ -1,44 +1,22 @@
-import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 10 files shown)


ui/src/message/MessagesStore.ts
```diff
@@ -100,6 +100,13 @@ export class MessagesStore {
         this.createEmptyStatesForApps(this.appStore.getItems());
     };
 
+    @action
+    public refreshByApp = async (appId: number) => {
+        const state = this.stateOf(appId);
+        const result = await this.fetchMessages(appId, 0).then((resp) => resp.data);
+        state.messages.replace(result.messages);
```
jmattheis: I think we should refresh all messages because when refreshing only one application the "All Messages"-Tab will not get updated resulting in inconsistent state.
So we can use `this.clearAll()` to discard our state and `this.loadMore()` to get the most recent messages.

ui/src/common/DefaultPage.tsx
```diff
@@ -1,44 +1,21 @@
...
 import Grid from '@material-ui/core/Grid';
 import Typography from '@material-ui/core/Typography';
 import React, {SFC} from 'react';
 
 interface IProps {
     title: string;
-    buttonTitle?: string;
-    fButton?: VoidFunction;
-    buttonDisabled?: boolean;
+    rightControl?: object;
```
jmattheis: This can probably be a React.ReactNode

Issue #159, opened 2019-03-12
Link to the main page


joy4eg:
Hey there, thanks for the great app!

From my point of view, clicking on `gotify` should redirect to the main page `/`, and not to github.
(Yes, I saw `all messages`, but from UX side it's a bit confusing)
We can add a separate button to the header/footer like `about` which will do the same thing :)

![image](https://user-images.githubusercontent.com/878385/54201332-08a17580-44d6-11e9-8f39-873a9e83fe41.png)

jmattheis:
I don't know, I currently see no problem here. @eternal-flame-AD what do you think?

eternal-flame-AD:
I think this can be improved, as I also sometimes make this mistake, but not a priority though. I would say let's keep it and see if anyone can submit a PR for this.

PR #234, opened 2019-10-19
Set main page link and add GitHub icon

foxbit19:
I've replaced link on Gotify title with `/` and I've added a new icon on header right hand side.

gotify_github

Due the version of `material-ui/icons`, an icon for GitHub reference isn't available. So, I added a new `common` component `GitHubIcon`. This one will need be removed in future releases (when `material-ui/icons` package will updated to at least version `4.5`).

This is my proposal to solve #159.


ui/src/common/GitHubIcon.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import SvgIcon from '@material-ui/core/SvgIcon';
+
+export default function GitHubIcon() {
```
jmattheis: If this Icon is available in a newer version feel free to update it :D.

ui/src/layout/Header.tsx
```diff
@@ -55,7 +56,7 @@ class Header extends Component<IProps> {
             <AppBar position="absolute" style={style} className={classes.appBar}>
                 <Toolbar>
                     <div className={classes.title}>
-                        <a href="https://github.com/gotify/server" className={classes.link}>
+                        <a href="/" className={classes.link}>
```
jmattheis: I think this does a wrong redirect when gotify is served under a subpath (See f.ex https://gotify.net//docs/nginx#at-a-subpath). This can be fixed by using `<Link>` from `react-router-dom`.

See f.ex the "All Messages"-Link
https://github.com/gotify/server/blob/6edfd8400e58cbf51dda76d9e06ccac0471f1c71/ui/src/layout/Navigation.tsx#L68-L72

ui/src/common/GitHubIcon.tsx
foxbit19: Ok, thanks :+1:

ui/src/layout/Header.tsx
foxbit19: Yeah, I think it's better for a correct routing.

PR #242, opened 2019-11-26
Fix websocket reconnect logic

quthla:
The code is obsolete now that there is the improved auto reconnect function

It also did never actually work anyway. [This](https://github.com/gotify/server/blob/master/ui/src/message/WebSocketStore.ts#L41) will call `listen` with an undefined callback, thus leading to the following error and effectively preventing the client from reconnecting automatically.

`TypeError: e is not a function WebSocketStore.ts:33:33`

### DIFF
--- a/ui/src/message/WebSocketStore.ts
+++ b/ui/src/message/WebSocketStore.ts
@@ -36,10 +36,6 @@ export class WebSocketStore {
             this.wsActive = false;
             this.currentUser
                 .tryAuthenticate()
-                .then(() => {
-                    this.snack('WebSocket connection closed, trying again in 30 seconds.');
-                    setTimeout(this.listen, 30000);
-                })
                 .catch((error: AxiosError) => {
                     if (error && error.response && error.response.status === 401) {
                         this.snack('Could not authenticate with client token, logging out.');
--- a/ui/src/message/WebSocketStore.ts
+++ b/ui/src/message/WebSocketStore.ts
@@ -34,11 +34,17 @@ export class WebSocketStore {
 
         ws.onclose = () => {
             this.wsActive = false;
-            this.currentUser.tryAuthenticate().catch((error: AxiosError) => {
-                if (error && error.response && error.response.status === 401) {
-                    this.snack('Could not authenticate with client token, logging out.');
-                }
-            });
+            this.currentUser
+                .tryAuthenticate()
+                .then(() => {
+                    this.snack('WebSocket connection closed, trying again in 30 seconds.');
+                    setTimeout(() => this.listen(callback), 30000);
+                })
+                .catch((error: AxiosError) => {
+                    if (error && error.response && error.response.status === 401) {
+                        this.snack('Could not authenticate with client token, logging out.');
+                    }
+                });
         };
 
         this.ws = ws;
--- a/ui/src/message/WebSocketStore.ts
+++ b/ui/src/message/WebSocketStore.ts
@@ -34,13 +34,11 @@ export class WebSocketStore {
 
         ws.onclose = () => {
             this.wsActive = false;
-            this.currentUser
-                .tryAuthenticate()
-                .catch((error: AxiosError) => {
-                    if (error && error.response && error.response.status === 401) {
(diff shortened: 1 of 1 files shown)


quthla:
It will still call tryAuthenticate on close which fires the reconnect logic so that's not an issue and the code is not needed anymore

jmattheis:
Are you sure? I think the reconnect logic (CurrentUser.connectionError) will only be executed if there was an error the first place (tryAuthenticate() will resolve successfully). In this case, only the web socket closed the rest of the state doesn't know of this and still thinks everything is fine.

quthla:
> 
> 
> Are you sure? I think the reconnect logic (CurrentUser.connectionError) will only be executed if there was an error the first place (tryAuthenticate() will resolve successfully). In this case, only the web socket closed the rest of the state doesn't know of this and still thinks everything is fine.

Honestly, just the fact that there was no report ever about this code being broken makes me think this was never an issue in the first place.

jmattheis:
It's not an obvious error, normal users doesn't look in the console for errors, so we don't know how frequent the bug occurs.

quthla:
Well, if you want to keep it, then that's ok with me. After all, you know the backend and how all is tied together better than me.

I've restored and fixed the old code. Could you run prettier on it in case check fails?

quthla:
Can you make a new release? Maybe gotify will then finally reliably reconnect for me 😂

jmattheis:
Do you use a reverse proxy? Gotify only closes the websocket connection if the user/client gets deleted, ever other close should not happen.

quthla:
I think it's more of a network issue which rarely occurs but yes I'm using a reverse proxy.

jmattheis:
Could you check the gotify log if there is something regarding websockets? Otherwise, it's most likely the fault of the reverse proxy. But, I'll create a new release after fixing #243

quthla:
No results grepping for websocket. Grepping error yields a few hundred lines like the following

`gotify    | Error #01: you need to provide a valid access token or user credentials to access this api`

jmattheis:
That seems normal, I'll add some better error messages for websockets. With https://github.com/gotify/server/pull/244/commits/d764a8db427b633c608a7814e19e3bad9ae67286 messages like `WebSocket: ReadError websocket: close 1006 (abnormal closure): unexpected EOF` will be logged. This hopefully makes it clear, on which end the error occurs.

jmattheis:
@quthla v2.0.12 is released https://github.com/gotify/server/releases/tag/v2.0.12

quthla:
@jmattheis 

```gotify    | WebSocket: ReadError read tcp 172.17.0.3:80->172.17.0.1:52202: i/o timeout
gotify    | WebSocket: ReadError read tcp 172.17.0.3:80->172.17.0.1:52362: i/o timeout
gotify    | WebSocket: ReadError read tcp 172.17.0.3:80->172.17.0.1:52378: i/o timeout
gotify    | Error #01: you need to provide a valid access token or user credentials to access this api
gotify    | Error #01: you need to provide a valid access token or user credentials to access this api
gotify    | WebSocket: ReadError websocket: close 1006 (abnormal closure): unexpected EOF
gotify    | WebSocket: ReadError websocket: close 1006 (abnormal closure): unexpected EOF
gotify    | Error #01: you need to provide a valid access token or user credentials to access this api
gotify    | WebSocket: ReadError websocket: close 1006 (abnormal closure): unexpected EOF
gotify    | Error #01: you need to provide a valid access token or user credentials to access this api
gotify    | WebSocket: ReadError websocket: close 1006 (abnormal closure): unexpected EOF
gotify    | Error #01: you need to provide a valid access token or user credentials to access this api
```

jmattheis:
@quthla Could you open a new issue for this and don't forget to include your nginx config.

quthla:
I'm using Apache. So you suspect this to be an issue with gotify or why should I open a new issue?

jmattheis:
Most likely not, but without the config I can't say it for sure. I don't want to discuss this issue in a pull request, therefore you've to create a new issue for it.

PR #252, opened 2019-12-29
Expose port 80 by default in the Dockerfiles

### LINKED ISSUE #251: Expose the default port 80 in the Docker images
Yes, with the latest Docker image (x86_64 tested, but I see the same issue in the armv7 image as well).

- [x] Docker
- [ ] Linux machine
- [ ] Windows machine

```
version: '3.7'

services:
  gotify:
    container_name: gotify
    image: gotify/server
    restart: unless-stopped
    init: true
    volumes:
      - gotify_data:/app/data
    environment:
      - VIRTUAL_HOST=gotify.example.com
      - VIRTUAL_PORT=80
      - LETSENCRYPT_HOST=gotify.example.com
    networks:
      - proxy_net

volumes:
  gotify_data:
```
</details>

- [ ] None
- [x] Nginx
- [ ] Apache
- [ ] Caddy

```
version: '3.7'

services:
  nginx_proxy:
    container_name: nginx-proxy
    image: jwilder/nginx-proxy:alpine
    restart: unless-stopped
    init: true
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - proxy_certs:/etc/nginx/certs:ro
      - proxy_vhostd:/etc/nginx/vhost.d:ro
      - proxy_html:/usr/share/nginx/html:ro
      - /var/run/docker.sock:/tmp/docker.sock:ro
    networks:
      - proxy_net

  letsencrypt_nginx_proxy_companion:
    container_name: letsencrypt-nginx-proxy-companion
    image: jrcs/letsencrypt-nginx-proxy-companion
    restart: unless-stopped
    init: true
    volumes:
      - proxy_certs:/etc/nginx/certs
      - proxy_vhostd:/etc/nginx/vhost.d
      - proxy_html:/usr/share/nginx/html
      - /var/run/docker.sock:/var/run/docker.sock:ro
    environment:
      - NGINX_PROXY_CONTAINER=nginx-proxy
      
volumes:
  proxy_certs:
  proxy_vhostd:
  proxy_html:

networks:
  proxy_net:
    driver: bridge
    driver_opts:
      com.docker.network.bridge.name: proxy-net
```
</details>

- [x] WebUI
- [ ] gotify-cli
- [ ] Android Client 

**What did you do?**
I'm attempting to run the gotify server through the [jwilder/nginx-proxy](https://github.com/jwilder/nginx-proxy) reverse proxy with Let's Encrypt provided via [jrcs/letsencrypt-nginx-proxy-companion](https://github.com/JrCs/docker-letsencrypt-nginx-proxy-companion) using `docker-compose`.

The gotify Web UI to be accessible and SSL-secured via the proxy.

Error 502 from the proxy due to the gotify server Docker image not exposing any ports. If I manually make the following Docker image the server becomes accessible without issues:
```
FROM gotify/server

# Expose the default port 80
EXPOSE 80
```

Would it be possible to add the `EXPOSE` directive to the gotify server Dockerfile? I have 10+ other containers running via the same proxy without issues, as they all `EXPOSE` by default. I can do a small PR that adds the line.

(There's also a cleaner way to add the `ca-certificates` package in the Dockerfiles, I can improve that as well in the PR if wanted.)

twelho:
Fixes #251. Adds the `EXPOSE 80` directive to both the `amd64` and `armv7` Dockerfiles enabling seamless usage with [jwilder/nginx-proxy](https://github.com/jwilder/nginx-proxy). Also makes the `ca-cerificates` installation line more readable.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM frolvlad/alpine-glibc:glibc-2.28
 WORKDIR /app
-RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
+RUN apk add --no-cache ca-certificates
 ADD gotify-app /app/
 EXPOSE 80
 ENTRYPOINT ["./gotify-app"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,4 +2,5 @@ FROM frolvlad/alpine-glibc:glibc-2.28
 WORKDIR /app
 RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
 ADD gotify-app /app/
+EXPOSE 80
 ENTRYPOINT ["./gotify-app"]
--- a/docker/Dockerfile.armv7
+++ b/docker/Dockerfile.armv7
@@ -1,4 +1,5 @@
 FROM arm32v7/debian
 WORKDIR /app
 ADD gotify-app /app/
+EXPOSE 80
 ENTRYPOINT ["./gotify-app"]


twelho:
Should I also include port 443 in the `EXPOSE` definition? It might be useful to advertise it as well especially since HTTPS autoredirection seems to be enabled by default according to the configuration docs.

jmattheis:
I think the docker image is mostly used with an reverse proxy like nginx/traefik, so I'd say, exposing 443 isn't needed. Gotify will not listen to port 443 when `server.ssl.enabled == false` which is the default. Therefore it could actually be misleading to define that it listens on 443.

>  It might be useful to advertise it as well especially since HTTPS autoredirection seems to be enabled by default according to the configuration docs.

The redirect will only be done, when ssl is enabled.

PR #267, opened 2020-02-09
Add custom notification sound for better support on Firefox

### LINKED ISSUE #263: Sound for browser notifications
In #29, notifications for the web UI were added. Since these are quiet, they can be easily missed.

Hence, I'd like to play a sound for new notifications as is already done by Telegram Web and WhatsApp Web clients. We could simply pick a default sound file, and play it using [this snippet](https://stackoverflow.com/a/39285990) given on Stack Overflow.

I'd like to help implementing this, but first I wanted to make sure that this feature would get merged, and ask if any preferences concerning the sound file exist.

eikendev:
This pull request fixes #263.

We
- add a custom notification sound, licensed under CC0,
- set a priority threshold to only play the sound for *important* notifications,
- and disable default notification sounds privided by other browsers.


eikendev:
I've proposed a change to the type definitions for `INotifyOption` at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42232. Once that is merged we can get rid of the cast to `any`.

ui/src/snack/browserNotification.ts
```diff
@@ -19,10 +19,17 @@ export function notifyNewMessage(msg: IMessage) {
...
         icon: msg.image,
+        silent: true,
         notifyClick: closeAndFocus,
         notifyShow: closeAfterTimeout,
-    });
+    } as any);
     notify.show();
+
+    if (msg.priority >= 4 && !Notify.needsPermission) {
+        const let src = 'static/notification.ogg';
```
jmattheis: Compile error
```suggestion
        const src = 'static/notification.ogg';
```

ui/src/snack/browserNotification.ts
```diff
@@ -19,10 +19,17 @@ export function notifyNewMessage(msg: IMessage) {
...
+        const let audio = new Audio(src);
```
jmattheis: Compile error
```suggestion
        const audio = new Audio(src);
```

ui/src/snack/browserNotification.ts
```diff
@@ -19,10 +19,17 @@ export function notifyNewMessage(msg: IMessage) {
     const notify = new Notify(msg.title, {
         body: removeMarkdown(msg.message),
         icon: msg.image,
+        silent: true,
```
jmattheis: The pr was merged, could you reference the newer version of the dependency?

ui/src/snack/browserNotification.ts
```diff
@@ -19,10 +19,17 @@ export function notifyNewMessage(msg: IMessage) {
...
         notifyClick: closeAndFocus,
         notifyShow: closeAfterTimeout,
-    });
+    } as any);
     notify.show();
+
+    if (msg.priority >= 4 && !Notify.needsPermission) {
```
jmattheis: I think the audio should be played regardless if the user allows notifications. Furthermore, I think the audio should be started inside the reaction.ts
```typescript
stores.wsStore.listen((message) => {
    stores.messagesStore.publishSingleMessage(message);
    Notifications.notifyNewMessage(message);
    if (message.priority >= 4) {
        const src = 'static/notification.ogg';
        const audio = new Audio(src);
        audio.play();
    }
});
```

eikendev:
Sorry, I was sure that I tested d954baa9535e3202ca5c5285fb7ce74b2e10f254, turns out I was wrong. I've now upgraded `@types/notifyjs` to the latest version and removed the cast. Also, I moved the block into `reactions.ts` as suggested.

PR #268, opened 2020-02-10
Add .editorconfig file

eikendev:
I thought it might be a good idea to introduce an [EditorConfig](https://editorconfig.org/) file.

The source code seems quite consistent, but different file types have different indentation styles. I tried to write this file to match existing conventions in this repository, so that new contributions are more likely to integrate with the code style-wise.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,28 @@
+root = true
+
+[*]
+charset = utf-8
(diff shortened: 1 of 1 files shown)


eikendev:
@jmattheis Do you have any preferences regarding the lengths of lines?

PR #270, opened 2020-02-11
Introduce event types for WebSocket connection

### LINKED ISSUE #266: Synchronize notification state among clients
Currently, notifications that are read on one client are not automatically read on other clients. This means, when working on a PC and reading the notifications, the mobile phone would end up showing a collection of messages that have already been read.

A change was proposed in #171 to introduce a new `type` property to messages and let clients subscribe to new messages only. Although this would probably work, it's a bit hacky in my view.

I propose that we introduce a new API, that supports different events. To make Gotify backwards compatible with "old" clients, the current `/stream` endpoint will be enhanced by a GET parameter that specifies an API version, like `/stream?version=1`. The default API (if no version is specified), for now version 1, will support the current messages only. The clients will receive the same message format as right now, so that "old" clients do not break.

The new notification system is then reachable under `/stream?version=2`, and will push the types `new_message` and `read_message`.

This gives us more flexibility for future changes. If breaking changes occur, we can decide if older versions are still worth supporting, and drop support for versions we no longer see value in supporting.

I'd like to implement this in the following days, so please tell me if these changes are acceptable.

eikendev:
This pull request fixes #266.

When a client connects to the `/stream` endpoint and sets `version=2` as a GET parameter, they will receive events of the following types.
```json
{
	"type": "message",
	"content": {
		"id": 58,
		"appid": 1,
		"message": "a message",
		"title": "a test",
		"priority": 5,
		"date": "2020-02-11T23:03:02.852206171+01:00"
	}
}
```
```json
{
	"type": "message_deletions",
	"content": {
		"ids": [
			58
		]
	}
}
```

For version 2, the two event types `message` and `message_deletions` are implemented.

For old clients (not specifying a `version` parameter, or setting `version=1`), the format remains unchanged. They will only receive messages containing new notifications.

Note that I have not adapted the tests yet. Feedback is very welcome.


jmattheis:
Hey there, thanks for the PR :+1:, I'm currently a little busy and I want to take time to think over this addition as it may have long lasting effects. I try to review it this weekend.

eikendev:
@jmattheis I cannot get the Swagger documentation to generate correctly. It would always end up with a wrong `x-go-package` member. Any idea how I can fix this?

api/stream/client.go
```diff
@@ -87,13 +90,22 @@ func (c *client) startWriteHandler(pingPeriod time.Duration) {
 
 	for {
 		select {
-		case message, ok := <-c.write:
+		case event, ok := <-c.write:
```
jmattheis: Just a thought, maybe the event could be
```
type WebSocketEvent interface {
    ToExternal(apiVersion int) (interface{}, error)
}
```
With this, we can basically omit the event encoder and just let our events handle how v1 and v2 is structured.

model/application.go
```diff
@@ -40,6 +48,52 @@ type Application struct {
 	// read only: true
 	// required: true
 	// example: image/image.jpeg
-	Image    string            `gorm:"type:text" json:"image"`
-	Messages []MessageExternal `json:"-"`
+	Image    string               `gorm:"type:text" json:"image"`
+	Messages []ApplicationMessage `json:"-"`
+}
+
+// ApplicationMessage Model
```
jmattheis: Not really into the code yet, but why do we need an ApplicationMessage besides the normal Message?

model/eventencoder/v2.go
```diff
@@ -0,0 +1,56 @@
...
+	// read only: true
+	// required: true
+	// example: message_deletion
+	Type string `json:"type"`
+	// The event content.
+	//
+	// read only: true
+	// required: true
+	// example: [14,15,16]
+	Content interface{} `json:"content"`
```
jmattheis: Maybe the content of the events should be included in the parent object. Then we could use the discriminator feature from openapi v3 (https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ see discriminator).

So the message would look like this:
```
{ 
    type: "message",
    message: "my content",
    title: "...",
    ...
}
```

This would allow us to f.ex. generate the java client with proper jackson bindings for serializing the message into the right event type.

docs/spec.json
```diff
@@ -2395,4 +2494,4 @@
       "in": "query"
     }
   }
-}
\ No newline at end of file
+}
```
jmattheis: The travis ci build probably doesn't like this newline

api/stream/client.go
eikendev: So the encoders have different ideas of how an event is to be encoded. Since I wanted all WebSocket messages to be reflected certain structs in the code, all event encoders maintain different structs for the events. I had this previously implemented as you said, but noticed it would be a lot cleaner to have a file per encoder, so the structs (which represent the JSON encoding) could be separated. I felt it made more sense this way, since the files separate the "ideas" on how to encode events.
Also, once a "version" is no longer needed (assuming clients will adopt the new version), we can simple remove one file and remove the mapping to it.

model/application.go
eikendev: I saw that the `MessageExternal` struct was used for two different things, namely
- encoding of messages _sent_ to a client,
- encoding of messages _received_ from an application.

Since these things are fundamentally different, I felt the need to create a dedicated struct for the latter. I can remove it if you whish it to be excluded from this pull request. However, since the `CreateMessage` is a bit cleaner by using internal messages, I thought it made sense to include it here.

Also, note the commit message for 20a4510ee7e95bc0bd196b7979fb47da0ac15c9c:
> We introduce the struct ApplicationMessage that is used for receiving new messages from applications. Since applications cannot specify a message ID or date, this ensures that the message sent from applications does not falsely include such members.

In short, the `ApplicationMessage` does not have members that are not accepted in messages sent by applications. Previously, we accepted these fields, but omitted them while processing the message.

docs/spec.json
eikendev: Hm, the Swagger specification check fails locally, too. I'm not sure, but the `x-go-package` in the specification seems to be generated wrongly to me. It starts with a `_` character, not with the `github.com` domain.

model/eventencoder/v2.go
eikendev: This sounds promising indeed. I'd like to first have the Swagger specification pass the tests, so I can adjust the specification from there. Any help getting the Travis CI test pass is very appreciated.

docs/spec.json
jmattheis: Yeah, you can ignore these changes and just remove the newline here and the spec should generate successfully. I'm guessing that travis builds gotify in $GOPATH and thus has there packages with github.com.

model/eventencoder/v2.go
jmattheis: Sadly, we currently use openapi v2 /swagger 2 so the change to openapi v3 would probably done in a separate pr.

model/application.go
jmattheis: Alright, Maybe we name it InputMessage or something similar, that it's clear what it does?

docs/spec.json
jmattheis: Here a successful build with your branch https://travis-ci.org/gotify/server/builds/651646654 with this change https://github.com/gotify/server/commit/138609f119d0f928fa5616b97440e2c9c0546a4e

model/eventencoder/v2.go
eikendev: Isn't [this](https://goswagger.io/use/spec/discriminated.html) exactly what we need?

dcousens:
Why overload the `/stream` API instead of encoding the version in the URL, akin to `/2/stream`?

model/eventencoder/v2.go
jmattheis: Yeah it is :D. I thought it was only possible in openapi v3.

model/application.go
eikendev: I renamed it. Can we close this conversation?

api/stream/client.go
eikendev: Can we close this conversation?

eikendev:
> Why overload the `/stream` API instead of encoding the version in the URL, akin to `/2/stream`?

Hm, I figure the current way of doing it is less of a change. @jmattheis, what do you say?

jmattheis:
@dcousens I give that question back to you, why should this be better? I'd say this is pure personal preference. The microsoft api guidelines also specify no preference https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#12-versioning

model/messagedeletions.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	return res
+}
+
+// MessageDeletionsExternal Model
+//
+// MessageDeletionsExternal holds information about deletions which will be sent to the clients.
+//
+// swagger:model MessageDeletions
+type MessageDeletionsExternal struct {
```
jmattheis: I've though a little about this. In case of a delete all we could send a "refresh state" event. Which just invalides all data on clientsides and forces a refetch. 

This is maybe easier to handle on clientside, as we don't have to check every message if it should be deleted.

What do you think?

model/messagedeletions.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	for key, value := range del.Messages {
+		res.IDs[key] = value.ID
+	}
+	return res
+}
+
+// MessageDeletionsExternal Model
+//
+// MessageDeletionsExternal holds information about deletions which will be sent to the clients.
```
jmattheis: I think we can omit will be "set to the clients". The description should only represent the event not who receives it.

model/messagedeletions.go
```diff
@@ -0,0 +1,31 @@
+package model
+
+// MessageDeletions holds information about deleted messages.
+type MessageDeletions struct {
+	Messages []*Message
+}
+
+// ToExternal converts the event into an external representation.
+func (del MessageDeletions) ToExternal() interface{} {
```
jmattheis: I wrote this after writing the text below. 

I think I'm now against supporting different versions of the websocket. I think it makes things much easier. The android app could easily support both version because the message event doesn't get breaking changes after we have a structure with discriminators https://github.com/gotify/server/pull/270/commits/59b3b38675e9b47e79ac091d4b54da628c7a31ea#r380298148

Changes to the api are pretty rare so maybe it is just overkill to support multiple versions, also there probably aren't many clients that use the stream api. 

>  Also, once a "version" is no longer needed (assuming clients will adopt the new version), we can simple remove one file and remove the mapping to it.

https://github.com/gotify/server/pull/270#discussion_r380301203
If the decoders will be deleted the major version of gotify/server has to be increased as it is a breaking change. So why not doing this now :D.

Also the change in the api is not severe, the easy fix for clients is to ignore all events which doesn't have the type message.

-----

Only for clarification. This is related to https://github.com/gotify/server/pull/270#discussion_r380292751

If we'd change the MessageDeletions in v3 to return whole messages. ToExternal isn't that nice anymore because it either has to be changed or another method like ToExternalV3 has to be added. Or the api version could be added as param.

Here some example code:
```go
func (del MessageDeletions) ToExternal(apiVersion uint) (interface{}, error) {
	switch apiVersion {
	case 1:
		return nil, ErrHiddenEvent
	case 2:
		res := &MessageDeletionsExternal{
			make([]uint, len(del.Messages)),
		}
		for key, value := range del.Messages {
			res.IDs[key] = value.ID
		}
		return &EventMessageV2{
			Type: "...",
			Content: res
		}
	case 3:
		res := &MessageDeletionsExternalV3{
			Messages: ConvertMessages(del.Messages),
		}
		return &EventMessageV3{
			Type: "...",
			Content: res
		}
	default:
		return nil, ErrUnsupportedApiVersion
	}
}
```
In the stream writer we can then check if the event != and then just send it without going through the decoder.

model/messagedeletions.go
eikendev: I'd argue that we would have a three-way message exchange then. The client needs to request all unread messages, and the server has to respond again. Doesn't that make things more power-consuming? Maybe just slightly...

What do you mean by checking every message on the client side? I don't think I understand what should be the challenge here.

model/messagedeletions.go
eikendev: Does the same apply to the documentation for `MessageExternal` then? I was trying to emphasize the difference that the `External` suffix makes.

model/messagedeletions.go
eikendev: As for the suggestion in the latter part of you comment I think you're right. I only put the `ToExternal` methods outside of the handler as I didn't see a good reason for moving it just now. You're right that new versions of the handler would have to write their own conversion function, so we'd have to move the ToExternal into the handler (as the handler should take care of distinguishing client versions).

Now, do I get it right that you want me to "remove" the v1 handler, so we only keep the v2 one (using discriminators)? I think this would make things much easier to read, and we get rid of the additional layer of handling client versions.

eikendev:
> Sorry that I'm not really fast in reviewing this but this isn't the only thing that needs attention in the gotify project and I also want to have some leisure time between working and maintaining an open source project.

No worries, take your time. This doesn't hurry as long as the codebase doesn't change drastically. I understand that maintaining open source projects is very time consuming.

model/messagedeletions.go
jmattheis: Yes, only keep version 2 and remove the version handling altogether.

model/messagedeletions.go
jmattheis: If the client has currently 100k messages in memory and all messages were deleted. Then the client would receive a message_deleted event with 100k messages. The client now has to remove every message with an id that is included inside the websocket event.

Anyway, I think the current solution is okay. The client can still decide what todo with the delete event. maybe just refreshes all its data when the websocket event contains more than X messages.

model/messagedeletions.go
jmattheis: I'd say yes, I prefer that the documentation contains information about what the struct represents and not how its used. But this is only a nice to have :D.

jmattheis:
Because of inactivity.

eikendev:
@jmattheis I was waiting on you finishing the review, so I can rework this. In your last review you said you were still not done with it.

jmattheis:
Uff, I though, as the change would require a somewhat bigger rewrite, that I'll wait till you fix the already existing comments. Sooorrrryy .____. my bad, I should've said that.

eikendev:
Do you still think replacing the old WebSocket endpoint is fine? This feature is still on my wishlist. Having multiple devices set up with Gotify is so confusing for me sometimes.

jmattheis:
Yeah, I think it is still fine. But I've a new suggestion. How about we add a `subtype` query parameter to the websocket endpoint. Default is `subtype=message`. With this flag it should be possible to subscribe to specific types f.ex `stream?subtype=message&subtype=deleteMessage`. We still change the content of the json object that it includes a type property, because this isn't breaking and the api stays compatible.

PR #279, opened 2020-03-11
Skip /etc/gotify/config.yml in tests

jmattheis:
Our config test ensures that the correct values will be extracted from
the config file and environment variables.
A globally defined config may change settings which are expected to have
default values.

See
https://github.com/gotify/server-aur-git/pull/2#issuecomment-597598574

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/gotify/configor"
+	"github.com/gotify/server/mode"
 )
 
 // Configuration is stuff that can be configured externally per env variables or config file (config.yml).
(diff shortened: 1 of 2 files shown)


jmattheis:
What do you mean with build directives? Yeah, the /etc config will be ignored in `go run` but I don't think that this is a problem.

eternal-flame-AD:
Something like this:
```go
// +build gotify-use-local-config

var configFiles []string = []string{"config.yml"}
```

```go
// +build !gotify-use-local-config

var configFiles []string = []string{"config.yml", "/etc/config.config.yml"}
```
And we use this tag when running tests.

jmattheis:
Does this give us any benefits to the current solution? I mean, now we have to specify a tag when running tests but we always want to exclude the /etc config in tests.

eternal-flame-AD:
I just don't think it's good to change the behavior of a simple `go build` just for a test. The commands to run tests are in Makefile so it does not really hurt to add a tag to that

jmattheis:
Okay, fair point. I'll use a different mode that the behavior of go run and go build doesn't change.

`go test ./...` should work without any extra tags. Like go build should :D.

PR #301, opened 2020-04-19
Allow multiple CORS origins 

### LINKED ISSUE #299: Allow multiple CORS origins
First off, thanks for this project. I make use of it literally every day.

I realize this is a fairly niche request.

I would be happy to implement it myself, and do all necessary documentation/testing should this be approved.

I am currently actively developing a [browser extension](https://github.com/StewartThomson/Gotify-Ext) for gotify. One issue I've run into is the setting of the CORS `Access-Control-Allow-Origin` header. Currently, it can only be set to one origin at a time. Which means that I, or any user, can't use both the Chrome extension and the Firefox extension (or any other browser client they desire).

I think it would make the most sense to make this similar to the 'stream: allowedorigins' config already present, as that permits multiple origins. It could be 'cors: allowedorigins'. The code would just need to loop through that list and see if the request matches any of the specified origins, in order to set the response header correctly.

Currently, the only way to achieve this functionality is to set 'Access-Control-Allow-Origin: "*"' which would certainly work but is not ideal for security purposes

Let me know! Thanks

StewartThomson:
- Added Cors configuration section
- added official gin cors middleware to handle this
- CORS origins can now be defined with multiple origins and with regex, just like with the streams origins
- still backwards compatible
- explicitly defined headers defined in "ResponseHeaders" will take precedence over those defined in the new Cors section

Fixes #299 

Should deploy https://github.com/gotify/website/pull/32 along with this

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -24,7 +24,7 @@ server:
 #    Access-Control-Allow-Origin: "*"
 #    Access-Control-Allow-Methods: "GET,POST"
 
-  cors: # A better way to define CORS headers. NOTE: setting the CORS headers above will override the CORS config defined here
+  cors: # Sets cors headers only when needed and provides support for multiple allowed origins. Overrides Access-Control-* Headers in response headers.
     alloworigins:
 #      - ".+.example.com"
 #      - "otherdomain.com"
--- a/config.example.yml
+++ b/config.example.yml
@@ -24,6 +24,16 @@ server:
 #    Access-Control-Allow-Origin: "*"
 #    Access-Control-Allow-Methods: "GET,POST"
 
+  cors: # A better way to define CORS headers. NOTE: setting the CORS headers above will override the CORS config defined here
+    alloworigins:
+#      - ".+.example.com"
+#      - "otherdomain.com"
+    allowmethods:
+#      - "GET"
+#      - "POST"
+    allowheaders:
+#      - "Authorization"
+#      - "content-type"
   stream:
     allowedorigins: # allowed origins for websocket connections (same origin is always allowed)
 #      - ".+.example.com"
--- a/config.example.yml
+++ b/config.example.yml
@@ -21,8 +21,7 @@ server:
 #      - myotherdomain.tld
 
   responseheaders: # response headers are added to every response (default: none)
-#    Access-Control-Allow-Origin: "*"
-#    Access-Control-Allow-Methods: "GET,POST"
+#    X-Custom-Header: "custom value"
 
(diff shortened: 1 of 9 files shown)


jmattheis:
Hey, Thanks for your contribution, I'll have a look at this PR tomorrow in the afternoon.

router/router_test.go
```diff
@@ -108,6 +110,105 @@ func TestHeadersFromConfiguration(t *testing.T) {
...
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Origin", "http://test123.com")
+	assert.Nil(t, err)
+
+	res, err := client.Do(req)
+	assert.Nil(t, err)
+	assert.Equal(t, "http://test123.com", res.Header.Get("Access-Control-Allow-Origin"))
+}
+
+// We want headers in responseheaders to override the cors config, as it is more specific
```
jmattheis, line 171: It is the opposite actually, if you set a valid origin 
```
config.Server.Cors.AllowOrigins = []string{"http://test123.com"}
```
the headers actually will be overriden by the cors config. I'm okay with this, but the comment must be fixed.
(only Access-Control-Allow-Origin will be overridden, the other will be overriden of http method options is used)

router/router_test.go
```diff
@@ -108,6 +110,105 @@ func TestHeadersFromConfiguration(t *testing.T) {
 	assert.Equal(t, "Nice", res.Header.Get("New-Cool-Header"))
 }
 
+func TestHeadersFromCORSConfig(t *testing.T) {
```
jmattheis, line 113: Could you add a test with an invalid origin? It should return statuscode 403 I guess.

config.example.yml
```diff
@@ -24,6 +24,16 @@ server:
 #    Access-Control-Allow-Origin: "*"
 #    Access-Control-Allow-Methods: "GET,POST"
 
+  cors: # A better way to define CORS headers. NOTE: setting the CORS headers above will override the CORS config defined here
```
jmattheis, line 27:
```suggestion
  cors: # Sets cors headers only when needed and provides support for multiple allowed origins. Overrides Access-Control-* Headers in response headers.
```

config.example.yml
```diff
@@ -24,6 +24,16 @@ server:
 #    Access-Control-Allow-Origin: "*"
```
jmattheis, line 24: I think we need different header examples here, otherwise It looks like useless settings :D.
```
responseheaders:
  X-Custom-Header: "custom value"
```

StewartThomson:
Hey @jmattheis thanks for the review

I've made the changes as requested

Issue #186, opened 2019-05-03
Allow the apps to send messages to specific clients


Mike325:
Right now apps broadcast messages to all subscribed clients, it would be nice if the apps could send  messages just some of the clients, something like this

    /message?token=<app_token>&client=<client_id>

Broadcast messages would still be available when the `client_id` is missing

jmattheis:
Hey, thanks for your issue!

Do you have a specific use case for this?

We have a similar issue with a different solution https://github.com/gotify/android/issues/61 in there the client chooses to what applications it subscribes.

Mike325:
I thought this could be useful in a couple of use case like  

* There are multiple clients that react to the notifications, something like tasker may be listening and trigger some actions with specific messages
* Some messages may not be relevant to all the clients, if I tablet is running out of battery it may send a message to the phone informing about it, right now it would broadcast the message and the tablet itself would receive the notification.

I also think gotify/android#61 can work together with this in the way you may want to silence the notifications even if they are target directly to your device

kallisti5:
This was one of the first requested features when I brought this up to my team. It would be nice to at minimum set target clients for notifications, or put clients into groups.

As a use case example, we get alerts for things when they break.  If we could define "groups of clients interested in windows-related notifications, and groups of clients interested in linux-related notifications" it would go a long way.

van-beek:
Yes this would be a nice feature. especialy for the stream websocket. subscribe to one application or send message to a client. Then one server can be used for mulitple apps.

happymario:
When are you gonna add this feature?

jmattheis:
I currently have other projects that I work on. Still, someone from the community can implement this feature if it is really needed :D.

Before someone takes this issue, it would be helpful to discuss the changes needed for this to work.. F.ex. if a client can see messages that were sent with another client id and only won't receive a push notification for it.

artfulrobot:
I came to this issue thinking "Why can't I configure which apps get to send to which users?"

It seems the model at the mo is 

`many apps >--< send many messages >--< many users`

I'm just using gotify personally, so that's fine, but I'd like to start using it for other purposes. Like I'd like to be able to add clients who will only get certain categories of message (and must not see others), which seems to be what this issue is about, although there are quite different ideas.

1. [this comment](https://github.com/gotify/server/issues/186#issuecomment-489196218) suggested that the *client chooses* which apps, from *all* of the apps available.

[...]

PR #305, opened 2020-04-24
add api to send notification by client token

happymario:
I added 'message_client' api to send message by client token.
It is assigned to issuse '#186.
Please check it.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -25,6 +25,7 @@ type MessageDatabase interface {
 	DeleteMessagesByApplication(applicationID uint) error
 	CreateMessage(message *model.Message) error
 	GetApplicationByToken(token string) (*model.Application, error)
+	GetClientByToken(token string) (*model.Client, error)
 }
 
 var timeNow = time.Now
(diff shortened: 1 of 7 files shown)


api/message.go
```diff
@@ -414,3 +417,58 @@ func toExternalMessages(msg []*model.Message) []*model.MessageExternal {
...
+func (a *MessageAPI) CreateMessageForClient(ctx *gin.Context) {
+	message := model.MessageExternal{}
+	if err := ctx.Bind(&message); err == nil {
+		token := auth.GetTokenID(ctx)
+		client, err := a.DB.GetClientByToken(token)
+		if success := successOrAbort(ctx, 500, err); !success {
+			return
+		}
+
+		message.ApplicationID = 0
```
jmattheis, line 464: Hey, Thanks for your contribution.

Adding a send message api with client authentication is not possible with the current design. Each message must have an application assigned otherwise the ui don't know which icon to show for the message.

Like written in https://github.com/gotify/server/issues/186 the /message api should be extended to accept a client_id query parameter to only send the message to a specific client. The authentication should still be done with an application token.

How the client_id parameter should behave isn't clearly defined in https://github.com/gotify/server/issues/186 this still needs some discussion before implementing this feature. See my last comment.

PR #321, opened 2020-05-28
Support custom ports in Docker healthcheck

eikendev:
Currently, the Docker container running with a custom port will always be shown unhealthy. This fix introduces a default port, which is overwritten when a user specifies a custom environment variable via Docker.

Please verify again that this works. I had to test a downstripped version, because my Podman didn't like it as is.


jmattheis:
Build currently fails because (I guess) formatting errors inside generated files, I'll fix this tomorrow/on the weekend.

Issue #257, opened 2020-01-15
Multi-Arch Docker Image


mattheys:
**Is your feature request related to a problem? Please describe.**
I want to create and maintain a platform independent docker-compose file to build my home automation system on any hardware, x86 VPS or Arm SBC.

**Describe the solution you'd like**
It would be useful therefore if Gotify were to use the Docker Multi-Arch build process so that it only be referenced by gotify/server regardless of which platform it is running on. (https://www.docker.com/blog/multi-arch-images/)

**Describe alternatives you've considered**
I could attempt to do the build process myself but it would then mean it would slowly drift out of sync with any updates and Watchtower would not work unless I went in and rebuilt the image myself when I knew there was an update.

**Additional context**
The Multi-Arch build process also builds more Arm versions than just v7 for example it also builds v6 and arm64 which would enable wider support of devices (older Pi's) or performance gains on more modern arm64 devices.

mattheys:
Here is a link to someone who has done this without Docker Desktop using GitLab CI. https://medium.com/@bamnet/building-multiarch-docker-images-8a70002b3476

jmattheis:
I've already created a multi-arch build for another project of mine -> https://github.com/traggo/server/blob/a46aba9b928a7d3cd59146b3b2750d820687b636/Makefile#L139

this can probably be reused.

KROSF:
[buildx](https://github.com/docker/buildx) is the best option to this.
We can also set up actions and use the [crazy-max/ghaction-docker-buildx](https://github.com/crazy-max/ghaction-docker-buildx).

@jmattheis if you think it is fine. I will try this weekend.

jmattheis:
I haven't worked with buildx yet, so I can't give an answer to this now. This project currently uses travis-ci for the build pipeline, so using a github action isn't an option.

ngosang:
Another example with GitHub actions. ARM32, ARM64 & AMD64
https://github.com/NoahCardoza/CloudProxy/pull/32/files

ionutz89:
Any update about multi arch?

moinologics:
any update here?? it is hard make it work if image arch is separated by tags.

eternal-flame-AD:
Looks doable and make people want this , I ill work on it this eee5! @moinologics .

Reference: https://docs.docker.com/build/building/multi-platform/

PR #374, opened 2020-12-19
Add arm64 docker build

random-dwi:
this PR adds building an arm64 docker image to the build.

its related to https://github.com/gotify/server/issues/257 and solves the issue at least for arm64.


jmattheis (approved):
Thanks!

Issue #276, opened 2020-02-26
Send from Web UI?


STaRDoGG:
So, I'm in the Gotify WebUI, and I want to send a message to one of the Clients in my list of Clients (or perhaps even "broadcast" the same message to all of them), how do I do this?

One simple example is I have a password that I created on my desktop, and want to send it over to my phone via a gotify msg.

jmattheis:
Hey @STaRDoGG. 

Clients aren't supposed to send messages only applications are (See https://gotify.net/docs/). To send a message you've to create an application and then push a message. See https://gotify.net/docs/pushmsg

Still, it may be useful to send messages via the ui, so I'm changing this issue to a feature request.

4oo4:
I would like to see this too, but understand why it's currently not there. I know when I switched to this from Pushbullet, I had to get used to the difference between an application and a client, whereas with Pushbullet a device kinda acted like both. So not the most important thing to have but might help people switch from Pushbullet.

STaRDoGG:
Hmmmm... I promise I'm not trying to be a snot, lol, but I can't:

Send any notifs/messages from the Web UI to 1 or more apps (or devices)
Can't send any notifs/messages to other apps/clients/devices from the official Android app
Can't view notifs/msgs as theyc ome in from any windows client other than loading a browser

I was hoping to 100% replace Pushbullet, as it's very unreliable, and the devs are out to lunch, plus I love the idea of a self hosted notif server, but it doesn't really seem that gotify is meant as a replacement, so to speak? Some very basic (imho) things aren't there like just passing notifications around to all of your clients in any direction at will (i.e. my Android phone, tablet, windows machine, pi, etc.)

Is any of this planned (other than the web ui feature request you added)?

STaRDoGG:
> I know when I switched to this from Pushbullet, I had to get used to the difference between an application and a client, whereas with Pushbullet a device kinda acted like both.

I agree; it is a lil 'weird' to get used to I spose, lol. But what also adds to the confusion is even in the literal Android 'app' you can't create and send from the app to anything. It's kinda just 'dead' unless it receives something from elsewhere.

pm-pm:
Why would my mother want to send something back when she gets the info that a package for her arrived? When i want to chat with her, i use a chat program.

pm-pm:
@STaRDoGG
Stop telling what you can't do. You can't even make coffee with the app!

With the gotify server and the gotify app you can be very well informed automatically!

jmattheis:
[...]

PR #384, opened 2021-01-22
Fix sidebar height & hide ScrollUpButton until scrolled (fade in)

### LINKED ISSUE #383: Web ui: make sidebar full height
PR incoming:

UI looks like this when scrolling:

image

Fix it like so:

image

StewartThomson:
Fixes #383 and issue found within #276

### DIFF
--- a/ui/src/common/ScrollUpButton.tsx
+++ b/ui/src/common/ScrollUpButton.tsx
@@ -8,25 +8,21 @@ class ScrollUpButton extends Component {
         opacity: 0,
     };
     componentDidMount() {
-        if (typeof window !== 'undefined') {
-            window.addEventListener('scroll', this.scrollHandler.bind(this));
-        }
+        window.addEventListener('scroll', this.scrollHandler);
     }
 
     componentWillUnmount() {
-        if (typeof window !== 'undefined') {
-            window.removeEventListener('scroll', this.scrollHandler.bind(this));
-        }
+        window.removeEventListener('scroll', this.scrollHandler);
     }
 
-    scrollHandler() {
-        let currentScrollPos = window.pageYOffset;
-        if (currentScrollPos > 0) {
-            this.setState({display: 'inherit', opacity: Math.min(currentScrollPos / 500, 1)});
-        } else {
-            this.setState({display: 'none', opacity: 0});
+    scrollHandler = () => {
+        const currentScrollPos = window.pageYOffset;
+        const opacity = Math.min(currentScrollPos / 500, 1);
+        const nextState = {display: currentScrollPos > 0 ? 'inherit' : 'none', opacity};
+        if (this.state.display !== nextState.display || this.state.opacity !== nextState.opacity) {
+            this.setState(nextState);
         }
-    }
+    };
 
     public render() {
         return (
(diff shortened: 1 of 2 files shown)


ui/src/common/ScrollUpButton.tsx
```diff
@@ -3,11 +3,36 @@ import KeyboardArrowUp from '@material-ui/icons/KeyboardArrowUp';
 import React, {Component} from 'react';
 
 class ScrollUpButton extends Component {
+    state = {
+        display: 'none',
+        opacity: 0,
+    };
+    componentDidMount() {
+        if (typeof window !== 'undefined') {
+            window.addEventListener('scroll', () => {
```
jmattheis, line 12: The handler probably should be removed on https://reactjs.org/docs/react-component.html#componentwillunmount

ui/src/common/ScrollUpButton.tsx
```diff
@@ -3,11 +3,36 @@ import KeyboardArrowUp from '@material-ui/icons/KeyboardArrowUp';
...
         return (
             <Fab
                 color="primary"
-                style={{position: 'fixed', bottom: '30px', right: '30px', zIndex: 100000}}
+                style={{
+                    position: 'fixed',
+                    bottom: '30px',
+                    right: '30px',
+                    zIndex: 100000,
+                    display: `${this.state.display}`,
```
jmattheis, line 33:
```suggestion
                    display: this.state.display,
```

ui/src/common/ScrollUpButton.tsx
```diff
@@ -3,11 +3,36 @@ import KeyboardArrowUp from '@material-ui/icons/KeyboardArrowUp';
...
+        display: 'none',
+        opacity: 0,
+    };
+    componentDidMount() {
+        if (typeof window !== 'undefined') {
+            window.addEventListener('scroll', () => {
+                let currentScrollPos = window.pageYOffset;
+                if (currentScrollPos > 0) {
+                    this.setState({display: 'inherit'});
+                    this.setState({opacity: currentScrollPos / 500});
```
jmattheis, line 16: Could you merge the 2 setStates into one? Same for the one below
```suggestion
                    this.setState({opacity: currentScrollPos / 500, opacity: Math.min(currentScrollPos / 500, 1)});
```
Also added Math.min that opacity doesn't exceed 1.

jmattheis (changes requested):
Thanks for your contribution. This actually has been bugging me too.

I've added remarks in subcomments.

ui/src/common/ScrollUpButton.tsx
jmattheis, line 16: Could you merge the 2 setStates into one? Same for the one below
```suggestion
                    this.setState({opacity: currentScrollPos / 500, opacity: Math.min(currentScrollPos / 500, 1)});
```
Also added Math.min that opacity doesn't exceed 1.

StewartThomson:
Feedback implemented @jmattheis

ui/src/common/ScrollUpButton.tsx
```diff
@@ -3,11 +3,43 @@ import KeyboardArrowUp from '@material-ui/icons/KeyboardArrowUp';
...
+    };
+    componentDidMount() {
+        if (typeof window !== 'undefined') {
+            window.addEventListener('scroll', this.scrollHandler.bind(this));
+        }
+    }
+
+    componentWillUnmount() {
+        if (typeof window !== 'undefined') {
+            window.removeEventListener('scroll', this.scrollHandler.bind(this));
```
jmattheis, line 18: I think this doesn't work as .bind will give a new reference and thus, doesn't unregister the old handler. I'll fix this. Also, I'll prevent some rerenders, by checking if the state will be changed.

jmattheis (approved):

PR #394, opened 2021-02-21
Allow registrations

p1gp1g:
This commit allow registrations for users. It might be useful to add a config parameter to enable or disable it.

### DIFF
--- a/api/user.go
+++ b/api/user.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"errors"
+	"fmt"
+	"net/http"
 
 	"github.com/gin-gonic/gin"
 	"github.com/gotify/server/v2/auth"
@@ -59,6 +61,7 @@ type UserAPI struct {
 	DB                 UserDatabase
 	PasswordStrength   int
 	UserChangeNotifier *UserChangeNotifier
+	Registration       bool
 }
 
 // GetUsers returns all the users
@@ -126,10 +129,13 @@ func (a *UserAPI) GetCurrentUser(ctx *gin.Context) {
 	ctx.JSON(200, toExternalUser(user))
 }
 
-// CreateUser creates a user
+// CreateUser create a user.
 // swagger:operation POST /user user createUser
 //
 // Create a user.
+//
+// With enabled registration: non admin users can be created without authentication.
+// With disabled registrations: users can only be created by admin users.
 //
 // ---
 // consumes: [application/json]
@@ -167,6 +173,32 @@ func (a *UserAPI) CreateUser(ctx *gin.Context) {
 		if success := successOrAbort(ctx, 500, err); !success {
 			return
 		}
+
+		var requestedBy *model.User
+		uid := auth.TryGetUserID(ctx)
+		if uid != nil {
+			requestedBy, err = a.DB.GetUserByID(*uid)
+			if err != nil {
+				ctx.AbortWithError(http.StatusInternalServerError, fmt.Errorf("could not get user: %s", err))
+				return
+			}
+		}
+
+		if requestedBy == nil || !requestedBy.Admin {
+			status := http.StatusUnauthorized
+			if requestedBy != nil {
+				status = http.StatusForbidden
+			}
+			if !a.Registration {
+				ctx.AbortWithError(status, errors.New("you are not allowed to access this api"))
+				return
+			}
+			if internal.Admin {
+				ctx.AbortWithError(status, errors.New("you are not allowed to create an admin user"))
+				return
+			}
+		}
+
 		if existingUser == nil {
 			if success := successOrAbort(ctx, 500, a.DB.CreateUser(internal)); !success {
 				return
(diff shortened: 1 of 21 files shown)


jmattheis (commented):
Hey, thanks for your contribution, could you first open an issue, to get some feedback from other users, if that is a needed feature? I rather not add stuff that is not really used even if it is behind a feature flag.

config.example.yml
```diff
@@ -50,3 +50,4 @@ defaultuser: # on database creation, gotify creates an admin user
 passstrength: 10 # the bcrypt password strength (higher = better but also slower)
 uploadedimagesdir: data/images # the directory for storing uploaded images
 pluginsdir: data/plugins # the directory where plugin resides
+registration: true # enable registrations
```
jmattheis, line 53:
```suggestion
registration: false # enable registrations
```
This should reflect the default values

api/user.go
```diff
@@ -182,6 +182,62 @@ func (a *UserAPI) CreateUser(ctx *gin.Context) {
...
+//         $ref: "#/definitions/Error"
+//   401:
+//     description: Unauthorized
+//     schema:
+//         $ref: "#/definitions/Error"
+//   403:
+//     description: Forbidden
+//     schema:
+//         $ref: "#/definitions/Error"
+func (a *UserAPI) Register(ctx *gin.Context) {
```
jmattheis, line 218: Please combine this method with the existing create user api, as the signature is exactly the same. 

The admin property on the user can then only be set if the current user is an administrator.

Please also add tests for the newly added functionality.

model/version.go
```diff
@@ -19,4 +19,10 @@ type VersionInfo struct {
 	// required: true
 	// example: 2018-02-27T19:36:10.5045044+01:00
 	BuildDate string `json:"buildDate"`
+
+	// Registration is allowed
+	//
+	// required: false
+	// example: true
+	Registration bool `json:"registration"`
```
jmattheis, line 27: I dislike this, as registration has really nothing to do with the version endpoint. There is probably a better solution by injecting the into the HTML at runtime, I'll do this after the PR.

ui/src/user/Login.tsx
```diff
@@ -61,6 +77,24 @@ class Login extends Component<Stores<'currentUser'>> {
         this.props.currentUser.login(this.username, this.password);
     };
 
+    private register = async (name: string, pass: string) => {
+        await axios.post('/registration', {name, pass});
```
jmattheis, line 81:
```suggestion
        await axios.post(config.get('url') + 'registration', {name, pass});
```
With just / requests will fail / be missrouted if gotify is [proxied to a subpath](https://gotify.net/docs/nginx#at-a-subpath).

ui/src/user/Login.tsx
```diff
@@ -4,21 +4,31 @@ import TextField from '@material-ui/core/TextField';
...
 import {observer} from 'mobx-react';
-import {inject, Stores} from '../inject';
+import {inject} from '../inject';
+import RegistrationDialog from './Register';
+import axios from 'axios';
+
+type Props = {
+    registration: boolean;
+    currentUser: CurrentUser;
+};
```
jmattheis, line 17: Use Stores in here and rename registration to showRegister.
```suggestion
type Props = Stores<'currentUser'> & {
    showRegister: boolean;
};
```

jmattheis (changes requested):
Hey, thanks for your contribution, remarks in subcomments.

api/user.go
karmanyaahm: I've been helping @p1gp1g on this PR, and I'm wondering, is this approach kind of what you mean here? [1]
The code isn't fully working, but I wanted to check in with you before testing/cleaning/debugging etc.
[1] https://github.com/karmanyaahm/server/commit/c0b4a897723114d2a4a12d84000d2138d161ceef

jmattheis: Yes, this goes in the right direction.

JuniorJPDJ:
bump?

karmanyaahm:
> bump?

I *think* the commits from the sub PRs 2 and 3 fix all issues from the review? I should've explicitly mentioned that here after both sub-PRs were merged, sorry.

jmattheis:
@karmanyaahm I've rebased the branch (you may have to update your ui dependencies with `yarn start`) and added some fixes to this PR, could you review them and test that it works well? And can I squash our commits into one, this way @p1gp1g will be the author of all changes related to registration? This is for a clean git history as in this PR the same files got changed multiple times.

karmanyaahm:
> I've rebased the branch (you may have to update your ui dependencies with `yarn install`) and added some fixes to this PR, could you review them and test that it works well? 

I read through all the commits and tested Register/CreateUser related things and everything seems to work very nicely. (some of the finishing touches you put in really help with user experience) 

> And can I squash our commits into one

:+1:

jmattheis (approved):

JuniorJPDJ:
is it possible to enable this feature through env var?

jmattheis:
@JuniorJPDJ Yes, it will be (it is not released yet).

PR #396, opened 2021-02-23
fix: download-tools

p1gp1g:
The download-tools command on Makefile fails with:

```
make download-tools
GO111MODULE=off go get -u github.com/go-swagger/go-swagger/cmd/swagger
cannot find package "github.com/hashicorp/hcl/hcl/printer" in any of:
	/opt/hostedtoolcache/go/1.15.8/x64/src/github.com/hashicorp/hcl/hcl/printer (from $GOROOT)
	/home/runner/go/src/github.com/hashicorp/hcl/hcl/printer (from $GOPATH)
Makefile:41: recipe for target 'download-tools' failed
make: *** [download-tools] Error 1
```

Removing the GO111MODULE=off fixes this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ check-js:
 	(cd ui && yarn testformat)
 
 download-tools:
-	GO111MODULE=off go get -u github.com/go-swagger/go-swagger/cmd/swagger
+	go get -u github.com/go-swagger/go-swagger/cmd/swagger
 
 embed-static:
 	go run hack/packr/packr.go


Makefile
```diff
@@ -38,7 +38,7 @@ check-js:
 	(cd ui && yarn testformat)
 
 download-tools:
-	GO111MODULE=off go get -u github.com/go-swagger/go-swagger/cmd/swagger
```
jmattheis, line 41: This changes versions of dependencies defined inside the go.mod, something like this shouldn't be done on the build.

p1gp1g: Ok, then maybe some dependencies have to be updated since it always fails, even for the CI worker.

Issue #450, opened 2021-11-16
Provide Docker tags for updates within major releases


kirbylink:
**Is your feature request related to a problem? Please describe.**
Major updates can break api calls but for security reasons, docker container should always be up to date. 
I am using watchtower for auto updates, to get the newest release for a specific tag. But with the "latest" tag it could happen, that a new major version is released and the application may not work anymore.

**Describe the solution you'd like**
To prevent updates that breaks the application, it would be nice to have images with only the major version as tag.
E. g. `gotify/server:2` so you get all the necessary bugfixes and minor updates but it will not break the api.

**Describe alternatives you've considered**
Using tags with minor number e. g. `gotify/server:2.1` is possible but you will miss newer updates if v2.2 is released.

jmattheis:
I'd accept a PR for this.

kirbylink:
Okay, i added a PR #452 (as draft version).
Since i can't test it locally, i need some advice, if the changes are correct and complete

PR #452, opened 2021-11-18
Adding docker builds for tagging builds with major version only

kirbylink:
See #450 
Additional docker build commands added to tag builds also with major version only


jmattheis (approved):
Thanks.

Issue #350, opened 2020-08-30
Add curl package on ARM Dockerfile


sadmin91:
On issue #308, curl package has been added on standard Dockerfile.
is it possible to add this package on ARM Dockerfile?

jmattheis:
That's not that easy because we need some kind of emulation on travis-ci to install packages while building the arm docker image.
I'll see what I can do.

minecraftchest1:
Github actions supports multiple operating systems on multiple architectures, including arm64.

Enissay:
Please update on this. It is needed to enable container healthcheck

PR #467, opened 2022-01-24
Update the ARM Dockerfiles to align with the default Dockerfile

dimo414:
I wanted to add a `HEALTHCHECK` to the Dockerfile and noticed there already is one for the default file but not for the ARM variants. I copied the directives in the default file over so they're all consistent.

### DIFF
--- a/docker/Dockerfile.arm64
+++ b/docker/Dockerfile.arm64
@@ -1,5 +1,12 @@
 FROM arm64v8/debian
+ENV GOTIFY_SERVER_PORT="80"
 WORKDIR /app
+RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install -yq \
+  tzdata \
+  curl \
+  ca-certificates \
+  && rm -rf /var/lib/apt/lists/*
 ADD gotify-app /app/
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.arm64
```diff
@@ -1,5 +1,12 @@
 FROM arm64v8/debian
```
jmattheis, line 1: This doesn't work this easily, because the build of the docker images is done on an amd64 system, thus, it cannot execute apt-get.

dimo414: Hmm ok, that seems like a misconfigured build system but I don't know enough Go to dig into it. Is there a bug tracking fixing this? It'd be great if the .arm* Dockerfiles (or some other documentation) had a comment noting that they are less functional than the default Dockerfile.

jmattheis: Not directly, but there is #350. This problem is more or less unrelated to Go, there should be a way for emulating arm on a amd64 system, but this needs to be integrated into the build pipeline.

Enissay: Any update please ?
not sure what is the difficulty, but if `wget` is easier, then that should work as well

PR #468, opened 2022-01-27
Add a pointer to #350 from the ARM Dockerfiles

dimo414:


docker/Dockerfile.arm64
```diff
@@ -1,5 +1,8 @@
+# Note: Due to https://github.com/gotify/server/issues/350 this Dockerfile
```
jmattheis, line 1: Well, it can be added, but then the ci build must be adjusted accordingly. I don't think such a comment is needed, because the issue already exists.

dimo414:
Well I, at least, was confused. Reading the source it seems like an oversight (the "main" Dockerfile is more feature-rich than the ARM variants, with no comment explaining the difference). Adding a pointer to the tracking issue would help readers understand a) the disparity is intentional and b) where to follow along for updates.

Issue #264, opened 2020-02-05
Use button for requesting notification permissions


jmattheis:
See https://blog.mozilla.org/futurereleases/2019/11/04/restricting-notification-permission-prompts-in-firefox/

PR #480, opened 2022-05-02
Implement service worker for offline notifications

### LINKED ISSUE #344: Webpush API for offline notifications?
Hi,
I'm searching for a (self hosted) webpush server, which can send offline notifications to browsers via existing vendor push server with a simple rest api.

Looks like gotify works only with websocket, but misses a offline notification to Chrome (for Android), Firefox, Edge.

There is a go lib maybe could Help to integrate webpush with service-worker client side?

MexHigh:
**This is a draft**

Fixes #344 

### How does it work
After logging in, a new service worker is created. The `gotify-login-key` gets passed to it via query parameter so that it can subscribe to the Websocket `/stream` endpoint. On incoming messages, it creates a notification. This way, no third party server must be used (e.g. a vendor server for the Webpush API).

The worker keeps running until the user logs out and the client is removed. It keeps running when the browser is closed.

It also seems like notifications work on android from within the browser now, as android only supports notifications via Service Worker ([source](https://web.dev/push-notifications/#are-service-workers-involved)).

(When testing, do not forget applying [this workaround](https://stackoverflow.com/a/54934302) when not testing over HTTPS or `localhost`.)

### Todos
- [x] Currently, when the UI is open, you get two notifications - one from the UI and one from the worker
- [x] Worker notification sound (?)
- [ ] Stability testing (does the worker really persist? How about browser or OS restarts?)

Please add your changes if necessary. I've kept the `console.log` statements for easier testing. They need to be removed before merging, of course.

### DIFF
--- a/ui/public/static/notification-worker.js
+++ b/ui/public/static/notification-worker.js
@@ -25,11 +25,37 @@ self.addEventListener("activate", () => {
     const ws = new WebSocket(`${wsProto}//${host}/stream?token=${gotifyKey}`)
     console.log("Notification worker connected to websocket, waiting for messages")
 
-    ws.onmessage = function(event) {
-        msgObj = JSON.parse(event.data)
-        self.registration.showNotification("WORKER: " + msgObj.title, {
-            body: msgObj.message
+    ws.onmessage = (event) => {
+
+        // check if any client is currently visible
+        // if so, skip sending notification from worker
+        self.clients.matchAll({
+            type: "window",
+            includeUncontrolled: true
         })
+            .then((windowClients) => {
+                var clientVisible = false
+                for (var i = 0; i < windowClients.length; i++) {
+                    const windowClient = windowClients[i]
+                    // check if a client is visible, then break
+                    if (windowClient.visibilityState === "visible") {
+                        clientVisible = true
+                        break
+                    }
+                }
+                return clientVisible
+            }) // use the bool to evaluate whether to send a notification 
+            .then((clientVisible) => {
+                if (!clientVisible) {
+                    var msgObj = JSON.parse(event.data) // parse event data, only if not visible
+                    self.registration.showNotification("WORKER: " + msgObj.title, {
+                        body: msgObj.message
+                    })
+                } else {
+                    console.log("not sending worker notification, as gotify window is visible")
+                }
+            })
+
     }
 
 })
--- a/ui/public/static/notification-worker.js
+++ b/ui/public/static/notification-worker.js
@@ -27,7 +27,9 @@ self.addEventListener("activate", () => {
 
     ws.onmessage = function(event) {
         msgObj = JSON.parse(event.data)
-        self.registration.showNotification("WORKER: " + msgObj.message)
+        self.registration.showNotification("WORKER: " + msgObj.title, {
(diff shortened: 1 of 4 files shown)


MexHigh:
Now, the service worker only displays a message, when no gotify window is open (implemented in 5aca059).

I wasn't able to find a way to play custom sounds with service worker notifications, sadly.

ui/src/index.tsx
```diff
@@ -61,7 +61,14 @@ const initStores = (): StoreMapping => {
...
 
-    stores.currentUser.tryAuthenticate().catch(() => {});
+    stores.currentUser.tryAuthenticate().then(() => {
+        // always request notification permission when logged in
+        Notification.requestPermission()
+            .then(perm => console.log("Notification permissions " + perm))
+            .catch(console.error)
+        // reregister worker
+        registerNotificationWorker(stores.currentUser.token());
+    }).catch(() => {});
```
jmattheis, line 71: Please add another button under the [enable notifications](https://github.com/gotify/server/blob/master/ui/src/layout/Navigation.tsx#L101) button, something like "enable background notifications". Should probably only visible if the notification permission was given. Gotify must not request this permission without user interaction, see https://github.com/gotify/server/issues/264

The button should be a toggle, enabling and disabling the notifications.

ui/public/static/notification-worker.js
```diff
@@ -0,0 +1,61 @@
...
+self.addEventListener("activate", () => {
+
+    console.log("Notification worker activated")
+
+    const host = location.host
+    const wsProto = location.protocol === "https:" ? "wss:" : "ws:"
+    const ws = new WebSocket(`${wsProto}//${host}/stream?token=${gotifyKey}`)
+    console.log("Notification worker connected to websocket, waiting for messages")
+
+    ws.onmessage = (event) => {
```
jmattheis, line 28: We probably need some reconnect logic here, otherwise the connection can break in the background and then the user doesn't get notifications anymore.

jmattheis (commented):
Hey @MexHigh, looks good so far, I've added some remarks in sub comments.

MexHigh:
I'm afraid I have some bad news. After testing this for a while now, I have come to the conclusion several times that the service worker is unfortunately not persistent. It stays online for a few minutes up to about half an hour after the last Gotify tab is closed.

I have read in some resources that you can keep Service worker "alive" by having it send itself regular messages via `postMessage`, however this is far from a stable solution.

The `onpush` event on the other hand seems to be able to "wake up" a service worker again, but so far this can only be used with the Push API. This in turn forces the use of vendor servers.

So I think this will not work here as intended. I'm closing the pull request now, but maybe it will be useful in the future if there is a way to actually use the Push API with a self-hosted server, or the "push" event can be triggered otherwise.

My sources are attached, in case someone wants to do some more research or to follow the whole thing.

* https://stackoverflow.com/a/29750936
* https://github.com/w3c/ServiceWorker/issues/980
* https://github.com/w3c/ServiceWorker/issues/980#issuecomment-782094985
* https://github.com/w3c/ServiceWorker/issues/1558

jmattheis:
@MexHigh thanks for your research!

PR #489, opened 2022-05-26
Fix login form password-manager support

rwese:
This only adds the `autoComplete` property to the Login TextField element.


jmattheis (approved):
Thanks.

Issue #466, opened 2022-01-13
Using large integers can lead to database corruption


d-k-bo:
**Can the issue be reproduced with the latest available release? (y/n)**

**Which one is the environment gotify server is running in?**
- [ ] Docker
- [X] Linux machine
- [ ] Windows machine
<details><summary>Docker startup command or config file here (please mask sensitive information)</summary><br>

```
export GOTIFY_SERVER_PORT=30080
./gotify-linux-amd64
```
</details>

**Do you have an reverse proxy installed in front of gotify server? (Please select None if the problem can be reproduced without the presense of a reverse proxy)**
- [X] None
- [ ] Nginx
- [ ] Apache
- [ ] Caddy
<details><summary>Reverse proxy configuration (please mask sensitive information)</summary><br>

```

```
</details>

**On which client do you experience problems? (Select as many as you can see)**
- [ ] WebUI
- [ ] gotify-cli
- [ ] Android Client 
- [X] 3rd-party API call (python w/ requests)

```python

import requests

r = requests.post(
    "http://localhost:30080/application?token=Cxxx",
    json={"name": "TestApplication", "id": 9223372036854775808},
)

print(r.json())

r = requests.get("http://localhost:30080/application?token=Cxxx")

print(r.json())

```

**What did you do?**

I created an application with id 9223372036854775808 and tried to get the list of all applications.

**What did you expect to see?**

An error while creating the application because the value is too large or an error because of specifying the 'id' field (which is read-only according to the [swagger configuration](https://raw.githubusercontent.com/gotify/server/v2.1.4/docs/spec.json)).

**What did you see instead? (Include screenshots, android logcat/request dumps if possible)**

The applications couldn't be listed anymore because the id overflowed and was stored as a negative value in the database.
```
{'id': 9223372036854775808, 'token': 'A_pITbCDbj.7S1Y', 'name': 'TestApplication', 'description': '', 'internal': False, 'image': 'static/defaultapp.png'}
{'error': 'Internal Server Error', 'errorCode': 500, 'errorDescription': 'sql: Scan error on column index 0, name "id": converting driver.Value type int64 ("-9223372036854775808") to a uint: invalid syntax'}
```
I think it should also be clear if fields like 'id', 'image' etc. can be used by the client or not. If they are supported, they should be specified in the API documentation, otherwise you should get a `400 Bad Request` response when using them.

jmattheis:
[...]

PR #498, opened 2022-07-20
Adds intermediate struct to handle application creation/update params

mateuscelio:
https://github.com/gotify/server/issues/466

An intermediate struct was added to handle params used in creation and update controllers preventing the edition of fields that are `readOnly` .

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -29,6 +29,23 @@ type ApplicationAPI struct {
 	ImageDir string
 }
 
+// Application Params Model
+//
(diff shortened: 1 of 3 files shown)


jmattheis (changes requested):
Looks good, but the build is currently failing.

mateuscelio:
> Looks good, but the build is currently failing.

Fixed!

jmattheis (approved):
thanks.

PR #516, opened 2022-10-10
Added Different notification color for priority

### LINKED ISSUE #269: Different notification color for priority
Hi

It would be nice to see the Notification in different colors in regards to the priority.
As priority information is stored in the db.

It would be nice to define this as variables as general in config file.
for example:
set_priority:
    0:  #FFFFFF
    1:  #000000

Thanks in advance

Tert0:
Closes #269 

Implemented different colors for message borders, which are based on the message priority.

image

**Note:**
I'll try to implement this to [gotify/android](https://github.com/gotify/android)

### DIFF
--- a/ui/src/message/Message.tsx
+++ b/ui/src/message/Message.tsx
@@ -73,10 +73,21 @@ interface IProps {
     image?: string;
     date: string;
     content: string;
+    priority: number;
     fDelete: VoidFunction;
     extras?: IMessageExtras;
     height: (height: number) => void;
 }
+
+const priorityColor = (priority: number) => {
+    if (priority >= 4 && priority <= 7) {
+        return 'rgba(230, 126, 34, 0.7)';
+    } else if (priority > 7) {
+        return '#e74c3c';
+    } else {
+        return 'transparent';
+    }
+};
 
 class Message extends React.PureComponent<IProps & WithStyles<typeof styles>> {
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -52,6 +52,10 @@ type Configuration struct {
 	UploadedImagesDir string `default:"data/images"`
 	PluginsDir        string `default:"data/plugins"`
 	Registration      bool   `default:"false"`
+	PriorityColors    []struct {
```
jmattheis, line 55: I don't think this needs to be configurable. We can hardcode the colors. i think something like this as border color should be enough:
```
rgba(243, 156, 18, Math.min(1, msg.priority/10))
```

api/application.go
```diff
@@ -31,7 +31,7 @@ type ApplicationAPI struct {
 
 // Application Params Model
 //
-// Params allowed to create or update Applications
+// # Params allowed to create or update Applications
```
jmattheis, line 34: Please revert all the backend changes related to golint, this should be done in a separate PR.

jmattheis (changes requested):
thanks for your contribution, I've added remarks in sub comments

jmattheis (approved):
Thanks, I've changed the colors a little bit, so that messages are colored orange for medium priority.

Tert0:
Could you please label this PR as `hacktoberfest-accepted`?

Issue #512, opened 2022-09-28
App Authentication.


zimbres:
This is a nice and useful tool, I see two opportunities of improvement.

Pass the token in the query string exposes it on the network, so its not recommended.

Using the X-Gotify-Key in the header avoid its exposure, but several tools, Grafana for example, has predefined way to authenticate and it is using "Authorization" header.

Thanks.

schoerg:
>Pass the token in the query string exposes it on the network, so its not recommended.

only if used unencrypted

zimbres:
No sure, for example, in a proxy server, its possible to see any url called, even if is https.

mateuscelio:
@jmattheis  I can implement this if it isn't in development yet. So the ideia is try to retrieve the token from  headers `Authorization` and `X-Gotify-Key` and from the query parameter as well, right?

jmattheis:
@mateuscelio yes

mateuscelio:
Ok, working on it

PR #527, opened 2022-12-01
Adds Authorization header auth method

mateuscelio:
It closes https://github.com/gotify/server/issues/512


jmattheis (approved):
Thanks, I've added a description to the authentication thingy in the docs and renamed a method.

PR #560, opened 2023-04-24
Adds ClientParams strcut to handle creation and update params

mateuscelio:
https://github.com/gotify/server/issues/466

An intermediate struct was added to handle params used in creation and update client controllers preventing the edition of fields that are readOnly.


jmattheis (dismissed):
Thanks!

Issue #447, opened 2021-11-01
Add copy app token button in web gui


PanderMusubi:
**Is your feature request related to a problem? Please describe.**
To prevent mistakes and offer more easy copying of app tokens, a button to copy a token could be added for each token.

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**
This is common functionality also found in i.e. Firefox.

**Additional context**
The button could be added directly to the right of where the token is shown.

tessus:
I think this issue can be closed, since it has been implemented.

PR #561, opened 2023-04-26
Add copy app token button in web gui

Zlendy:
This pull request implements #447.

### DIFF
--- a/ui/src/application/Applications.tsx
+++ b/ui/src/application/Applications.tsx
@@ -13,7 +13,7 @@ import React, {ChangeEvent, Component, SFC} from 'react';
 import ConfirmDialog from '../common/ConfirmDialog';
 import DefaultPage from '../common/DefaultPage';
 import Button from '@material-ui/core/Button';
-import ToggleVisibility from '../common/ToggleVisibility';
+import CopyableSecret from '../common/CopyableSecret';
 import AddApplicationDialog from './AddApplicationDialog';
 import {observer} from 'mobx-react';
 import {observable} from 'mobx';
@@ -166,7 +166,7 @@ const Row: SFC<IRowProps> = observer(
             </TableCell>
             <TableCell>{name}</TableCell>
             <TableCell>
-                <ToggleVisibility value={value} style={{display: 'flex', alignItems: 'center'}} />
+                <CopyableSecret value={value} style={{display: 'flex', alignItems: 'center'}} />
             </TableCell>
             <TableCell>{description}</TableCell>
             <TableCell align="right" padding="none">
--- a/ui/src/client/Clients.tsx
+++ b/ui/src/client/Clients.tsx
@@ -12,13 +12,13 @@ import React, {Component, SFC} from 'react';
 import ConfirmDialog from '../common/ConfirmDialog';
 import DefaultPage from '../common/DefaultPage';
 import Button from '@material-ui/core/Button';
-import ToggleVisibility from '../common/ToggleVisibility';
 import AddClientDialog from './AddClientDialog';
 import UpdateDialog from './UpdateClientDialog';
 import {observer} from 'mobx-react';
 import {observable} from 'mobx';
 import {inject, Stores} from '../inject';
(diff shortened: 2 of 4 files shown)


ui/src/client/Clients.tsx
```diff
@@ -110,6 +112,15 @@ interface IRowProps {
     fDelete: VoidFunction;
 }
 
+const copyToClipboard = async (text: string) => {
+    try {
+        await navigator.clipboard.writeText(text);
+        console.info("Copied to clipboard:", text);
```
mateuscelio, line 118: For the notifications, Instead of using `cosole.info` and `console.error` I suggest you to make use of `SnackReporter`  to make it look like the example below: 
image

mateuscelio (changes requested):

ui/src/client/Clients.tsx
```diff
@@ -119,6 +130,11 @@ const Row: SFC<IRowProps> = ({name, value, fEdit, fDelete}) => (
                 style={{display: 'flex', alignItems: 'center', width: 200}}
             />
         </TableCell>
+        <TableCell align="right" padding="none">
+            <IconButton onClick={() => {copyToClipboard(value)}} className="copy" title="Copy to clipboard">
```
jmattheis, line 134: The end-to-end test currently fail, probably because the column index is out of sync here
https://github.com/gotify/server/blob/9d4e37aa87e5aea5d04b9a6285ca807099c46fbd/ui/src/tests/client.test.ts#L17-L22

ui/src/client/Clients.tsx
Zlendy: Thanks for the tip! I'm trying to add an additional action to ClientStore to implement it!

jmattheis: You can directly inject the SnackManager into the Clients component by adding it to the inject function:
```
export default inject('clientStore', 'snackManager')(Clients);
```

Zlendy: Is it better to do it that way? I have already implemented it with a ClientStore.

Zlendy:
Okay. I added a Copy to Clipboard button to "Applications" and "Plugins" too.

image
image

I have personally tested the button and it works in "Clients" and "Applications". I tried to install a plugin to test it there too but unfortunately I was not able to do it.

This PR is still not ready to be merged due to e2e tests on "Applications" and "Plugins" failing.

Zlendy:
Hi. I finally realized the embarrassing thing that was making the test fail. All tests are passing now and the button is working as it should.

ui/src/application/Applications.tsx
```diff
@@ -168,6 +173,14 @@ const Row: SFC<IRowProps> = observer(
             <TableCell>
                 <ToggleVisibility value={value} style={{display: 'flex', alignItems: 'center'}} />
             </TableCell>
+            <TableCell>
+                <IconButton
```
mateuscelio, line 177: This code is pretty much the same on the other components, so would be great create a separate component(`CopyToClipboardButton`) for it inside `src/common` directory.

mateuscelio (changes requested):
@Zlendy I left some comments. To summarize, I think that a better implementation would be create a component for the copy button and inject in that class the snack manager, this way, it wouldn't be necessary to change SnackManager and component's stores. To help, I created a [PR into your fork](https://github.com/Zlendy/gotify-server/pull/1). Let me know If you have more questions.

ui/src/application/Applications.tsx
Zlendy: Thank you very much!
@jmattheis also suggested injecting snackManager into a component. But I didn't understand what they meant until I saw your example.
I'm going to apply the changes you suggested asap, cheers!

jmattheis (approved):
Thanks. I've moved the method into the ToggleVisibility component and renamed it to CopyableSecret. It fits there pretty well, because you normally want to copy all secrets, and it also requires fewer changes inside the codebase.

PR #566, opened 2023-05-23
log: filter requests to /health from localhost

### LINKED ISSUE #428: Remove "/health" from logging?
The log file I created with `entrypoint: /bin/bash -c '/app/gotify-app | tee /app/data/gotify.log'` in my docker-compose.yml is pretty quick filling up with entries like `[GIN] 2021/08/21 - 19:02:50 | 200 |     440.732µs |       127.0.0.1 | GET      "/health"`. Is there a way to prevent these kind of entries in the log file?

tessus:
fixes #428 

I have tested it and it works perfectly. Returning an empty string will not print an empty line, but doesn't output a line at all. I added the if block at the top so that it's an early exit without unnecessary processing if a match occurs.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -176,6 +176,10 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 var tokenRegexp = regexp.MustCompile("token=[^&]+")
 
 func logFormatter(param gin.LogFormatterParams) string {
+	if param.ClientIP == "127.0.0.1" && param.Path == "/health" {
+		return ""
+	}
+
 	var statusColor, methodColor, resetColor string
 	if param.IsOutputColor() {
 		statusColor = param.StatusCodeColor()


router/router.go
```diff
@@ -176,6 +176,10 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 var tokenRegexp = regexp.MustCompile("token=[^&]+")
 
 func logFormatter(param gin.LogFormatterParams) string {
+	if param.ClientIP == "127.0.0.1" && param.Path == "/health" {
```
jmattheis, line 179: Could you also check against ipv6 localhost `::1`. I'm offline for now, but will probably test this tomorrow.

tessus: will do

tessus: done

jmattheis (approved):
:+1:

PR #568, opened 2023-05-24
add API endpoint to remove an app's image

tessus:
I have a few comments:

1. ~~I don't know if the swagger comments are generated or if I have to manually add them before the function.~~
2. ~~I am also not sure whether an error message makes sense, in case there is no image. IMO it should be a warning or info message, but not sure how to generate such a type.~~
3. I am not a UI person. I have ideas, but how to realize them with HTML/CSS is another story. It would take me a few days to figure these things out. But I suggest that there is an overlay image (an `x`) in the bottom right corner of the icon when hovered over the icon in the frontend. Upon click, the API is called. (The onclick js I can most likely do, but no idea how to do the overlay image.)

Update: ~~I will also write the test cases as soon as the specs are finished for the API call.~~ Tests added.

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -409,11 +409,12 @@ func (a *ApplicationAPI) RemoveApplicationImage(ctx *gin.Context) {
 				return
 			}
 
+			image := app.Image
 			app.Image = ""
 			if success := successOrAbort(ctx, 500, a.DB.UpdateApplication(app)); !success {
 				return
 			}
-			os.Remove(a.ImageDir + app.Image)
+			os.Remove(a.ImageDir + image)
 			ctx.JSON(200, withResolvedImage(app))
 		} else {
 			ctx.AbortWithError(404, fmt.Errorf("app with id %d doesn't exists", id))
--- a/api/application.go
+++ b/api/application.go
@@ -358,6 +358,69 @@ func (a *ApplicationAPI) UploadApplicationImage(ctx *gin.Context) {
 	})
 }
 
+// RemoveApplicationImage deletes an image of an application.
+// swagger:operation DELETE /application/{id}/image application removeAppImage
+//
+// Deletes an image of an application.
+//
+// ---
+// consumes: [application/json]
+// produces: [application/json]
+// parameters:
+// - name: id
+//   in: path
+//   description: the application id
+//   required: true
+//   type: integer
+//   format: int64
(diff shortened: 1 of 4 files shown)


api/application.go
```diff
@@ -358,6 +358,30 @@ func (a *ApplicationAPI) UploadApplicationImage(ctx *gin.Context) {
 	})
 }
```
jmattheis, line 360: > I don't know if the swagger comments are generated or if I have to manually add them before the function.

You need to create the swagger comment manually above the method and then run `make download-tools update-swagger` afterward.

api/application.go
```diff
@@ -358,6 +358,30 @@ func (a *ApplicationAPI) UploadApplicationImage(ctx *gin.Context) {
...
 
+func (a *ApplicationAPI) RemoveApplicationImage(ctx *gin.Context) {
+	withID(ctx, "id", func(id uint) {
+		app, err := a.DB.GetApplicationByID(id)
+		if success := successOrAbort(ctx, 500, err); !success {
+			return
+		}
+		if app != nil && app.UserID == auth.GetUserID(ctx) {
+			if app.Image == "" {
+				ctx.AbortWithError(400, fmt.Errorf("app with id %d does not have a customized image", id))
```
jmattheis, line 369: > I am also not sure whether an error message makes sense, in case there is no image. IMO it should be a warning or info message, but not sure how to generate such a type.

I'm fine with 400, the UI shouldn't allow deleting an image when there isn't one.

api/application.go
tessus: Will do.

tessus:
Please let me know, if anything is still missing.

api/application.go
```diff
@@ -358,6 +358,69 @@ func (a *ApplicationAPI) UploadApplicationImage(ctx *gin.Context) {
...
+				return
+			}
+
+			app.Image = ""
+			if success := successOrAbort(ctx, 500, a.DB.UpdateApplication(app)); !success {
+				return
+			}
+			os.Remove(a.ImageDir + app.Image)
```
jmattheis, line 416: the app.Image was already set to an empty string here, so the old image file won't be removed correctly here. I'll fix that.

tessus: Argh. Yes. I moved it down, so that it won't be deleted should the database call fail for any reason. 

But you are correct of course. Stupid me.

PR #581, opened 2023-07-14
migrate from packr to go:embed

eternal-flame-AD:
Go [embed](https://pkg.go.dev/embed) has been out for a while now and packr [recommends](https://pkg.go.dev/github.com/gobuffalo/packr/v2#section-readme)  moving to //go:embed as well, this PR removes packr from the dependency list and uses go:embed to embed swagger and UI files.


ui/serve.go
```diff
@@ -31,19 +42,22 @@ func Register(r *gin.Engine, version model.VersionInfo, register bool) {
 	ui.GET("/index.html", serveFile("index.html", "text/html", noop))
 	ui.GET("/manifest.json", serveFile("manifest.json", "application/json", noop))
 	ui.GET("/asset-manifest.json", serveFile("asset-manifest.json", "application/json", noop))
-	ui.GET("/static/*any", gin.WrapH(http.FileServer(box)))
+	ui.GET("/static/*any", gin.WrapH(&AddPrefix{
+		Prefix:       "/build",
+		InnerHandler: http.FileServer(http.FS(box)),
+	}))
```
jmattheis, line 48: Please use [fs.Sub](https://pkg.go.dev/io/fs#Sub) instead of the custom AddPrefix
```
//go:embed build/*
var embedFS embed.FS
var buildUI, _ = fs.Sub(embedFS, "build")

func Register() {
	ui.GET("/static/*any", gin.WrapH(http.FileServer(http.FS(buildUI))))
}
```

docs/swagger.go
```diff
@@ -1,14 +1,15 @@
...
 	"strings"
 
 	"github.com/gin-gonic/gin"
-	"github.com/gobuffalo/packr/v2"
 	"github.com/gotify/location"
 )
 
-var box = packr.New("docs", "./")
+//go:embed spec.json
+var box embed.FS
```
jmattheis, line 12: It's possible to directly inject a single file into a string, then we don't have the overhead with the embed.FS, the import of embed is still needed.
```go
import (
	_ "embed"
	"strings"

	"github.com/gin-gonic/gin"
	"github.com/gotify/location"
)

//go:embed spec.json
var spec string
```

jmattheis (changes requested):
Looks good, only two remarks in subcomments.

jmattheis (approved):
:shipit:

PR #582, opened 2023-07-24
add last seen field to client (fixes #400)

### LINKED ISSUE #400: last message/last ping in /clients
If you use `gotify` to handle notifications it would be nice to know if clients are still reachable (or have the app installed, app killed). You can then fallback to another channel (email, sms ...) to not miss updates.

`/client` already returns the list of clients, add another field like `last_message` or `last_keepalive`/`lastping` to show when that client has last received a message or last responded to a ping.

eternal-flame-AD:
This PR fixes #400 by adding a last seen field to the client model. The last seen field is updated on connection & every pong response.

### DIFF
--- a/model/client.go
+++ b/model/client.go
@@ -1,5 +1,7 @@
 package model
 
+import "time"
+
 // Client Model
 //
 // The Client holds information about a device which can receive notifications (and other stuff).
@@ -24,4 +26,9 @@ type Client struct {
 	// required: true
 	// example: Android Phone
 	Name string `gorm:"type:text" form:"name" query:"name" json:"name" binding:"required"`
+	// The last time the client token was used.
+	//
+	// read only: true
+	// example: 2019-01-01T00:00:00Z
+	LastUsed *time.Time `json:"lastUsed"`
 }
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -38,6 +38,7 @@ func (s *IntegrationSuite) BeforeTest(string, string) {
 	var err error
 	s.db = testdb.NewDBWithDefaultUser(s.T())
 	assert.Nil(s.T(), err)
+
 	g, closable := Create(s.db.GormDatabase,
 		&model.VersionInfo{Version: "1.0.0", BuildDate: "2018-02-20-17:30:47", Commit: "asdasds"},
 		&config.Configuration{PassStrength: 5},
--- a/ui/src/application/Applications.tsx
+++ b/ui/src/application/Applications.tsx
@@ -21,6 +21,7 @@ import {inject, Stores} from '../inject';
 import * as config from '../config';
 import UpdateDialog from './UpdateApplicationDialog';
 import {IApplication} from '../types';
+import {LastUsedCell} from '../common/LastUsedCell';
 
 @observer
 class Applications extends Component<Stores<'appStore'>> {
@@ -67,6 +68,7 @@ class Applications extends Component<Stores<'appStore'>> {
                                     <TableCell>Token</TableCell>
                                     <TableCell>Description</TableCell>
                                     <TableCell>Priority</TableCell>
+                                    <TableCell>Last Used</TableCell>
                                     <TableCell />
                                     <TableCell />
                                 </TableRow>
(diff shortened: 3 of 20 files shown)


model/client.go
```diff
@@ -24,4 +26,9 @@ type Client struct {
 	// required: true
 	// example: Android Phone
 	Name string `gorm:"type:text" form:"name" query:"name" json:"name" binding:"required"`
+	// The last time the client was seen.
+	//
+	// read only: true
+	// example: 2019-01-01T00:00:00Z
+	LastSeen *time.Time `json:"lastSeen"`
```
jmattheis, line 33: I think I'd prefer the name LastUsed, then it's more related to the token.

model/client.go
```diff
@@ -24,4 +26,9 @@ type Client struct {
 	// required: true
 	// example: Android Phone
 	Name string `gorm:"type:text" form:"name" query:"name" json:"name" binding:"required"`
+	// The last time the client was seen.
```
jmattheis, line 29: Can we also update last seen when the token was used for api calls. I think the update could be done in auth/authentication.go in the RequireClient call. Similar to what I described in this comment: https://github.com/gotify/server/issues/400#issuecomment-1247082718

ui/src/client/Clients.tsx
```diff
@@ -119,6 +123,7 @@ const Row: SFC<IRowProps> = ({name, value, fEdit, fDelete}) => (
                 style={{display: 'flex', alignItems: 'center', width: 200}}
             />
         </TableCell>
+        <TableCell>{lastSeen ? <TimeAgo date={lastSeen} /> : 'Never'}</TableCell>
```
jmattheis, line 126: Maybe we can do something like if (now - lastSeen) < 5min then it only displays the text "recently".

router/router_test.go
```diff
@@ -332,12 +336,30 @@ func (s *IntegrationSuite) TestSendMessage() {
 	json.NewDecoder(res.Body).Decode(token)
 	assert.Equal(s.T(), "backup-server", token.Name)
 
+	req = s.newRequest("POST", "client", `{"name": "mock-client"}`)
```
jmattheis, line 339: This feels a little out of place here, the test only tests the message api and now it also checks clients. I think it'd be enough to check if last seen was updated in the authentication package, and that the proposed GetActiveClientTokens method works correctly one of the stream tests.

eternal-flame-AD:
Thanks for the review, its been some time since I was active here. Yeah I should have read #400 more carefully you are correct we are discussing more about token last used instead of client alive-or-not.

router/router_test.go
eternal-flame-AD: I agree, this was done to make coverage happy,

router/router_test.go
```diff
@@ -400,6 +404,12 @@ func (s *IntegrationSuite) TestAuthentication() {
 	assert.Equal(s.T(), "android-client", token.Name)
 }
 
+func (s *IntegrationSuite) connectWebSocket(url string) *websocket.Conn {
```
jmattheis, line 407: Unused, the changes in this file can probably be reverted.

ui/src/client/Clients.tsx
```diff
@@ -119,6 +123,17 @@ const Row: SFC<IRowProps> = ({name, value, fEdit, fDelete}) => (
                 style={{display: 'flex', alignItems: 'center', width: 200}}
             />
         </TableCell>
+        <TableCell>
```
jmattheis, line 126: The token overlaps the last used column, we could you make this non overlapping?

image

ui/src/application/Applications.tsx
```diff
@@ -172,6 +187,17 @@ const Row: SFC<IRowProps> = observer(
             <TableCell>
                 <CopyableSecret value={value} style={{display: 'flex', alignItems: 'center'}} />
             </TableCell>
+            <TableCell>
```
jmattheis, line 190: I prefer having the last_used column in the column before the controls (delete/edit).

jmattheis (changes requested):

jmattheis (dismissed):
:+1: I've also extracted the ui code for the last used cell to a separate file.

Issue #405, opened 2021-04-08
Icons in sidebar?


YouveGotMeowxy:
- **Is your feature request related to a problem? Please describe.**

I like to  be able to identify items at a quick glance, rather than always have to read.

- **Describe the solution you'd like**

Icons in sidebar would do the trick. The same icons that the user adds, added to the left of the text.

Right now it's just plain ol' boring, text.  :)

- **Additional context**

Icons being the same size as the ones in for example, Organizr would  be good. Bonus points if  the sidebar can auto-close/expand the same way Organizr does as well.

mateuscelio:
@jmattheis, can I take this issue?

jmattheis:
@mateuscelio sure.

PR #602, opened 2023-10-13
Adds applications icons in sidebar

mateuscelio:
It closes https://github.com/gotify/server/issues/405

image

### DIFF
--- a/ui/src/layout/Navigation.tsx
+++ b/ui/src/layout/Navigation.tsx
@@ -66,8 +66,12 @@ class Navigation extends Component<
                           to={'/messages/' + app.id}
                           key={app.id}>
                           <ListItem button>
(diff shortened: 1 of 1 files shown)


ui/src/layout/Navigation.tsx
```diff
@@ -59,6 +66,13 @@ class Navigation extends Component<
                           to={'/messages/' + app.id}
                           key={app.id}>
                           <ListItem button>
+                              <ListItemAvatar style={{minWidth: 42}}>
+                                  <Avatar
+                                      style={{width: 32, height: 32}}
```
jmattheis, line 71: I've changed the sizes a little bit so that it doesn't increase the height of the ListItem.

jmattheis (approved):

Issue #358, opened 2020-09-10
Reorganize the list of "Apps"


EVOTk:
**Is your feature request related to a problem? Please describe.**
On the side, the applications are displayed in the order of creation. I wish I could change this order.

**Describe the solution you'd like**
In Apps, choose an order number for each "Apps"

**Additional context**
![2020-09-10 21_51_26-Gotify](https://user-images.githubusercontent.com/45015615/92794357-d36a3200-f3af-11ea-9869-8f565995baf7.png)
For example, I would like to be able to put "EVOserv" at the top of the list, but for that it would have been necessary to create it first!
Now, for that I will have to delete all the Apps "above", and re-create them, which implies a change of token for this one!

Thx

jmattheis:
Yeah, seems like a good feature. You could also rename the first application, then you don't have to remove all applications.

EVOTk:
Thanks for your feedback :)
Renaming does not solve the problem of the "token" which is linked to an application. Renaming everything is the same as deleting, it will be necessary to change the tokens in the all applications, so that they correspond again.

narol69:
I like the suggestion of EVOTk.

junpet:
Can I hop on the train to ask optional sorting by alphabet?

nocturnalmc:
+1 for this. Currently using gotify to manage 7 apps. Would like to rearrange my application list

Raytestrak:
Bump to an old thread, it can be done via the database. Just reorder the IDs of the applications.

xavier-GitHub76:
The sort is OK with : 
const apps = appStore.getItems().slice().sort((a,b) => a.name.localeCompare(b.name));

Tests OK : 

Applications created in this order :
- Apple
- Orange
- Banana

Items are sorted on screen : 
![image](https://github.com/gotify/server/assets/82046143/2d61123b-404b-45ae-b09c-e1d9dd1e2f36)

samcro1967:
> Bump to an old thread, it can be done via the database. Just reorder the IDs of the applications.

Here is a script to do this automatically:

**gotify_sort_apps_by_names.sh**
```
#!/bin/bash

# Define the database file path
DB_PATH="gotify.db"

# Step 1: Rename the current applications table to applications_old
sqlite3 "$DB_PATH" "ALTER TABLE applications RENAME TO applications_old;"

# Step 2: Create a new applications table with the same schema
[...]

PR #611, opened 2023-11-14
features : Applications sorted by name (Update Navigation.tsx)

xavier-GitHub76:
Before : Applications were sorted by id (so by date of creation).
Now : Applications are sorted by name

this pull request is linked to https://github.com/gotify/server/issues/358

### DIFF
--- a/ui/src/layout/Navigation.tsx
+++ b/ui/src/layout/Navigation.tsx
@@ -54,7 +54,7 @@ class Navigation extends Component<
     public render() {
         const {classes, loggedIn, appStore, navOpen, setNavOpen} = this.props;
         const {showRequestNotification} = this.state;
-        const apps = appStore.getItems();
+        const apps = appStore.getItems().slice().sort((a,b) => a.name.localeCompare(b.name));
 
         const userApps =
             apps.length === 0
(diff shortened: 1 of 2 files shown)


tessus:
I think there should be an option. I like the fact that they are sorted by creation time.
Don't tables usually allow to clich the column header to sort the table? Isn't there a HTML5 option for that? As mentioned before, I suck when it comes to UIs.

ui/src/layout/Navigation.tsx
```diff
@@ -54,7 +54,7 @@ class Navigation extends Component<
     public render() {
         const {classes, loggedIn, appStore, navOpen, setNavOpen} = this.props;
         const {showRequestNotification} = this.state;
-        const apps = appStore.getItems();
```
jmattheis, line 57: This probably should be done in the back end, so that the ordering is consistent with the android app.

ui/src/layout/Navigation.tsx
```diff
@@ -54,7 +54,7 @@ class Navigation extends Component<
     public render() {
```
jmattheis, line 54: I'm not sure if I like the sorting by name. I'd rather have the applications order fully customizable via drag&drop.

PR #620, opened 2024-01-11
Update README.md

nallej:
Guid how to use Gotify on Proxmox and iPhone


jmattheis:
Please add your guide here: https://github.com/gotify/contrib

PR #624, opened 2024-01-20
feat: Trusted Proxies

### LINKED ISSUE #554: Option to log X-Forwarded-For client IP
Gotify logs the client IP address by using only the remote IP address.
But if the requests comes from a reverse proxy, the only IP logged is the reverse proxy IP address.

Reverse proxies typically add headers like X-Forwarded-For containing the real client IP address.
It should be optional though because if there's no reverse proxy, a client may forge a fake X-Forwarded-For header. Gotify can't determine by itself alone if there's a reverse proxy, so the configuration should be explicit.

Viewing the reverse proxy logs instead of gotify logs, but it's not as convenient.

**Additional context**
None

LaurenceJJones:
Fix #554

Added the option to configuration to allow trusted proxies, is handled the same way as every other `[]string`

However, I had to find a workaround for unix sockets since there was no "official" way for Gin to handle these.

So in short if the request comes from unix socket then the `RemoteAddr` is set to `127.0.0.1` so if the user would like to trust a webserver that communicates on the socket then they would need to trust `127.0.0.1` via the new option.

By default the RemoteAddr is set to `@` on unix socket requests , however, gin only allows IP's within the trusted proxies

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -27,6 +27,7 @@ import (
 func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Configuration) (*gin.Engine, func()) {
 	g := gin.New()
 
+	g.RemoteIPHeaders = []string{"X-Forwarded-For"}
 	g.SetTrustedProxies(conf.Server.TrustedProxies)
 	g.ForwardedByClientIP = true
 
--- a/router/router.go
+++ b/router/router.go
@@ -27,10 +27,8 @@ import (
 func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Configuration) (*gin.Engine, func()) {
 	g := gin.New()
 
-	if conf.Server.TrustedProxies != nil {
-		g.SetTrustedProxies(conf.Server.TrustedProxies)
-		g.ForwardedByClientIP = true
-	}
+	g.SetTrustedProxies(conf.Server.TrustedProxies)
+	g.ForwardedByClientIP = true
 
 	g.Use(func(ctx *gin.Context) {
 		// Map sockets "@" to 127.0.0.1, because gin-gonic can only trust IPs.
--- a/router/router.go
+++ b/router/router.go
@@ -27,6 +27,18 @@ import (
 func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Configuration) (*gin.Engine, func()) {
 	g := gin.New()
 
+	if conf.Server.TrustedProxies != nil {
+		g.SetTrustedProxies(conf.Server.TrustedProxies)
+		g.ForwardedByClientIP = true
+	}
+
+	g.Use(func(ctx *gin.Context) {
+		// Map sockets "@" to 127.0.0.1, because gin-gonic can only trust IPs.
+		if ctx.Request.RemoteAddr == "@" {
+			ctx.Request.RemoteAddr = "127.0.0.1:65535"
+		}
+	})
+
 	g.Use(gin.LoggerWithFormatter(logFormatter), gin.Recovery(), gerror.Handler(), location.Default())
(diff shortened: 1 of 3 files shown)


router/router.go
```diff
@@ -27,6 +28,17 @@ import (
 func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Configuration) (*gin.Engine, func()) {
 	g := gin.New()
 
+	if conf.Server.TrustedProxies != nil {
+		g.SetTrustedProxies(conf.Server.TrustedProxies)
```
jmattheis, line 32: Hmm the handling of this in gin really sucks. I'm inclined to not use any of gin for this and do custom client ip parsing. This would also fix the security bug described in https://github.com/gotify/server/issues/554#issuecomment-1560059081

What do you think about this?

```go
gin.LoggerWithFormatter(useXRealIP(conf.Server.TrustXRealIP, logFormatter))
```
```go
func useXRealIP(trustedProxy bool, inner gin.LogFormatter) gin.LogFormatter {
	return func(params gin.LogFormatterParams) string {
		params.ClientIP = getClientIp(trustedProxy, params.Request)
		return inner(params)
	}
}

func getClientIp(trustedProxy bool, req *http.Request) string {
	if trustedProxy {
		realIpParts := strings.SplitN(req.Header.Get("x-real-ip"), ",", 2)
		if ip := strings.TrimSpace(realIpParts[0]); ip != "" {
			return ip
		}
	}

	addr := req.RemoteAddr
	if addr == "@" {
		return "socket"
	} else if host, _, err := net.SplitHostPort(addr); err == nil {
		return host
	}
	return addr
}
```
x-real-ip is used because it's already defined in the gotify reverse proxy docs: https://gotify.net/docs/nginx

LaurenceJJones: Interesting the only issue is that we don't actually test if the upstream proxy is trusted, so if a user misconfigures and exposes the gotify port they could get spoofed.

jmattheis: The same is possible when a user configures 0.0.0.0/0 as trusted proxy and the user still has to enable a trustXRealIP setting. Either way, I don't think this is a big problem, because the ip is only logged and not used for anything else in gotify.

LaurenceJJones: I agree users are users 😅 I'm happy with above since it won't functionally change Gin and just for logging

router/router.go
```diff
@@ -247,3 +248,27 @@ func (fs *onlyImageFS) Open(name string) (http.File, error) {
...
+func useXRealIP(trustedProxy bool, inner gin.LogFormatter) gin.LogFormatter {
+	return func(params gin.LogFormatterParams) string {
+		params.ClientIP = getClientIP(trustedProxy, params.Request)
+		return inner(params)
+	}
+}
+
+func getClientIP(trustedProxy bool, req *http.Request) string {
+	if trustedProxy {
+		realIPParts := strings.SplitN(req.Header.Get("x-real-ip"), ",", 2)
```
jmattheis, line 261: Hmm, it seems like most applications use x-forwarded-for. I'm unsure again if this is a better solution or x-forwarded-for should be used instead. Let me think a little about this .____.

jmattheis: Sorry to change my opinion this many times. X-real-ip seems to be legacy so x-forwarded for seems to be the standard. I changed this back to your original implementation as 
I don't really think it's worth to custom implement this checks for only ip logging, so your workaround with setting the ip to 127.0.0.1 when the connection originates from a socket is good enough.

I've simplified the socket check to `Request.RemoteAddr == "@"`, added the setting to the example config file and now always all SetTrustedProxies even if the list is empty. This is to prevent the bug described in https://github.com/gotify/server/issues/554#issuecomment-1560089638.

Issue #636, opened 2024-03-11
Removal of disabled flag on Refresh button


iarp:
**Is your feature request related to a problem? Please describe.**

Two scenarios:

- I have gotify pinned as a browser tab so i get desktop notifications. If the browser closes for any reason and reopens it uses a cached page result which may or may not have notifications. 
- If I sleep my machine at some point and i know a notification came through (noise from phone) and boot back up, the notification list maybe blank

If no notifications are present then in both scenarios i need to fully refresh the tab. If a notification was present already either in the cached page or before i slept my machine, the Refresh button is enabled and i can click it.

Clicking on the Refresh seems to reconnect the page to the server without having to browser refresh.

**Describe the solution you'd like**

Remove `disabled={!hasMessages}` from the refresh button. Why disable the button in the first place?

PR #637, opened 2024-03-11
change: removal of disabled flag from messages Refresh button

iarp:
In reference to #636

### DIFF
--- a/ui/src/message/Messages.tsx
+++ b/ui/src/message/Messages.tsx
@@ -70,7 +70,6 @@ class Messages extends Component<IProps & Stores<'messagesStore' | 'appStore'>, 
                         <Button
                             id="refresh-all"
                             variant="contained"
-                            disabled={!hasMessages}
                             color="primary"
                             onClick={() => messagesStore.refreshByApp(appId)}
                             style={{marginRight: 5}}>


ui/src/message/Messages.tsx
```diff
@@ -70,7 +70,6 @@ class Messages extends Component<IProps & Stores<'messagesStore' | 'appStore'>,
                         <Button
                             id="refresh-all"
                             variant="contained"
-                            disabled={!hasMessages}
```
jmattheis, line 73: If the webapp is cache-reopened, then it's likely that the websocket connection isn't reconnected. So you will be able to refresh the messages, but this won't reconnect the websocket.

jmattheis (approved):
Thanks. It's weird, that this was disabled when there are no messages (:.

ui/src/message/Messages.tsx
iarp: Could we somehow get the refresh button to check connection status and call upon WebSocketStore.listen if needed? I'm not all familiar with how react works.

jmattheis: Possible yes, but I'd rather solve this automatically on page load but I currently don't know a way to reproduce this cached reopen where the site state is kept.

iarp: Assuming your browser reopens previously opened tabs, open gotify in a tab and close the browser. Reopen the browser and thats it. Or open gotify, leave the browser open and sleep the computer.

jmattheis: I'm doing this, but my firefox correctly reestablishes the websocket connection and has messages that were sent while the browser was closed.

PR #659, opened 2024-06-18
Loading animations instead of "no messages"

### LINKED ISSUE #658: [UI] Show loading indicator when loading messages instead of "No messages"
I have around 10 applications in Gotify, all of them receiving messages regularly. When I want to check them in the UI, I usually do this app by app and not via the "All messages" tab. When switching between application tabs, there is a short amount of time, where the messages are being fetched and the messages panel is empty saying "No messages".

If I want to quickly take a look at all application tabs, sometimes I'm not waiting long enough for the messages to come in and then assume that there are none, because there was no loading indicator.

There should be some kind of loading indicator (e.g. a spinner) when opening an application tab the first time. This should only introduce changes in the ui, not the backend code.

None.

**Additional context**
Not required, I think.

MexHigh:
This PR adds loading animations to the application panels when loading messages, instead of just showing "no messages". I've used the already existing `LoadingSpinner` component for this.

Hint: This PR also changes the size of the `LoadingSpinner` from 150 to 40 (used after login and (since now) on message load). This looks way more modern in my opinion. Feel free to revert the last commit of this PR if you disagree.

Closes #658


MexHigh:
Sorry, wrong branch :D

PR #688, opened 2024-09-18
head method in health endpoint

### LINKED ISSUE #687: HEAD method on /health endpoint throw 404 error
yes

- [x] Docker
- [ ] Linux machine
- [ ] Windows machine

**What did you do?**
actually i'm using uptimerobot service to track uptime which make HEAD request by default to the given healthcheck endpoint (`HEAD /health` endpoint in our case). but as it is getting 404 so it is considering health endpoint to be **not working**.

it should return response with 200 http status code

404 http status code

moinologics:
Fixes #687

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -104,7 +104,7 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 
 	ui.Register(g, *vInfo, conf.Registration)
 
-	g.GET("/health", healthHandler.Health)
+	g.Match([]string{"GET", "HEAD"}, "/health", healthHandler.Health)
 	g.GET("/swagger", docs.Serve)
 	g.StaticFS("/image", &onlyImageFS{inner: gin.Dir(conf.UploadedImagesDir, false)})
 


router/router.go
```diff
@@ -104,7 +104,7 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 
 	ui.Register(g, *vInfo, conf.Registration)
 
-	g.GET("/health", healthHandler.Health)
+	g.Match([]string{"GET", "HEAD"}, "/health", healthHandler.Health)
```
eternal-flame-AD, line 107: This is not compliant with HTTP specs. Please refer to the MDN docs and not return a body. Thanks!

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

eternal-flame-AD (changes requested):
Thanks for the PR. It is weird some company would do a HEAD preflight for a health check service, but good detail to include.

eternal-flame-AD:
@moinologics  I tested this with socat, seems like gin internally strips the body if the verb was HEAD, so this code produces a compliant behavior (except dynamically generated JSON should not have a content-type on HEAD). However I still think this is not ideal, we shouldn't rely on undocumented error-catching features and write code with a looks-wrong semantics.

Feel free to @ jmattheis  for a second opinion :) Everybody has different styles and preferences

eternal-flame-AD:
Here's a solution I think is robust, (maybe even a little overkill?) , but at the very least I think the content type and content-encoding must be the same, and content length must not be present.

https://github.com/gotify/server/commit/b43e7cdc8ccdea2c09042704c05f5c3561e96d1f

router/router.go
jmattheis: @eternal-flame-AD I would merge this as is. The RFC states:

> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. However, a server MAY omit header fields for which a value is determined only while generating the content. 

I'd see it as recommendation to include content-length, but you can/MAY optionally omit it in some circumstances. So the current implementation complies with the recommendation.

> I tested this with socat, seems like gin internally strips the body if the verb was HEAD

I think it's actually the native go http server, and it makes sense, as you need the body to get the content-length and given HEAD requests must not have a body I'd see this as working as intended and not as error-catching feature.

eternal-flame-AD (approved):
Thanks.

Cc @moinologics

PR #690, opened 2024-09-20
bug: fixed a race condition in database test

eternal-flame-AD:
This won't work if two tests happened on the same second


jmattheis (approved):
:+1:

Issue #699, opened 2024-10-15
Automatic docker releases


jmattheis:
Gotify depends on the debian-slim image which should be updated regularly to get the most recent bug/security fixes. We should add a github action that automatically releases gotify to use the latest base image.

eternal-flame-AD:
The multi arch image I was working on already has configurable base image on the docker build arguments. I'm not sure how to integrate it into the https://github.com/gotify/build ecosystem though. Maybe put this configuration logic there, and we tag the container by release version (and alias the newest tag to the current unversioned tags?)

https://github.com/gotify/server/blob/7c174336ffc9da237998ec4d9d20d1c2a83fc5fb/docker/Dockerfile#L1

jmattheis:
> Maybe put this configuration logic there, and we tag the container by release version (and alias the newest tag to the current unversioned tags?)

Could you elaborate? I don't quite understand what you mean or how this is related to this ticket. Do you mean re-releasing an existing version e.g. 2.5.0 with an updated base image?

eternal-flame-AD:
The concern is about compatibility with existing plugins which uses the above image and are not aware of base image differences.

The current plugin template only considered go version but not the base image version. I think we should have both the traditional tag and a different tag that looks like gotify-v1.1 (we don't need arch anymore because it will be multi arch) and alias the newest tag to the traditional go-version-only tag syntax. (In practice, if you release v1.1.1 which only update the container itself, you reset the legacy tag of the same to version here.

https://github.com/gotify/plugin-template/blob/7f7ad4153075f8656be1e1642040d300fdb1057c/Makefile#L34

jmattheis:
> The current plugin template only considered go version but not the base image version. I think we should have both the traditional tag and a different tag that looks like gotify-v1.1 (we don't need arch anymore because it will be multi arch) and alias the newest tag to the traditional go-version-only tag syntax. (In practice, if you release v1.1.1 which only update the container itself, you reset the legacy tag of the same Go version here.

Sorry, still don't get it, do you want to make the gotify/build image multi arch? And do you mean releasing one gotify/build image with an updated base image?

I don't think we really need multiple releases for one Go version in gotify/build as Go is released every month, so we can go along this pattern and have ~1month up-to-date build images.

[...]

PR #702, opened 2024-10-15
Add bump-go action

eternal-flame-AD:
Rel: #699

Current logic:

If the stable Go version does not equal GO_VERSION on master and does not equal the version on the bump-go branch then create or update a PR.

Whether to go for a fully automated release I think let's wait and see for maybe 2 cycles and if minimal intervention is required then we can make this automated. I think this should have already taken off a large portion of wasted time for releases which is waiting for CI to finish..

For a fully automated workflow I think the general idea of what to do next is:
- Pull in updates daily if master is updated but this branch is not.
- Detect a review submission and after some sanity checks (no new pushes to master, etc), merge to master and generate a new tag automatically using the current release infrastructure.

### DIFF
--- a/ci/bump_go.sh
+++ b/ci/bump_go.sh
@@ -6,11 +6,13 @@ PR_LABEL="bump-go"
 
 set -e
 
+# Check if the working directory is clean
 if [ ! -z "$(git status -s -uall)" ]; then
     echo "Working directory is not clean" 2>&1
     exit 1
 fi
 
+# Create the bump-go branch if it doesn't exist
 git show-ref --verify --quiet refs/heads/bump-go || git branch -c $REF_BRANCH bump-go
 
 # The version in the GO_VERSION file
@@ -25,8 +27,10 @@ echo "Installed version: $latest_version"
 bump_candidate_version=$(git show bump-go:GO_VERSION 2>/dev/null || echo "")
 echo "Bump candidate version: $bump_candidate_version"
 
+# Lookup number of existing PRs
 existing_prs=$(gh pr list --state open --base $REF_BRANCH --label "$PR_LABEL" --json "number" | jq -r 'map(.number) | .[]')
 
+# Check if the version is up to date
 if [ "$current_version" == "$latest_version" ]; then
     echo "Go is up to date"
     exit 0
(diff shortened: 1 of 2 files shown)


eternal-flame-AD:
What do you think of this plan? @jmattheis 

To test the current solution on actual GH actions, merge, close #701, delete the bump-go branch, then trigger the workflow by dispatch or cron.

ci/bump_go.sh
```diff
@@ -0,0 +1,80 @@
...
+    
+    git push origin bump-go
+
+    if [ -z "$existing_prs" ]; then
+        gh pr create --base $REF_BRANCH --head bump-go \
+            --title "$PR_KEYWORD Bump Go to $latest_version" --label "$PR_LABEL" \
+            --body "This PR was automatically created by the bump-go action."
+    else
+        first_id=$(echo "$existing_prs" | sort -nr | head -n 1)
+        gh pr edit $first_id --title "$PR_KEYWORD Bump Go to $latest_version"
```
jmattheis, line 77: Have you considered using renovate / dependabot for this https://docs.renovatebot.com/modules/manager/regex/. I haven't used this specific feature of renovate, but renovate should provide the logic for opening PRs when there is a new version, and updating existing PRs with newer versions. I think the config could look like this
```json
{
  "regexManagers": [
    {
      "fileMatch": ["^GO_VERSION$"],
      "matchStrings": ["^(?<currentValue>\\d+\\.\\d+\\.\\d+)$"],
      "datasourceTemplate": "docker",
      "depNameTemplate": "golang"
    }
  ]
}
```
We probably could reference the gotify/build image there, so that it would only create a PR when there is a newer gotify/build image. We could then also add renovate to gotify build to update when there are newer golang docker images.

Generally, I'd rather use these tools than custom scripts as they seem easier to maintain, and I think gotify could also benefit from automatic dep updates (at least for the backend).

eternal-flame-AD: Yeah makes sense I support this, although you are in a better position to introduce a new tool like this (I think I don't even have the permission to do it).

jmattheis: I've added renovate to gotify/server and gotify/build. It needs configuration in the repo, there we probably have to only whitelist Go / docker, so that it doesn't update the js deps for now.

eternal-flame-AD:
Replaced by #706

PR #703, opened 2024-10-16
Docker multiarch image 

### LINKED ISSUE #257: Multi-Arch Docker Image

### LINKED ISSUE #350: Add curl package on ARM Dockerfile

eternal-flame-AD:
fixes #257 .
fixes #350 

Simply `make build-docker-multiarch` and then `docker run` should work. I implemented building JS from docker as well but opt-in, just for people who want's to build a usable local container without having to do anything outside of docker.

Current blockers: (nevermind I had a nap and figured out how to do it)

The last step of installing runtime dependency need qemu, it's not easy to get around that unfortunately

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,4 @@ coverage.txt
 data/
 images/
 .git/
+*/node_modules/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+vendor/
+.idea/
+build/
+licenses/
+coverage.txt
+data/
+images/
+.git/
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,6 @@ package-zip: extract-licenses
 
 build-docker-multiarch: require-version
 	docker buildx build --sbom=true --provenance=true \
-		--output type=local,dest=container-out \
 		-t gotify/server:latest \
 		-t gotify/server:${VERSION} \
 		-t gotify/server:$(shell echo $(VERSION) | cut -d '.' -f -2) \
--- a/Makefile
+++ b/Makefile
@@ -105,13 +105,6 @@ build-docker-multiarch: require-version
 		--platform linux/amd64,linux/arm64,linux/386,linux/arm/v7,linux/riscv64 \
 		-f docker/Dockerfile .
 
-# Backwards compatibility
-build-docker-amd64: build-docker-multiarch
-build-docker-i386: build-docker-multiarch
-build-docker-arm64: build-docker-multiarch
-build-docker-arm-7: build-docker-multiarch
-build-docker-riscv64: build-docker-multiarch
-
 build-docker: build-docker-multiarch
 
 _build_within_docker: OUTPUT = gotify-app
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ package-zip: extract-licenses
 
 build-docker-multiarch: require-version
 	docker buildx build --sbom=true --provenance=true \
+		$(if $(DOCKER_BUILD_PUSH),--push) \
 		-t gotify/server:latest \
 		-t gotify/server:${VERSION} \
 		-t gotify/server:$(shell echo $(VERSION) | cut -d '.' -f -2) \
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ DOCKER_BUILD_IMAGE=gotify/build
 DOCKER_WORKDIR=/proj
 DOCKER_RUN=docker run --rm -v "$$PWD/.:${DOCKER_WORKDIR}" -v "`go env GOPATH`/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
 DOCKER_GO_BUILD=go build -mod=readonly -a -installsuffix cgo -ldflags "$$LD_FLAGS"
+DOCKER_TEST_LEVEL ?= 1 # should be unnecessary but good for a last-line of defense
 NODE_OPTIONS=$(shell if node --help | grep -q -- "--openssl-legacy-provider"; then echo --openssl-legacy-provider; fi)
 
 test: test-coverage test-js
@@ -64,7 +65,8 @@ package-zip: extract-licenses
     done
 
 build-docker-multiarch: require-version
-	docker buildx build --sbom=true \
+	docker buildx build --sbom=true --provenance=true \
+		--output type=local,dest=container-out \
 		-t gotify/server:latest \
 		-t gotify/server:${VERSION} \
 		-t gotify/server:$(shell echo $(VERSION) | cut -d '.' -f -2) \
@@ -81,20 +83,6 @@ build-docker-multiarch: require-version
 		-t ghcr.io/gotify/server-arm64:${VERSION} \
 		-t ghcr.io/gotify/server-arm64:$(shell echo $(VERSION) | cut -d '.' -f -2) \
 		-t ghcr.io/gotify/server-arm64:$(shell echo $(VERSION) | cut -d '.' -f -1) \
-		--build-arg GO_VERSION=$(shell cat GO_VERSION) \
-		--platform linux/amd64,linux/arm64,linux/386 \
-		-f docker/Dockerfile .
-
-build-docker-amd64: build-docker-multiarch
-
-build-docker-i386: build-docker-multiarch
-
-build-docker-arm64: build-docker-multiarch
-
-build-docker-arm-7: require-version
-	cp ${BUILD_DIR}/gotify-linux-arm-7 ./docker/gotify-app
-	cd ${DOCKER_DIR} && \
-		docker build -f Dockerfile.armv7 \
 		-t gotify/server-arm7:latest \
 		-t gotify/server-arm7:${VERSION} \
 		-t gotify/server-arm7:$(shell echo $(VERSION) | cut -d '.' -f -2) \
@@ -102,13 +90,7 @@ build-docker-arm-7: require-version
 		-t ghcr.io/gotify/server-arm7:latest \
 		-t ghcr.io/gotify/server-arm7:${VERSION} \
 		-t ghcr.io/gotify/server-arm7:$(shell echo $(VERSION) | cut -d '.' -f -2) \
-		-t ghcr.io/gotify/server-arm7:$(shell echo $(VERSION) | cut -d '.' -f -1) .
-	rm ${DOCKER_DIR}gotify-app
-
-build-docker-riscv64: require-version
-	cp ${BUILD_DIR}/gotify-linux-riscv64 ./docker/gotify-app
-	cd ${DOCKER_DIR} && \
-		docker build -f Dockerfile.riscv64 \
+		-t ghcr.io/gotify/server-arm7:$(shell echo $(VERSION) | cut -d '.' -f -1) \
 		-t gotify/server-riscv64:latest \
 		-t gotify/server-riscv64:${VERSION} \
 		-t gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -2) \
@@ -116,10 +98,20 @@ build-docker-riscv64: require-version
 		-t ghcr.io/gotify/server-riscv64:latest \
 		-t ghcr.io/gotify/server-riscv64:${VERSION} \
 		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -2) \
-		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -1) .
-	rm ${DOCKER_DIR}gotify-app
+		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -1) \
+		--build-arg RUN_TESTS=$(DOCKER_TEST_LEVEL) \
+		--build-arg GO_VERSION=$(shell cat GO_VERSION) \
+		--platform linux/amd64,linux/arm64,linux/386,linux/arm/v7,linux/riscv64 \
+		-f docker/Dockerfile .
 
-build-docker: build-docker-multiarch build-docker-arm-7 build-docker-riscv64
+# Backwards compatibility
+build-docker-amd64: build-docker-multiarch
+build-docker-i386: build-docker-multiarch
+build-docker-arm64: build-docker-multiarch
+build-docker-arm-7: build-docker-multiarch
+build-docker-riscv64: build-docker-multiarch
+
+build-docker: build-docker-multiarch
 
 build-js:
 	(cd ui && NODE_OPTIONS="${NODE_OPTIONS}" yarn build)
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,9 @@ SHELL := /bin/bash
 GO_VERSION=`cat GO_VERSION`
 DOCKER_BUILD_IMAGE=gotify/build
 DOCKER_WORKDIR=/proj
-DOCKER_RUN=docker run --rm -v "$$PWD/.:${DOCKER_WORKDIR}" -v "`go env GOPATH`/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
+DOCKER_RUN=docker run --rm -e LD_FLAGS="$$LD_FLAGS" -v "$$PWD/.:${DOCKER_WORKDIR}" -v "`go env GOPATH`/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
 DOCKER_GO_BUILD=go build -mod=readonly -a -installsuffix cgo -ldflags "$$LD_FLAGS"
-DOCKER_TEST_LEVEL ?= 1 # should be unnecessary but good for a last-line of defense
+DOCKER_TEST_LEVEL ?= 0 # Optionally run a test during docker build
 NODE_OPTIONS=$(shell if node --help | grep -q -- "--openssl-legacy-provider"; then echo --openssl-legacy-provider; fi)
 
 test: test-coverage test-js
@@ -101,6 +101,7 @@ build-docker-multiarch: require-version
 		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -1) \
 		--build-arg RUN_TESTS=$(DOCKER_TEST_LEVEL) \
 		--build-arg GO_VERSION=$(shell cat GO_VERSION) \
+		--build-arg LD_FLAGS="$$LD_FLAGS" \
 		--platform linux/amd64,linux/arm64,linux/386,linux/arm/v7,linux/riscv64 \
 		-f docker/Dockerfile .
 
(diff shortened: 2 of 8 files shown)


eternal-flame-AD:
Here's the Docker scout cloud analysis, you should be able to reproduce it with `docker scout` CLI.

<details>
image
image
</details>

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
-FROM amd64/debian:stable-slim
-ENV GOTIFY_SERVER_PORT="80"
+ARG BUILDKIT_SBOM_SCAN_CONTEXT=true
```
jmattheis, line 1: I like having this all in one Dockerfile, but I dislike that the binaries for the linux platforms are built twice (one for the GH release, one for docker). I'd really like that the binaries for the same arch/os are the same for docker and the binary release.

This is probably a nice to have, as I don't really see a good / easy way for this. The binary could be copied from the built image, but this also feels a little weird.

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+    mkdir -p /target/app && \
+    if [ "$RUN_TESTS" = "1" ] && [ "$BUILDPLATFORM" = "$TARGETPLATFORM" ]; then \
+    go test -v ./...; \
+    fi && \
+    go build ${GO_BUILD_FLAGS} -o /target/app/gotify-app
+
+FROM debian:${DEBIAN}
+
+# build-time configurables
+ARG GOTIFY_SERVER_EXPOSE=80
```
jmattheis, line 81: I rather not have this configurable, I mean it's possible to adjust this for an already built container, and we don't use this setting.

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+RUN if [ "$BUILD_JS" = "1" ]; then \
+    cp -r --update /ui-build /src/gotify/ui/build; \
+    fi
+
+RUN cd /src/gotify && \
+    mkdir -p /target/app && \
+    if [ "$RUN_TESTS" = "1" ] && [ "$BUILDPLATFORM" = "$TARGETPLATFORM" ]; then \
+    go test -v ./...; \
+    fi && \
+    go build ${GO_BUILD_FLAGS} -o /target/app/gotify-app
```
jmattheis, line 76: I feel like the docker and non docker build process should share more code via the makefile. I mean here we could execute something like `make build-bin-locally` or similar, so that the build commands are exactly the same.

Makefile
```diff
@@ -63,63 +64,54 @@ package-zip: extract-licenses
...
 		-t ghcr.io/gotify/server-arm64:$(shell echo $(VERSION) | cut -d '.' -f -2) \
-		-t ghcr.io/gotify/server-arm64:$(shell echo $(VERSION) | cut -d '.' -f -1) .
-	rm ${DOCKER_DIR}gotify-app
-
-build-docker-riscv64: require-version
-	cp ${BUILD_DIR}/gotify-linux-riscv64 ./docker/gotify-app
-	cd ${DOCKER_DIR} && \
-		docker build -f Dockerfile.riscv64 \
+		-t ghcr.io/gotify/server-arm64:$(shell echo $(VERSION) | cut -d '.' -f -1) \
+		-t gotify/server-arm7:latest \
```
jmattheis, line 86: (only FYI, should be done separately) I think I'd like to deprecate these. So that they maybe just output "use the gotify/server image, this image is deprecated".

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+    \
+    cp -r /src/gotify/ui/build /target; \
+    \
+    else \
+    mkdir -p /target; \
+    fi
+
+# --- Go Builder ---
+
+FROM --platform=${BUILDPLATFORM} gotify/build:${GO_VERSION}-${__TARGETPLATFORM_GO_NOTATION} AS builder
```
jmattheis, line 51: > The last step of installing runtime dependency need qemu, it's not easy to get around that unfortunately

Yeah, at least it's only the last step, I tried compiling a go binary via qemu and this was soooo slow. Here it's negligible. Nicely done (:.

Makefile
```diff
@@ -63,63 +64,54 @@ package-zip: extract-licenses
...
+		--build-arg GO_VERSION=$(shell cat GO_VERSION) \
+		--platform linux/amd64,linux/arm64,linux/386,linux/arm/v7,linux/riscv64 \
+		-f docker/Dockerfile .
+
+# Backwards compatibility
+build-docker-amd64: build-docker-multiarch
+build-docker-i386: build-docker-multiarch
+build-docker-arm64: build-docker-multiarch
+build-docker-arm-7: build-docker-multiarch
+build-docker-riscv64: build-docker-multiarch
```
jmattheis, line 112: I think this can be removed. We don't really have to keep backwards compat in the makefile.

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+
+ARG BUILDPLATFORM
+ARG TARGETPLATFORM
+ARG BUILD_JS=0
```
jmattheis, line 55: Do we have to define this BULID_JS ARG twice, or can this be moved to the top, so we only have one?

jmattheis (changes requested):
Nicely done, some remarks in subcomments.

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+ARG GO_VERSION=PLEASE_PROVIDE_GO_VERSION
+
+# Hack to normalize platform to match the chosed build image
+# Get the gotify/build image tag
+ARG __TARGETPLATFORM_DASHES=${TARGETPLATFORM/\//-}
+ARG __TARGETPLATFORM_GO_NOTATION=${__TARGETPLATFORM_DASHES/arm\/v7/arm-7}
+# Set the unstable tag if the target platform is not stable
+ARG __TARGETPLATFORM_IS_STABLE=${TARGETPLATFORM/linux\/riscv64/}
+ARG __DEBIAN_HAS_STABLE_TAG=${__TARGETPLATFORM_IS_STABLE/?*/${DEBIAN_STABLE_ARCH}} # if the last variable is not empty, set it to the debian stable tag
+ARG DEBIAN=${__DEBIAN_HAS_STABLE_TAG:-${DEBIAN_EXPERIMENTAL_ARCH}} # else set it to the debian experimental tag
```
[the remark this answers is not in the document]
eternal-flame-AD: Riscv doesn't have slim ... I was pretty annoyed I had to do this too since docker doesn't have real "decision making" in dockerfile

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+    ca-certificates && \
+    rm -rf /var/lib/apt/lists/* && \
+    if [ "$GOTIFY_UID" != "0" ]; then \
+    useradd -u $GOTIFY_UID -r gotify && \
+    chown gotify:gotify /app; \
+    fi
+
+USER $GOTIFY_UID
+
 HEALTHCHECK --interval=30s --timeout=5s --start-period=5s CMD curl --fail http://localhost:$GOTIFY_SERVER_PORT/health || exit 1
```
jmattheis, line 99: This multiarch build also fixes #350, nice!

docker/Dockerfile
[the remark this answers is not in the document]
eternal-flame-AD: Sure, I was writing this to preserve the same behavior of current builds but if that's not an issue we can just simply this :)

docker/Dockerfile
eternal-flame-AD: It's a quirk in dockerfile, of you use an argument in different containers you have to declare them twice or they won't be available:(

docker/Dockerfile
eternal-flame-AD: You can build the image, create a dummy container and `docker cp` it out I suppose?  I agree your behavior is better. I think this is the easiest fix?

jmattheis: Yeah seems better. It would probably be cool we don't have to built the full multiarch image for this. E.g. I run build-bin-arm64, then it only runs with --platform=linux/arm64. But we should probably wait until someone requests this, as I don't think many need this feature.

docker/Dockerfile
```diff
@@ -1,12 +1,104 @@
...
+    if [ "$RUN_TESTS" = "1" ] && [ "$BUILDPLATFORM" = "$TARGETPLATFORM" ]; then \
+    go test -v ./...; \
+    fi && \
+    go build ${GO_BUILD_FLAGS} -o /target/app/gotify-app
+
+FROM debian:${DEBIAN}
+
+# build-time configurables
+ARG GOTIFY_SERVER_EXPOSE=80
+ARG GOTIFY_UID=0
```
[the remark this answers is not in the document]
eternal-flame-AD: I feel this doesn't really hurt (just 2 lines) and additionally prevents misconfiguration by overlaying a config.yml that have inconsistent `port:` values inside.

docker/Dockerfile
eternal-flame-AD: I wrote an internal recipe to keep the build commands and LD_FLAGS the same. I think it is fine to just build twice.

docker/Dockerfile
eternal-flame-AD: I think this is okay for two lines, first it removes the magic :80 in the healthcheck, additionally the environment variable prevents misconfiguration where a user overlays a config.yml that says a different port. I think this is more flexible and worth the two lines.

docker/Dockerfile
```diff
@@ -1,12 +1,85 @@
...
+
+COPY . /src/gotify
+COPY --from=js-builder /target /ui-build
+
+RUN if [ "$BUILD_JS" = "1" ]; then \
+    cp -r --update /ui-build /src/gotify/ui/build; \
+    fi
+
+RUN cd /src/gotify && \
+    mkdir -p /target/app && \
```
jmattheis, line 60: This line isn't needed, go build will create this if it does not exist.

docker/Dockerfile
```diff
@@ -1,12 +1,85 @@
...
+ARG __TARGETPLATFORM_GO_NOTATION=${__TARGETPLATFORM_DASHES/arm\/v7/arm-7}
+
+# --- JS Builder ---
+
+FROM --platform=${BUILDPLATFORM} node:23 AS js-builder
+
+ARG BUILD_JS=0
+ARG NODE_OPTIONS
+
+COPY . /src/gotify
```
jmattheis, line 18: (optional) I think the cache ability  could be improved. Currently any change will cause a full rebuild of the image, because of the blank copies. E.g. this could be done for the go part:
```
COPY ./go.* /src/gotify/
RUN (cd /src/gotify && go mod download)
```
and this for node
```
COPY ./ui/package.json ./ui/yarn.lock /src/gotify/ui/
RUN [ "$BUILD_JS" = "1" ] && (cd /src/gotify/ui && yarn install)
```
so that the deps downloads are cached as the yarn install is pretty slow.

docker/Dockerfile
jmattheis: :+1:

docker/Dockerfile
```diff
@@ -1,12 +1,85 @@
...
+    export NODE_OPTIONS=${NODE_OPTIONS:-$NODE_OPTIONS_DEFAULT} && \
+    echo "Using NODE_OPTIONS=$NODE_OPTIONS" && \
+    cd /src/gotify/ui && \
+    \
+    npm install -g yarn && \
+    \
+    yarn install && \
+    yarn build && \
+    \
+    cp -r /src/gotify/ui/build /target; \
```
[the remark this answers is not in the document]
eternal-flame-AD: Sure, usually I like to decouple path from within the source tree in the build and copy everything I need to /target to clearly mark it, but if you want shorter this is better.

docker/Dockerfile
```diff
@@ -1,12 +1,85 @@
...
+# Get the gotify/build image tag
+ARG __TARGETPLATFORM_DASHES=${TARGETPLATFORM/\//-}
+ARG __TARGETPLATFORM_GO_NOTATION=${__TARGETPLATFORM_DASHES/arm\/v7/arm-7}
+
+# --- JS Builder ---
+
+FROM --platform=${BUILDPLATFORM} node:23 AS js-builder
+
+ARG BUILD_JS=0
+ARG NODE_OPTIONS
```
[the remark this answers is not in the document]
eternal-flame-AD: I was thinking too ahead of myself.. I though what if there is a different "legacy provider" workaround needed in the future, but is is already in docker pinned to a specific base image I don't think this is an issue.

docker/Dockerfile
eternal-flame-AD: I removed the dependency from JS to Go code but the other way around I think the dependency still exists because go build will package the JS code.

If BUILD_JS=0 the only overhead is pulling the node image once for the whole build, I think it is fine (Dockerfile should really be scriptable...)

If BUILD_JS=1 docker is smart enough to know there is only one JS build needed cross the whole build.

jmattheis (dismissed):
:+1:

.dockerignore
```diff
@@ -0,0 +1,8 @@
+vendor/
```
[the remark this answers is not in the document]
eternal-flame-AD: I was thinking you can save some useless download by pre installing dependencies at the cost of some needless cache invalidation, but I think the probability of actual hitting the problem is low in real development is lower in the latter case.

In the current solution if you want to build the container locally while testing the BUILD_JS you can pre install the dependencies to make it much quicker.

In the other way around I think theoretically you can have a situation where you have a large amount of arch-specific node modules but that's probably not currently true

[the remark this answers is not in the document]
jmattheis: For me the copying of the node_modules folder takes ~50s. Which is basically the same as installing the deps for me.

With node_modules
```
 => [internal] load build definition from Dockerfile                     ... 0.0s
 => => transferring dockerfile: 2.07kB                                   ... 0.0s
 => [internal] load metadata for docker.io/library/debian:sid-slim       ... 1.0s
 => [internal] load metadata for docker.io/library/node:23               ... 1.0s
 => [internal] load metadata for docker.io/gotify/build:1.22.4-linux-amd6... 0.0s
 => [auth] library/node:pull token for registry-1.docker.io              ... 0.0s
 => [auth] library/debian:pull token for registry-1.docker.io            ... 0.0s
 => [internal] load .dockerignore                                        ... 0.0s
 => => transferring context: 105B                                        ... 0.0s
 => [internal] load build context                                        ... 9.0s
 => => transferring context: 801.08MB                                    ... 8.8s
 => [js-builder 1/4] FROM docker.io/library/node:23@sha256:9d09fa506f5b84... 0.0s
 => [stage-2 1/4] FROM docker.io/library/debian:sid-slim@sha256:7bee69f27... 0.0s
 => CACHED [builder 1/7] FROM docker.io/gotify/build:1.22.4-linux-amd64  ... 0.0s
 => [builder 2/7] RUN apt-get update && apt-get install -yq --no-install-... 5.9s
 => [builder 3/7] RUN ls -lah                                            ... 0.3s
 => [builder 4/7] COPY . /src/gotify                                     ...47.9s
 => CACHED [js-builder 2/4] COPY ./Makefile /src/gotify/Makefile         ... 0.0s
 => [js-builder 3/4] COPY ./ui /src/gotify/ui                            ...47.8s
 => [js-builder 4/4] RUN if [ "0" = "1" ]; then     (cd /src/gotify/ui &&... 0.4s
 => [builder 5/7] COPY --from=js-builder /src/gotify/ui/build /ui-build  ... 0.3s
 => CANCELED [builder 6/7] RUN if [ "0" = "1" ]; then     cp -r --update ... 0.5s
```
With ignored node_modules
```
 => [internal] load build definition from Dockerfile                     ...0.1s
 => => transferring dockerfile: 2.07kB                                   ...0.0s
 => [internal] load metadata for docker.io/library/debian:sid-slim       ...0.5s
 => [internal] load metadata for docker.io/library/node:23               ...0.5s
 => [internal] load metadata for docker.io/gotify/build:1.22.4-linux-amd6...0.0s
 => [internal] load .dockerignore                                        ...0.1s
 => => transferring context: 122B                                        ...0.0s
 => [stage-2 1/4] FROM docker.io/library/debian:sid-slim@sha256:7bee69f27...0.0s
 => [internal] load build context                                        ...1.7s
 => => transferring context: 20.23kB                                     ...1.6s
 => [js-builder 1/4] FROM docker.io/library/node:23@sha256:9d09fa506f5b84...0.0s
 => [builder 1/7] FROM docker.io/gotify/build:1.22.4-linux-amd64         ...0.0s
 => CACHED [builder 2/7] RUN apt-get update && apt-get install -yq --no-i...0.0s
 => CACHED [builder 3/7] RUN ls -lah                                     ...0.0s
 => CACHED [js-builder 2/4] COPY ./Makefile /src/gotify/Makefile         ...0.0s
 => CACHED [js-builder 3/4] COPY ./ui /src/gotify/ui                     ...0.0s
 => CACHED [js-builder 4/4] RUN if [ "0" = "1" ]; then     (cd /src/gotif...0.0s
 => [builder 4/7] COPY . /src/gotify                                     ...2.3s
 => [builder 5/7] COPY --from=js-builder /src/gotify/ui/build /ui-build  ...0.2s
 => [builder 6/7] RUN if [ "0" = "1" ]; then     cp -r --update /ui-build...0.3s
 => CANCELED [builder 7/7] RUN cd /src/gotify &&     if [ "0" = "1" ] && ...4.2s
```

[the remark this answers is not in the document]
jmattheis: okay, maybe nvm, my storage seems pretty slow.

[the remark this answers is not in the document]
eternal-flame-AD: I ran on act (local GitHub actions clone) it wasn't too bad (10 seconds around if I recall). But it heavily depends on file system optimizations, I have an SSD with XFS, if you have an HDD with something like ext4 that may be very slow.

I will push an update later and make this more stable, if you have this issue other people may have it to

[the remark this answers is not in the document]
jmattheis: Okay, yeah I've an SSD with ext4, need to debug this. Maybe it's some docker specific thing
```
$ dd if=/dev/zero of=foo conv=fdatasync bs=8k count=50k

419430400 bytes (419 MB, 400 MiB) copied, 4.52201 s, 92.8 MB/s
$ hdparm -Tt /dev/sda1
$ sudo hdparm -Tt /dev/sda1
/dev/sda1:
 Timing cached reads:   22502 MB in  2.00 seconds = 11278.21 MB/sec
 Timing buffered disk reads: 500 MB in  0.93 seconds = 538.35 MB/sec
```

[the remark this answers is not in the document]
eternal-flame-AD: https://github.com/gotify/server/actions/runs/11467399224/job/31910194239

You are right, it is faster to not have it.

Issue #716, opened 2024-11-01
fix web UI responsiveness


david-kalmakoff:
**Is your feature request related to a problem? Please describe.**
Currently the web UI responsiveness on mobile is broken. The overflow on the tables and word wrapping causes the UI to be extended past the screen.

**Describe the solution you'd like**
Adding x-overflow and word-wrapping as well as some minor padding/margin changes, the web UI can be much more usable on mobile.

**Additional context**
I have included an image as an example of the issue.

This is something I would be more than happy to implement.
I already have the changes done.

![image](https://github.com/user-attachments/assets/51a427b7-ce30-4c25-ac12-e557197335f1)

eternal-flame-AD:
Sounds like a good thing to have :) Probably the jmattheis will review it as I know nothing about web layouts, if you want a real green light mention him.

jmattheis:
:+1:

PR #720, opened 2024-11-02
update mobile responsiveness on web UI

david-kalmakoff:
Fixes https://github.com/gotify/server/issues/716

the following pages had responsiveness issues:
- applications (card overflow)
- clients (card overflow)
- plugins (card overflow, info card word wrapping, and card padding)
- individual plugin (info card word wrapping, and card spacing)


jmattheis (approved):
Thanks, it does improve it. I think it could be improved more by reducing the big padding / margin to the screen borders when on smaller devices.

Issue #721, opened 2024-11-02
have router navigate to expected route on page loads


david-kalmakoff:
**Is your feature request related to a problem? Please describe.**
If you are on a page other than the home page and refresh, you are brought back to the homepage.
If you paste in a url such as `https://<domain>/#/plugins` it brings you to the home page instead of that page.
It seems that the router navigates the `/login` then `/` on any page loads.

**Describe the solution you'd like**
It would be great if the router brought you to the expected page on page loads.
I would have to look some more into how the router is implemented to find a more specific solution.

**Additional context**
This is something I would be happy to dive into if others find value in it.

jmattheis:
Yeah, feel free to fix this, I see this as a bug.

It's likely that we only have to set authenticating = true, when calling tryAuthenticate in https://github.com/gotify/server/blob/master/ui/src/CurrentUser.ts#L101 (and this.authenticating = false when it's finished) as the Layout.tsx has this as first route
```
{authenticating ? (
    <Route path="/">
        <LoadingSpinner />
    </Route>
) : null}
```

PR #723, opened 2024-11-05
bug: fixed router navigating to homepage on page loads

david-kalmakoff:
Fixes https://github.com/gotify/server/issues/721

- updated `authenticating` status when `tryAuthenticating` is called
- changed a test to navigate with window location to make sure you can navigate to pages after page loads


ui/src/CurrentUser.ts
```diff
@@ -99,7 +99,9 @@ export class CurrentUser {
     };
```
jmattheis, line 99: FYI: your commits aren't linked to your GitHub account. See the email in https://github.com/gotify/server/pull/723/commits/eaa19ddcb110296b7fd49e8589f2c5a3b22b2b55.patch

ui/src/CurrentUser.ts
```diff
@@ -99,7 +99,9 @@ export class CurrentUser {
...
 
     public tryAuthenticate = async (): Promise<AxiosResponse<IUser>> => {
+        this.authenticating = true;
```
jmattheis: I've removed this line, as I think we don't want to show the loading spinner, when we try to reconnect after there was a successful connection. I've set this.authenticating to true on variable initialization, so it should behave the same.

ui/src/CurrentUser.ts
```diff
@@ -99,7 +99,9 @@ export class CurrentUser {
     };
 
     public tryAuthenticate = async (): Promise<AxiosResponse<IUser>> => {
```
jmattheis, line 95: The plugin detail page threw an exception because the plugin data wasn't loaded there.

jmattheis (approved):

PR #733, opened 2024-11-15
fix: enable silent mode

jmattheis:
Hides this log output

    Failed to find configuration /etc/gotify/config.yml
    Failed to find configuration config.yml, using example file config.example.yml

Introduced with the #727

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -66,7 +66,7 @@ func configFiles() []string {
 // Get returns the configuration extracted from env variables or config file.
 func Get() *Configuration {
 	conf := new(Configuration)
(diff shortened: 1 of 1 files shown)


eternal-flame-AD:
I actually intentionally didn't write silent: true... I thought it might not be a good idea someone can just docker run and magically have their server listening on port 80 with admin:admin as the password

jmattheis:
This is only log, the server starts correctly listening on port 80 with admin:admin in docker without any config. And given that gotify/server should be configured via env vars in docker, the error/warning is just misleading.

jmattheis:
Tho, I'm not really against logging stuff like that, just wanted to have the previous behavior for this release as I hadn't noticed this in the review.

PR #780, opened 2025-03-11
Cleaner authority of go version update (go -> gotify/build -> go.mod + GO_VERSION)

### LINKED ISSUE #779: Coordinate GOTOOLCHAIN with the legacy custom GO_VERSION file

Go has native toolchain management now and solves a lot headaches including:

- Renovate having problem bumping changes in docker container, go.mod and GO_VERSION at once. 
- Requiring explicit workarounds in plugin makefile to ensure go version compatibility.
- Difficult experimenting with a new version.

Reference: https://go.dev/doc/toolchain

We can extract the same information in GO_VERSION using `go mod edit -json | jq -r .Toolchain`.

I am still debating on whether to retire GO_VERSION altogether or just add a script for renovate to hardwire it to the same value in go.mod

eternal-flame-AD:
Fixes #779 

Now go version updates gets reported as one PR that update both GO_VERSION and go.mod file pulled from gotify/build and not official go release, so that further builds have consistent toolchain version regardless of whether `go` is used directly or through docker.

I tested the dependency extraction and grouping works locally when manually downgrading the version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,11 @@ LICENSE_DIR=./licenses/
 BUILD_DIR=./build
 DOCKER_DIR=./docker/
 SHELL := /bin/bash
-GO_VERSION=`cat GO_VERSION`
+ifndef GOTOOLCHAIN
+	GO_VERSION=$(GOTOOLCHAIN)
+else
+	GO_VERSION=$(shell go mod edit -json | jq -r .Toolchain | sed -e 's/go//')
+endif
 DOCKER_BUILD_IMAGE=gotify/build
 DOCKER_WORKDIR=/proj
 DOCKER_RUN=docker run --rm -e LD_FLAGS="$$LD_FLAGS" -v "$$PWD/.:${DOCKER_WORKDIR}" -v "`go env GOPATH`/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
@@ -100,7 +104,7 @@ build-docker-multiarch: require-version
 		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -2) \
 		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -1) \
 		--build-arg RUN_TESTS=$(DOCKER_TEST_LEVEL) \
-		--build-arg GO_VERSION=$(shell cat GO_VERSION) \
+		--build-arg GO_VERSION=$(GO_VERSION) \
 		--build-arg LD_FLAGS="$$LD_FLAGS" \
 		--platform linux/amd64,linux/arm64,linux/386,linux/arm/v7,linux/riscv64 \
 		-f docker/Dockerfile .
--- a/renovate.json
+++ b/renovate.json
@@ -30,7 +30,25 @@
       "autoReplaceStringTemplate": "{{{newValue}}}",
       "datasourceTemplate": "docker",
       "versioningTemplate": "docker"
+    },
+    {
+      "customType": "regex",
+      "fileMatch": [
+        "^go.mod$"
+      ],
+      "depTypeTemplate": "language",
+      "matchStrings": [
+        "toolchain go(?<currentValue>[0-9.]+)\\n"
+      ],
+      "extractVersionTemplate": "^(?<version>.+)-linux-amd64$",
+      "depNameTemplate": "docker.io/gotify/build",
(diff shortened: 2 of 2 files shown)


eternal-flame-AD:
There are not too many important updates in 1.24.1 so I am not in a rush to bump it but we can test the workflow immediately after merging.

Makefile
```diff
@@ -100,7 +104,7 @@ build-docker-multiarch: require-version
 		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -2) \
 		-t ghcr.io/gotify/server-riscv64:$(shell echo $(VERSION) | cut -d '.' -f -1) \
 		--build-arg RUN_TESTS=$(DOCKER_TEST_LEVEL) \
-		--build-arg GO_VERSION=$(shell cat GO_VERSION) \
```
jmattheis, line 103: So the GO_VERSION is basically now only used by the plugin builds?

Makefile
```diff
@@ -2,7 +2,11 @@ LICENSE_DIR=./licenses/
 BUILD_DIR=./build
```
jmattheis, line 2: I've read the toolchain docs, but I don't fully understand why we want a newer toolchain version while having an older go version in the go.mod file.

As I understand the go version in the go.mod file defines the min go version for using this project as a library and the min version for the toolchain. With toolchain we say, this is our preferred toolchain version. When someone has an older version, our preferred toolchain is downloaded, even if the existing (older) version, would satisfy the go.mod go version requirement.

We don't want to have an up-to-date go.mod go version requirement, as this project is buildable with older go versions and maybe some distros don't have an up-to-date go version. But I don't understand why we want to have an up-to-date toolchain version. Why don't we have the go and toolchain version in sync with an older version, users could have an more up-to-date version installed which would be used, and users with an older version may use this to build the project.

renovate.json
```diff
@@ -30,8 +30,26 @@
...
       "versioningTemplate": "docker"
+    },
+    {
+      "customType": "regex",
+      "fileMatch": [
+        "^go.mod$"
+      ],
+      "depTypeTemplate": "language",
+      "matchStrings": [
+        "toolchain go(?<currentValue>[0-9.]+)\\n"
```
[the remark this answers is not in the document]
eternal-flame-AD: I'm not sure, but it is being `go mod tidy`'ed at the end anyways so I didn't specifically research this, if you know it works I will get rid of it

Makefile
eternal-flame-AD: I thought about this and my understanding of the need of the toolchain directive is when you just install go on conservative distributions like Debian, etc. it just works out of the box.

> We don't want to have an up-to-date go.mod go version requirement,

I agree and I don't think this PR changes this.

>  users with an older version may use this to build the project.

I do agree with you it is kind of restricting options but I don't think it is a big issue here, there are nothing in this package intended for import and the worst case scenario is someone who may be able to build without a toolchain download now has to unless the manually set `GOTOOLCHAIN=local`.

> But I don't understand why we want to have an up-to-date toolchain version. 

With the problems above, on the bright side I think with this directive (roughly in the order I consider motivational to make this change):

- In the future it may eliminate the need for a separate build container for each version entirely, the container can stay the same we just mount the go toolchain cache there.
- CI can be ran more consistently with a real user's build environment. You can be reasonably confident a random `go build` will use a equal or newer go toolchain that we released the code with.
- some plugin compatibility issues can work more out of the box without hacks, we no longer have to download the GO_VERSION file from GitHub then use it to pull a specific `gotify/build` container, and the CI environment would be more in line with a real user install.

renovate.json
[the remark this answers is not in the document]
jmattheis: okay, don't think it hurts, but though it maybe fixes some other problem, can stay as is.

jmattheis (approved):

Makefile
```diff
@@ -2,7 +2,11 @@ LICENSE_DIR=./licenses/
...
 DOCKER_DIR=./docker/
 SHELL := /bin/bash
-GO_VERSION=`cat GO_VERSION`
+ifndef GOTOOLCHAIN
```
jmattheis, line 5: @eternal-flame-AD just to be sure: this should be `ifdef` instead of `ifndef` right? We only want to use the GOTOOLCHAIN variable if it's set otherwise fallback to the go.mod toolchain version. I'll change this on master.

eternal-flame-AD: Oops, slipped through, yes it does seem like a typo

PR #794, opened 2025-05-02
chore: add raycast link

bxb100:


eternal-flame-AD:
Hi, thanks for the PR but unfortunately we cannot accept this in this form.

if you have a feature extension to gotify you may be able to open a PR in gotify/contrib with better clarifications on what feature it adds, but in general if it only launches/manages gotify it doesn't belong to the official repositories.

bxb100:
Hi, I just thought that the Raycast Gotify extension could also be considered part of the Gotify ecosystem, and I believe the README is an appropriate place to showcase it.
Anyway, I respect your decision to close it. Thanks!

PR #802, opened 2025-05-28
🩹 — Make sure apps are always sent to client in the same order, no matter what DB is used.

ldidry:
Related to #358


jmattheis (approved):
Thanks!

PR #804, opened 2025-06-09
Add "Sort Order" field to Application model and order by it

jberlyn:
This is my attempt at resolving:
https://github.com/gotify/server/issues/358

This adds a new field to Application called "Sort Order". When fetching the list of Applications by user, it will order by this Sort Order. By default, Applications will have a Sort Order of 0.

This is my first time contributing to a Go codebase, I am not familiar with the database tooling so I am unsure if this will automatically migrate the Application table to include the "Sort Order" column.

Happy to take feedback on this feature implementation.

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -48,6 +48,10 @@ type ApplicationParams struct {
 	//
 	// example: 5
 	DefaultPriority int `form:"defaultPriority" query:"defaultPriority" json:"defaultPriority"`
+	// The order in which this application should appear in the UI. Defaults to 0.
+	//
+	// example: 7
+	SortOrder int `form:"sortOrder" query:"sortOrder" json:"sortOrder"`
 }
 
 // CreateApplication creates an application and returns the access token.
@@ -90,6 +94,7 @@ func (a *ApplicationAPI) CreateApplication(ctx *gin.Context) {
 			Name:            applicationParams.Name,
 			Description:     applicationParams.Description,
 			DefaultPriority: applicationParams.DefaultPriority,
+			SortOrder:		 applicationParams.SortOrder,
 			Token:           auth.GenerateNotExistingToken(generateApplicationToken, a.applicationExists),
 			UserID:          auth.GetUserID(ctx),
 			Internal:        false,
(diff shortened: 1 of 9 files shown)


eternal-flame-AD:
Thanks! It looks generally good. I have no problem going with this design, might be a couple nitpicks for a deep look before merge but otherwise looking good. However there might be a more user friendly UX.

Do you know about sortable drag and drop list? The one where you have "Apple" on one line, "Banana" on the second, and you can swap orders between them? I think it would be more intuitive than inputting a number.

you can implement this by labeling the "sort order" of Apps from 0-N, and then when the user drag and drop you upload a new sequence, or swap the index of two apps. I would prefer the full update approach since we are displaying Apps on one page anyways so there won't be too much.

Let me know if you are up to it!

jmattheis:
Another option would be to use decimals when reordering, this way only one app has to be updated in the database. E.g. You have this order (the number is the sort order index)
```
1. Apple
2. Banana
3. Orange
```
When now reordering Orange between Apple and Banana it would get the sort order 1.5.

But the full update proposal from eternal-flame-AD is fine too.

jberlyn:
I am not sure when I will get around to making any improvements to this pull request, but happy for others to contribute if they would like to round out this feature.

xavier-GitHub76:
I test the feature, I validate.
image
image

:+1: 

The feature could be added to master
and drag n drop could be added after

I'm OK to try add the dnd

eternal-flame-AD:
@jberlyn Do you still plan on finishing this PR or are you open to me finishing it when I got some spare time to spend since it seems there is community need for this?

jberlyn:
> @jberlyn Do you still plan on finishing this PR or are you open to me finishing it when I got some spare time to spend since it seems there is community need for this?

More than happy for you to take a look at it.

xavier-GitHub76:
I will push a pull request today to add drag and drop on sortOrder

xavier-GitHub76:
The pull request is created https://github.com/gotify/server/pull/856
I merged the changes made by @jberlyn to resolve conflicts
This PR could be close

Issue #790, opened 2025-04-14
Add the possibility to collapse notifications


sanderkoenders:
**Is your feature request related to a problem? Please describe.**
I have been using Gotify for the past couple of weeks and it has been serving me very well. The only minor annoyance I have, which is unrelated to Gotify, is the notifications I receive from my Proxmox backups. It is a very lengthy message because the logs are attached.

**Describe the solution you'd like**
The solution I was thinking about is adding the possibility to collapse the notification when it is longer than `x` characters. Ideally there would be a "view more" button that allows to show the collapsed contents.

**Describe alternatives you've considered**
I have considered turning off Proxmox backup notifications altogether but unfortunately then I would lose the lion share of useful information.

jmattheis:
👍, seems good, we can probably collapse after 20 or so lines.

jeremygooch:
Assuming this is in the Messages area of the UI, this one should be simple to implement using existing Material components. Please let me know if that is the case and I can submit a PR for it asap.

jmattheis:
Yeah, it should be there. Additionally the height of the Message component has to be updated in the parent component when it's collapsed / uncollapsed. so that the virtual list works correctly.

PR #810, opened 2025-06-29
Adding functionality to collapse long running messages

jeremygooch:
In reference to: https://github.com/gotify/server/issues/790

This PR adds a Collapse component to long running messages. The height of the message preview is configurable in the Message component, as is the animation timeout. The animation timeout is also used to determine when to recalculate the height of the Message so it can be sent to the parent component via the `height` function. 

I didn't see any test files for this component that needed to be updated, but lmk if there's anything that needs to be addressed there, or if anything else on the PR needs attention.

Thanks also to @jmattheis for the pointer to the `height` function.

### DIFF
--- a/ui/src/message/Message.tsx
+++ b/ui/src/message/Message.tsx
@@ -1,14 +1,20 @@
+import {Button, Collapse} from '@material-ui/core';
 import IconButton from '@material-ui/core/IconButton';
 import {createStyles, Theme, withStyles, WithStyles} from '@material-ui/core/styles';
+import {ClassNameMap} from '@material-ui/core/styles/withStyles';
 import Typography from '@material-ui/core/Typography';
+import {ExpandLess, ExpandMore} from '@material-ui/icons';
 import Delete from '@material-ui/icons/Delete';
-import React from 'react';
+import React, {RefObject} from 'react';
 import TimeAgo from 'react-timeago';
 import Container from '../common/Container';
-import * as config from '../config';
 import {Markdown} from '../common/Markdown';
-import {RenderMode, contentType} from './extras';
+import * as config from '../config';
 import {IMessageExtras} from '../types';
+import {contentType, RenderMode} from './extras';
+
+const PREVIEW_LENGTH = 500;
+const ANIMATION_TIMEOUT_MS = 500;
 
 const styles = (theme: Theme) =>
     createStyles({
@@ -52,7 +58,11 @@ const styles = (theme: Theme) =>
             whiteSpace: 'pre-wrap',
         },
         content: {
+            maxHeight: PREVIEW_LENGTH,
             wordBreak: 'break-all',
+            '&.expanded, &.collapsed': {
+                maxHeight: 'none',
+            },
             '& p': {
                 margin: 0,
             },
@@ -79,6 +89,11 @@ interface IProps {
     height: (height: number) => void;
 }
 
+interface IState {
+    expanded: boolean;
+    isOverflowing: boolean;
+}
+
 const priorityColor = (priority: number) => {
     if (priority >= 4 && priority <= 7) {
         return 'rgba(230, 126, 34, 0.7)';
@@ -91,9 +106,45 @@ const priorityColor = (priority: number) => {
 
 class Message extends React.PureComponent<IProps & WithStyles<typeof styles>> {
     private node: HTMLDivElement | null = null;
+    public state: IState = {expanded: false, isOverflowing: false};
+    previewRef: RefObject<HTMLDivElement>;
 
-    public componentDidMount = () =>
+    constructor(props: IProps) {
+        super(
+            props as IProps & {
+                classes: ClassNameMap<
+                    | 'header'
+                    | 'image'
+                    | 'content'
+                    | 'headerTitle'
+                    | 'trash'
+                    | 'wrapperPadding'
+                    | 'messageContentWrapper'
+                    | 'date'
+                    | 'imageWrapper'
+                    | 'plainContent'
+                >;
+            }
+        );
+        this.previewRef = React.createRef();
+    }
+
+    public componentDidMount = () => {
+        if (this.previewRef.current) {
+            this.setState({
+                ...this.state,
+                isOverflowing:
+                    this.previewRef.current.scrollHeight > this.previewRef.current.clientHeight,
+            });
+        }
+        return this.props.height(this.node ? this.node.getBoundingClientRect().height : 0);
+    };
+
+    public togglePreviewHeight = async () => {
+        this.setState({...this.state, expanded: !this.state.expanded});
+        await new Promise((resolve) => setTimeout(resolve, ANIMATION_TIMEOUT_MS + 1));
         this.props.height(this.node ? this.node.getBoundingClientRect().height : 0);
+    };
 
     private renderContent = () => {
         const content = this.props.content;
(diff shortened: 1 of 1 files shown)


ui/src/message/Message.tsx
```diff
@@ -91,9 +106,45 @@ const priorityColor = (priority: number) => {
...
+                    | 'header'
+                    | 'image'
+                    | 'content'
+                    | 'headerTitle'
+                    | 'trash'
+                    | 'wrapperPadding'
+                    | 'messageContentWrapper'
+                    | 'date'
+                    | 'imageWrapper'
+                    | 'plainContent'
```
jeremygooch: Note, I was unable to make the TS compiler happy without using this long union type, and I really didn't want to add a `@ts-ignore` to the code. If there's a better/more preferred way to handle this lmk and I'll update asap.

jmattheis:
```
    constructor(props: IProps & WithStyles<typeof styles>) {
        super(props);
```

ui/src/message/Message.tsx
```diff
@@ -91,9 +106,45 @@ const priorityColor = (priority: number) => {
...
+                    this.previewRef.current.scrollHeight > this.previewRef.current.clientHeight,
+            });
+        }
+        return this.props.height(this.node ? this.node.getBoundingClientRect().height : 0);
+    };
 
-    public componentDidMount = () =>
+    public togglePreviewHeight = async () => {
+        this.setState({...this.state, expanded: !this.state.expanded});
+        await new Promise((resolve) => setTimeout(resolve, ANIMATION_TIMEOUT_MS + 1));
```
jmattheis: This seems hacky, and possibly not correct as the setState is also async so the animation could be a couple milliseconds in the future. The Collapse provides onExit and onEnter callbacks which are called when the element was opened / closed. https://stackoverflow.com/questions/71799622/material-ui-5-how-to-know-when-collapse-transition-is-finished

Edit: Solution possibly outdated by later comment, but there should be no setTimeout usage like this.

ui/src/message/Message.tsx
```diff
@@ -52,7 +58,11 @@ const styles = (theme: Theme) =>
             whiteSpace: 'pre-wrap',
         },
         content: {
+            maxHeight: PREVIEW_LENGTH,
             wordBreak: 'break-all',
+            '&.expanded, &.collapsed': {
+                maxHeight: 'none',
+            },
```
jmattheis: Why have two classes (expanded & collapsed when the do the same?

ui/src/message/Message.tsx
```diff
@@ -141,9 +192,36 @@ class Message extends React.PureComponent<IProps & WithStyles<typeof styles>> {
                                 <Delete />
                             </IconButton>
                         </div>
-                        <Typography component="div" className={`${classes.content} content`}>
-                            {this.renderContent()}
-                        </Typography>
+
+                        {!this.state.isOverflowing ? (
```
jmattheis: This seems overly complicated to work with the collapse element while also doing the collapsing with maxHeight in the css. It should be possible to do this without the Collapse element. Styles like this
```
            maxHeight: PREVIEW_LENGTH,
            wordBreak: 'break-all',
            overflowY: 'hidden',
            '&.expanded': {
                maxHeight: 'none',
            },
```
and message content like this
```
                        <Typography
                            component="div"
                            ref={this.previewRef}
                            className={`${classes.content} content ${!this.state.isOverflowing ? 'loading' : this.state.expanded ? 'expanded' : 'collapsed'}`}>
                            {this.renderContent()}
                        </Typography>
```

ui/src/message/Message.tsx
```diff
@@ -141,9 +192,36 @@ class Message extends React.PureComponent<IProps & WithStyles<typeof styles>> {
...
+                            </Collapse>
+                        )}
+
+                        {this.state.isOverflowing && (
+                            <Button
+                                onClick={() => this.togglePreviewHeight()}
+                                color="primary"
+                                size="small"
+                                startIcon={this.state.expanded ? <ExpandLess /> : <ExpandMore />}>
+                                {this.state.expanded ? 'Read Less' : 'Read More'}
```
jmattheis: This small button is difficult to press should be a full width botton like this.
image

ui/src/message/Message.tsx
jeremygooch: Ya, wasn't happy with the timeout as timing in JS is never as precise as it should be. Thanks for the pointers on the onExit and onEnter callbacks. That'll be helpful for future reference.

ui/src/message/Message.tsx
jeremygooch: Will tidy this up.

ui/src/message/Message.tsx
jeremygooch: Agreed. The one element was needed for measuring whether the content was overflowing so the Collapse element could be used, otherwise Collapse was throwing off the layout for non long-running content. I definitely like this simpler approach as well as it was similar to my initial approach. I shifted gears when I saw the animations on Collapse and thought it'd be a nice touch for the UI. Happy to drop it though.

ui/src/message/Message.tsx
jeremygooch: Good call, updated. What I went with looks a bit different than your screenshot here in that I kept with the standard template colors on the page, however, happy to change it to something else if that'd be better.

ui/src/message/Message.tsx
```diff
@@ -91,12 +105,36 @@ const priorityColor = (priority: number) => {
...
+                isOverflowing:
+                    this.previewRef.current.scrollHeight > this.previewRef.current.clientHeight,
+            });
+        }
+        return this.updateHeightInParent();
+    };
+
+    public togglePreviewHeight = () => {
+        this.setState({...this.state, expanded: !this.state.expanded});
+        this.updateHeightInParent();
```
jmattheis: This doesn't work, as setState is asynchronous, so the size change in the DOM has very likely not happened yet. I've changed this to
```
        this.setState(
            (state) => ({expanded: !state.expanded}),
            () => this.updateHeightInParent(),
        );
```
https://legacy.reactjs.org/docs/react-component.html#setstate
the callback should be executed after the setState was applied.

ui/src/message/Message.tsx
```diff
@@ -91,12 +105,36 @@ const priorityColor = (priority: number) => {
...
+    constructor(props: IProps & WithStyles<typeof styles>) {
+        super(props);
+        this.previewRef = React.createRef();
+    }
 
-    public componentDidMount = () =>
+    public componentDidMount = () => {
+        if (this.previewRef.current) {
+            this.setState({
+                ...this.state,
```
jmattheis: This can be omitted, as react already does a shallow merge. (already changed)

ui/src/message/Message.tsx
```diff
@@ -141,10 +180,28 @@ class Message extends React.PureComponent<IProps & WithStyles<typeof styles>> {
                                 <Delete />
                             </IconButton>
                         </div>
-                        <Typography component="div" className={`${classes.content} content`}>
+
+                        <Typography
+                            component="div"
+                            ref={this.previewRef}
+                            className={`${classes.content} content ${!this.state.isOverflowing ? 'loading' : this.state.expanded ? 'expanded' : 'collapsed'}`}>
```
jmattheis: I've simplified this to `this.state.isOverflowing && this.state.expanded ? 'expanded' : ''`

jmattheis (approved):
Awesome, looks pretty good. Thanks! I've made some minor cleanup and fixed the height updating. See sub comments.

PR #812, opened 2025-07-09
fix: skip dotfiles and directories in plugin loading (#811)

### LINKED ISSUE #811: Gotify not running due to Synology file generation (plugin folder)
On Synology systems it sometimes creates a directory called `@eaDir` when this happens in the `plugins` folder of Gotify it causes it to error out and not start up at all.

A list to ignore known patterns of auto generated files or directories for some systems.

Deleting the directory allows Gotify to run again, but it would be annoying if those directories were generated again, causing Gotify not to run.

```
**/.DS_Store
**/Thumbs.db
**/@eaDir

### PR DESCRIPTION
**Pull request by eternal-flame-AD**
Fixes #811 by ignoring dotfiles and directories in plugin discovery.

### DIFF
--- a/plugin/manager_test.go
+++ b/plugin/manager_test.go
@@ -147,6 +147,20 @@ func (s *ManagerSuite) TestInitializePlugin_noOpIfEmpty() {
 	assert.Nil(s.T(), s.manager.loadPlugins(""))
 }
 
+func (s *ManagerSuite) TestInitializePlugin_noOpIfDotFile() {
+	tmpDir := test.NewTmpDir("gotify_testinitializeplugin_dotfile")
+	defer tmpDir.Clean()
+	os.Mkdir(tmpDir.Path(".test"), 0o755)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugin/manager_test.go [hunk 1]
```diff
@@ -147,6 +147,20 @@ func (s *ManagerSuite) TestInitializePlugin_noOpIfEmpty() {
 	assert.Nil(s.T(), s.manager.loadPlugins(""))
 }
 
+func (s *ManagerSuite) TestInitializePlugin_noOpIfDotFile() {
+	tmpDir := test.NewTmpDir("gotify_testinitializeplugin_dotfile")
+	defer tmpDir.Clean()
+	os.Mkdir(tmpDir.Path(".test"), 0o755)
```
jmattheis: The test name implies it's a file, but it's a directory here? Should this be a create file?

jmattheis (approved):
:+1:

PR #815, opened 2025-07-23
Consistency and Accessibility Improvements of the user-related UIs

martinfrancois:
Thanks for making gotify, it's truly awesome!

Just some small things I noticed that could be improved, let me know in case you wouldn't want one of the changes, I intentionally split them up into multiple commits so it would be easy to remove any changes you wouldn't want.

I tried adding justifications to the commit's bodies, if anything is unclear don't hesitate to ask back.

### DIFF
--- a/ui/src/user/Login.tsx
+++ b/ui/src/user/Login.tsx
@@ -28,17 +28,21 @@ class Login extends Component<Stores<'currentUser'>> {
                         <form onSubmit={this.preventDefault} id="login-form">
                             <TextField
                                 autoFocus
+                                id="username"
                                 className="name"
                                 label="Username"
+                                name="username"
                                 margin="dense"
                                 autoComplete="username"
                                 value={username}
                                 onChange={(e) => (this.username = e.target.value)}
                             />
                             <TextField
+                                id="password"
                                 type="password"
                                 className="password"
                                 label="Password"
+                                name="password"
                                 margin="normal"
                                 autoComplete="current-password"
                                 value={password}
--- a/ui/src/user/AddEditUserDialog.tsx
+++ b/ui/src/user/AddEditUserDialog.tsx
@@ -56,6 +56,8 @@ export default class AddEditDialog extends Component<IProps, IState> {
                         label="Name *"
                         type="email"
                         value={name}
+                        name="username"
+                        id="username"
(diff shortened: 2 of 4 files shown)


ui/src/user/Login.tsx
```diff
@@ -28,17 +28,21 @@ class Login extends Component<Stores<'currentUser'>> {
                         <form onSubmit={this.preventDefault} id="login-form">
                             <TextField
                                 autoFocus
+                                id="username"
                                 className="name"
                                 label="Username"
+                                name="username"
```
jmattheis, line 34: Does name affects accessibility? For `id` it's documented in the mui link, but for name there is no additional notice for accessibility.

Otherwise, the PR looks good. Awesomely separated commits.

martinfrancois: Yes it does, thanks for asking! According to the [WCAG, the autocomplete feature](https://www.w3.org/WAI/standards-guidelines/act/rules/73f2c2/) is an important accessibility feature, especially for users with motor or dexterity impairments (reducing typing effort) and cognitive disabilities (lowering cognitive load by prefilling fields). 

However, the `autocomplete` attribute is **only** used by browsers **if** the [`id` **and**/or the `name` (depending on the browser)](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete?utm_source=chatgpt.com#description) attributes are set. So it's safer to have both set to make sure all browsers autofill those fields consistently. That's actually what inspired me to submit this PR, as the autocomplete wasn't working very consistently for me.

Also thanks for the praise on the PR and the compliment on the commits! I really appreciate your feedback and am glad the separation made things easier :)

martinfrancois:
> lgtm and great commit history management!
> 
> I will let the owner take a second look before merging. He knows frontend best.

Thanks for the quick review and the kind words about my commit history! I really appreciate you passing this along, I'm sure that helped speed things up :)

ui/src/user/Login.tsx
jmattheis: Thanks for the explanation!

jmattheis (approved):

ui/src/user/Login.tsx
martinfrancois: You're welcome!

PR #817, opened 2025-08-04
Add styling for code blocks in notifications

BerserkeR-Git:
styling for code blocks

### DIFF
--- a/ui/src/message/Message.tsx
+++ b/ui/src/message/Message.tsx
@@ -70,6 +70,10 @@ const useStyles = makeStyles()((theme: Theme) => ({
         },
         '& pre': {
             overflow: 'auto',
+            borderRadius: '0.25em',
+            backgroundColor:
+                theme.palette.mode === 'dark' ? 'rgba(255,255,255,0.05)' : 'rgba(0,0,0,0.05)',
+            padding: theme.spacing(1),
         },
         '& img': {
             maxWidth: '100%',


ui/src/message/Message.tsx
```diff
@@ -74,6 +74,20 @@ const styles = (theme: Theme) =>
             '& img': {
                 maxWidth: '100%',
             },
+            '& code': {
```
jmattheis: I think this should be enough.
```
            '& code': {
                display: 'inline-flex',
                backgroundColor: 'rgba(0, 0, 0, 0.25)',
                padding: theme.spacing(1),
                borderRadius: '0.25em',
            },
```
The `color` breaks when switching to light-theme. The padding should use the standard spacing. fontsize should be set by the parent. wordBreak/whiteSpace shouldn't be set as the code thingy should have a horizontal scrollbar. Boxshadow seems to much.

I'll change this myself, after #818 is done.

BerserkeR-Git: Thanks, I'm not that experienced with coding but I thought I'd make this suggestion.
Have a good day.

**Edit:** The only thing I would like to say is check if there is enough padding so the text doesn't almost jump out of the box, and also not too much where it pushes other notifications down a lot.

And also, a little faint shadow edge at the top and bottom can't hurt, so the code block looks properly embedded something like [this](https://i.imgur.com/VmT07VW.png)

jmattheis: The GitHub code styling doesn't have box shadow too. I'd rather not have it.

PR #818, opened 2025-08-04
Upgrade UI

jmattheis:
* Upgrade all UI dependencies to the latest version
* Migrate from react-scripts to vite
* Migrate from `@material-ui` to `@mui`
* Migrate from react-infinite to react-virtuoso (reused from #696)
* Migrate to react functional components (reused from #696)
* Migrate from custom impl to notistack
* Migrate away from "legacy decorators"
* Migrate from react-codemirror2 to `@uiw/codemirror`
* Remove NODE_OPTIONS workaround

Closes #696


Makefile
```diff
@@ -12,7 +12,6 @@ DOCKER_WORKDIR=/proj
 DOCKER_RUN=docker run --rm -e LD_FLAGS="$$LD_FLAGS" -v "$$PWD/.:${DOCKER_WORKDIR}" -v "`go env GOPATH`/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
```
jmattheis, line 12: @eternal-flame-AD Do you have time to give this PR a small manual test? It should work as the e2e tests are successful, but there may be something I overlooked.

The UI coloring changed a little because the defaults of the material-ui lib changed.

eternal-flame-AD: Sure, will do today or tomorrow!

ui/vite.config.ts
```diff
@@ -0,0 +1,36 @@
+import {defineConfig} from 'vite';
+import react from '@vitejs/plugin-react';
+
+const GOTIFY_PORT = process.env.GOTIFY_PORT ?? '80';
```
eternal-flame-AD: It should read from GOTIFY_SERVER_PORT or both because that is the key the server uses.

eternal-flame-AD (commented):
Maybe we can try some service worker stuff so latest messages always displays regardless of Internet connection, similar to the Android App.

eternal-flame-AD:
Here's a demo script that creates a variety of messages under realistic scenarios (and they are all free assets so we can use it for branding).

It would also be nice if we can get clickURL to work on web frontend.

https://gist.github.com/eternal-flame-AD/de79b3431281c802548d9c0b4dbbdc47

screenshot

ui/vite.config.ts
jmattheis: Yeah good catch!

eternal-flame-AD (dismissed):
lgtm, I don't see any regressions so good to merge. It's a long wanted feature so I am good with just get it out of the door first and then we can see maybe new features (the service worker and more interactive `extras` support)

Maybe update the screenshot because the UI now looks different.

Makefile
```diff
@@ -7,12 +7,11 @@ ifdef GOTOOLCHAIN
 else
```
jmattheis, line 7: Reconnecting doesn't correctly refresh the currently visible data.

[reconnect.webm](https://github.com/user-attachments/assets/73be1982-f405-4a92-89c5-92988c845d9c)

eternal-flame-AD: Not sure how you made this happen, I blocked `/message` and clicked refresh and can get stuck in this state.

Probably it's MessagesStore#loadMore needs a `.finally(() => (this.loading = false))`, I think it is an old bug bc IIRC I accidentally got it stuck refreshing before too.

jmattheis: This is a separate problem, I'll fix this too. The one I can reproduce happens after restarting gotify/server. The cause of this bug is that the state is cleared after a reconnect happens (there could be new messages while the websocket was disconnected, so we do a full refresh). This happens here https://github.com/gotify/server/blob/43574a075cd5315f6eae2e146f6d951dbfa12ffa/ui/src/reactions.ts#L37-L45 the pages like Messages currently do the initial fetch of the data via a useEffect. https://github.com/gotify/server/blob/43574a075cd5315f6eae2e146f6d951dbfa12ffa/ui/src/message/Messages.tsx#L30-L34

This useEffect isn't reexecuted after the state is cleared by the reaction. On the current master version this is done in a different way that is reexecuted.

PR #819, opened 2025-08-04
fix message layout

eternal-flame-AD:
Not sure if it shows up like this on your end as well. Tried fixing it, can't find a way to make App Logos show up on really small screens neatly so I think we should just hide it: it saves metered traffic as well.

Before:

image

image

After:

image
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ ifdef GOTOOLCHAIN
 else
 	GO_VERSION=$(shell go mod edit -json | jq -r .Toolchain | sed -e 's/go//')
 endif
-DOCKER_BUILD_IMAGE=gotify/build
+DOCKER_BUILD_IMAGE=docker.io/gotify/build
 DOCKER_WORKDIR=/proj
 DOCKER_RUN=docker run --rm -e LD_FLAGS="$$LD_FLAGS" -v "$$PWD/.:${DOCKER_WORKDIR}" -v "`go env GOPATH`/pkg/mod/.:/go/pkg/mod:ro" -w ${DOCKER_WORKDIR}
 DOCKER_GO_BUILD=go build -mod=readonly -a -installsuffix cgo -ldflags "$$LD_FLAGS"
--- a/ui/src/message/Message.tsx
+++ b/ui/src/message/Message.tsx
@@ -1,4 +1,4 @@
-import {Button, Theme} from '@mui/material';
+import {Button, Theme, useMediaQuery, useTheme} from '@mui/material';
 import IconButton from '@mui/material/IconButton';
 import {makeStyles} from 'tss-react/mui';
 import Typography from '@mui/material/Typography';
@@ -33,6 +33,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     },
     messageContentWrapper: {
         minWidth: 200,
+        width: '100%',
     },
     image: {
         marginRight: 15,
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ ifdef GOTOOLCHAIN
 else
```
jmattheis, line 7: I think we should display the app images even on small screens (I can already smell the issues that the app icons aren't visible :D)

Maybe it could be formatted like this?

image

Maybe this could even be the default formatting on all sizes.

ui/src/message/Message.tsx
```diff
@@ -11,6 +11,7 @@ import {Markdown} from '../common/Markdown';
 import * as config from '../config';
```
jmattheis, line 11: The changes currently break `<code>` blocks that are wider than the message box. And also markdown formatted messages

image

```
{
  "title": "Title",
  "message": "hello\n```\nmycodeaoeouanetohunaotheuntaoheuntoaheunt                                                                 oheuntoheunthoeaunthaoeunth\n```\n",
  "priority": 5,
  "extras": { "client::display": { "contentType": "text/markdown" } }
}
```

ui/src/message/Message.tsx
```diff
@@ -39,14 +40,34 @@ const useStyles = makeStyles()((theme: Theme) => ({
             width: 32,
```
jmattheis: Adding `width: '100%'` to messageContentWrapper seems to fix the issue where the time / delete button isn't on the right. I only tested the styling with big messages, thanks for finding this (:.

Makefile
jmattheis: Could we do this change in a separate PR independent of upgrade-ui?

eternal-flame-AD: I suppose, it is complicated and I think discussion potentially change is needed.

I personally don't use this icon feature very often: reason simply being it kind of becomes an ad-hoc graphics design problem where the icon needs to be readable on any screen and the file has to be appropriately compressed so it loads fast. More often than not I find just text description being better.

For example I name my servers after mountain names: nokogiri, gorgonio, ontake, etc. It is simply much more recognizable to write the name in text than find an appropriate image for these mountains that displays well under any size.

I think that is also why most "uptime monitoring" apps don't have app icon feature, users have difficulty sourcing images that display well in a list or card grid like fashion and it's more clean and neat without it.

But I feel the text app name absolutely should be there for both accessibility and user choice reasons.

ui/src/message/Message.tsx
eternal-flame-AD: Huhh, I did test with markdown formatted images. Let me double check .. 

Should we just have a sample database/script that posts a list of various messages so it is easy to check nothing breaks?

Makefile
jmattheis: I personally agree with the usefulness of the icons, but I think there are a significant  amount of users that would. Maybe we could hide the image if it's the default one, and only display it when the user has uploaded a custom one.

> But I feel the text app name absolutely should be there for both accessibility and user choice reasons.

Agreed, but we should probably omit it, if the title is equal to the app name.

ui/src/message/Message.tsx
jmattheis: Yeah, we definitely should have this.

eternal-flame-AD:
Some minor changes:

- Using the narrow time format feels better to use the horizontal space better in kind-of md devices.
- two small tweaks for development environment compatibility and consistency

PR #824, opened 2025-08-08
fix: less whitespace in message & add app name

jmattheis:
Suggestion from https://github.com/gotify/server/pull/819#discussion_r2255151054

@eternal-flame-AD what do you think?

desktop:

image

mobile:

image

### DIFF
--- a/ui/src/message/Message.tsx
+++ b/ui/src/message/Message.tsx
@@ -18,8 +18,11 @@ const PREVIEW_LENGTH = 500;
 const useStyles = makeStyles()((theme: Theme) => ({
     header: {
         display: 'flex',
-        flexWrap: 'wrap',
-        marginBottom: 0,
+        width: '100%',
+        alignItems: 'start',
+        alignContent: 'center',
+        paddingBottom: 5,
+        wordBreak: 'break-all',
     },
     headerTitle: {
         flex: 1,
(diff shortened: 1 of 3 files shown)


ui/src/message/Message.tsx
```diff
@@ -150,48 +153,52 @@ const Message = ({
                     borderLeftWidth: 6,
                     borderLeftStyle: 'solid',
                 }}>
-                <div style={{display: 'flex', width: '100%'}}>
+                <div className={classes.header}>
```
jmattheis: These changes break display of the message on smaller screens.

eternal-flame-AD:
Looks great to me, I personally prefer a more compact and information dense look like this.

(I haven't tried it live yet)

ui/src/message/Message.tsx
```diff
@@ -1,4 +1,4 @@
-import {Button, Theme, useMediaQuery, useTheme} from '@mui/material';
+import {Box, Button, Theme, useMediaQuery, useTheme} from '@mui/material';
```
jmattheis: I've updated the description with new screenshots. The mobile style now has a different header layout. @eternal-flame-AD do you have a better idea for the application image in the mobile layout, it looks a little out of place but at least it's more compact?

eternal-flame-AD (dismissed):
lgtm!

PR #828, opened 2025-08-17
Adds Redis pub/sub for multi-pod WebSocket support

MalukiMuthusi:
Implements a message broker system with Redis backend to enable WebSocket notifications across multiple Gotify server instances. Introduces a broker interface with Redis and no-op implementations, allowing deployments to scale horizontally while maintaining real-time message delivery.

Falls back gracefully to local-only mode when Redis is disabled or unavailable. Includes configuration options for Redis connection and channel prefix customization.

### DIFF
--- a/api/stream/stream.go
+++ b/api/stream/stream.go
@@ -1,6 +1,7 @@
 package stream
 
 import (
+	"log"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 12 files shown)


eternal-flame-AD:
Hi, is this a performance PR or some specific infrastructure that requires this horizontal/message bus setup?

If it's performance it would be nice if you have benchmark or load testing numbers (concurrent connections, etc).

If it's a specific infrastructure (maybe k8s?) you are targeting please open an concurrent issue/PR at gotify/website for how to configure the system to take advantage of this feature. (Example configurations, security considerations, etc)

MalukiMuthusi:
It is targeting an infrastructure like k8s.

eternal-flame-AD:
Okay thanks for the clarification, can you open PR for a sample k8s configuration with this feature on gotify/website, thanks.

MalukiMuthusi:
Thank you, I will open it.

PR #838, opened 2025-09-13
Fix ci

jmattheis:


Makefile
```diff
@@ -2,11 +2,7 @@ LICENSE_DIR=./licenses/
 BUILD_DIR=./build
 DOCKER_DIR=./docker/
 SHELL := /bin/bash
-ifdef GOTOOLCHAIN
-	GO_VERSION=$(GOTOOLCHAIN)
-else
-	GO_VERSION=$(shell go mod edit -json | jq -r .Toolchain | sed -e 's/go//')
-endif
+GO_VERSION=$(shell go mod edit -json | jq -r .Toolchain | sed -e 's/go//')
```
jmattheis, line 5: FYI @eternal-flame-AD I've removed the GOTOOLCHAIN check here, because in a new version of actions/setup-go the variable is set to local which breaks this logic here. I think if someone from the outside wants to change the use go version, they can directly set GO_VERSION.

Issue #836, opened 2025-09-12
GOTIFY_SERVER_SSL_LETSENCRYPT_HOSTS: environment variable not working


Helium-7:
**Have you read the documentation?**
- [ × ] Yes, but it does not include related information regarding my question.
- [ ] Yes, but the steps described in the documentation do not work on my machine.
- [ ] Yes, but I am having difficulty understanding it and want clarification.

**You are setting up gotify in**
- [ × ] Docker
- [ ] Linux native platform
- [ ] Windows native platform

**Describe your problem**
Trying to set up SSL for gotify, running in a container, set up with docker compose. 
Relevant compose.yaml file: 
```
services:
  gotify:
    image: gotify/server
    container_name: gotify
    restart: always
    ports:
      - "32768:80"
      - "42768:443"
    environment:
      GOTIFY_DEFAULTUSER_PASS: admin
      GOTIFY_SERVER_SSL_ENABLED: "true"
      GOTIFY_SERVER_SSL_LETSENCRYPT_HOSTS: gotify.url.xyz
      GOTIFY_SERVER_SSL_PORT: 443
      GOTIFY_SERVER_SSL_LETSENCRYPT_ENABLED: "true"
      GOTIFY_SERVER_SSL_LETSENCRYPT_ACCEPTTOS: "true"
      GOTIFY_SERVER_SSL_LETSENCRYPT_CACHE: /app/data/certs
      GOTIFY_SERVER_SSL_REDIRECTTOHTTPS: "true"
    volumes:
      - './gotify_data:/app/data'
```
I get this error when starting the container
```
gotify  | Starting Gotify version 2.6.3@2025-04-27-09:10:38
gotify  | panic: yaml: unmarshal errors:
gotify  |         line 1: cannot unmarshal !!str `gotify....` into []string
gotify  |
gotify  | goroutine 1 [running]:
gotify  | github.com/gotify/server/v2/config.Get()
gotify  |       /src/gotify/config/config.go:71 +0x1e6
gotify  | main.main()
gotify  |       /src/gotify/app.go:31 +0x18f
gotify exited with code 2
```

jmattheis:
You need to wrap the hosts into an array. E.g.
```
GOTIFY_SERVER_SSL_LETSENCRYPT_HOSTS: '[gotify.url.xyz]'
```

Helium-7:
> You need to wrap the hosts into an array. E.g.
> 
> ```
[...]

PR #843, opened 2025-09-16
feat: refine AutoCert logic

eternal-flame-AD:
It seems like the Autocert logic can be quite confusing and errors can be implicit ..

This tries to:

-  make errors more explicit and noisy
-  corrected a missing ALPN issue for the TLS challenge part, which should make success rate higher as long as one of TLS/HTTP challenge passes.
- allows the staging endpoint to be used for initial setup which should reduce friction or unexplained failures from ratelimiting

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -98,24 +98,25 @@ func getNetworkAndAddr(listenAddr string, port int) (string, string) {
 }
 
 type LoggingRoundTripper struct {
+	Name         string
 	RoundTripper http.RoundTripper
 }
 
 func (l *LoggingRoundTripper) RoundTrip(r *http.Request) (resp *http.Response, err error) {
 	resp, err = l.RoundTripper.RoundTrip(r)
 	if resp.StatusCode == 429 {
-		log.Printf("Let's Encrypt Client Rate Limited: Retry-After %s on %s %s\n", resp.Header.Get("Retry-After"), r.Method, r.URL.String())
+		log.Printf("%s Rate Limited: Retry-After %s on %s %s\n", l.Name, resp.Header.Get("Retry-After"), r.Method, r.URL.String())
 	} else if resp.StatusCode >= 400 {
-		log.Printf("Let's Encrypt Client Request Failed: Unexpected status code %d on %s %s\n", resp.StatusCode, r.Method, r.URL.String())
+		log.Printf("%s Request Failed: Unexpected status code %d on %s %s\n", l.Name, resp.StatusCode, r.Method, r.URL.String())
 	} else if err != nil {
-		log.Printf("Let's Encrypt Client Request Failed: %s on %s %s\n", err.Error(), r.Method, r.URL.String())
+		log.Printf("%s Request Failed: %s on %s %s\n", l.Name, err.Error(), r.Method, r.URL.String())
 	}
 	return
 }
 
 func applyLetsEncrypt(s *http.Server, conf *config.Configuration) {
 	httpClient := http.DefaultClient
-	httpClient.Transport = &LoggingRoundTripper{RoundTripper: http.DefaultTransport}
+	httpClient.Transport = &LoggingRoundTripper{Name: "Let's Encrypt", RoundTripper: http.DefaultTransport}
 	httpClient.Timeout = 60 * time.Second
 
 	acmeClient := &acme.Client{
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -32,7 +32,7 @@ func Run(router http.Handler, conf *config.Configuration) error {
 	if conf.Server.SSL.Enabled {
 		if conf.Server.SSL.LetsEncrypt.Enabled {
 			applyLetsEncrypt(s, conf)
-		} else if conf.Server.SSL.CertFile != "" && conf.Server.SSL.CertKey != "" {
+		} else if conf.Server.SSL.CertFile == "" || conf.Server.SSL.CertKey == "" {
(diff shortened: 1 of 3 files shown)


runner/runner.go
```diff
@@ -94,12 +97,43 @@ func getNetworkAndAddr(listenAddr string, port int) (string, string) {
...
+	} else if resp.StatusCode >= 400 {
+		log.Printf("%s Request Failed: Unexpected status code %d on %s %s\n", l.Name, resp.StatusCode, r.Method, r.URL.String())
+	} else if err != nil {
+		log.Printf("%s Request Failed: %s on %s %s\n", l.Name, err.Error(), r.Method, r.URL.String())
+	}
+	return
+}
+
 func applyLetsEncrypt(s *http.Server, conf *config.Configuration) {
+	httpClient := http.DefaultClient
```
jmattheis:
```suggestion
	httpClient := &http.Client{}
```
http.DefaultClient is a pointer so this would modify the default client.

runner/runner.go
```diff
@@ -94,12 +97,43 @@ func getNetworkAndAddr(listenAddr string, port int) (string, string) {
...
+	acmeClient := &acme.Client{
+		HTTPClient:   httpClient,
+		DirectoryURL: conf.Server.SSL.LetsEncrypt.DirectoryURL,
+	}
 	certManager := autocert.Manager{
-		Prompt:     func(tosURL string) bool { return conf.Server.SSL.LetsEncrypt.AcceptTOS },
+		Client: acmeClient,
+		Prompt: func(tosURL string) bool {
+			if !conf.Server.SSL.LetsEncrypt.AcceptTOS {
+				log.Fatalf("Let's Encrypt TOS must be accepted to use Let's Encrypt, please acknowledge TOS at %s and set GOTIFY_SERVER_SSL_LETSENCRYPT_ACCEPTTOS=true", tosURL)
```
jmattheis:
```suggestion
				log.Fatalf("Let's Encrypt TOS must be accepted to use Let's Encrypt, please acknowledge TOS at %s and set GOTIFY_SERVER_SSL_LETSENCRYPT_ACCEPTTOS=true\n", tosURL)
```

runner/runner.go
```diff
@@ -94,12 +97,43 @@ func getNetworkAndAddr(listenAddr string, port int) (string, string) {
...
+	} else if err != nil {
+		log.Printf("%s Request Failed: %s on %s %s\n", l.Name, err.Error(), r.Method, r.URL.String())
+	}
+	return
+}
+
 func applyLetsEncrypt(s *http.Server, conf *config.Configuration) {
+	httpClient := http.DefaultClient
+	httpClient.Transport = &LoggingRoundTripper{Name: "Let's Encrypt", RoundTripper: http.DefaultTransport}
+	httpClient.Timeout = 60 * time.Second
```
jmattheis: Is this a recommended timeout or where is the 60s coming from?

runner/runner.go
```diff
@@ -31,6 +32,8 @@ func Run(router http.Handler, conf *config.Configuration) error {
 	if conf.Server.SSL.Enabled {
 		if conf.Server.SSL.LetsEncrypt.Enabled {
 			applyLetsEncrypt(s, conf)
+		} else if conf.Server.SSL.CertFile != "" && conf.Server.SSL.CertKey != "" {
+			log.Fatalf("CertFile and CertKey must be set to use HTTPS when LetsEncrypt is disabled, please set GOTIFY_SERVER_SSL_CERTFILE and GOTIFY_SERVER_SSL_CERTKEY")
```
jmattheis:
```suggestion
			log.Fatalln("CertFile and CertKey must be set to use HTTPS when LetsEncrypt is disabled, please set GOTIFY_SERVER_SSL_CERTFILE and GOTIFY_SERVER_SSL_CERTKEY")
```

runner/runner.go
```diff
@@ -2,8 +2,8 @@ package runner
 
 import (
 	"context"
-	"crypto/tls"
 	"fmt"
+	"log"
 	"net"
 	"net/http"
 	"os"
```
jmattheis, line 9: I think we should add docker releases for master pushes similar to gotify/cli:master. So that we don't have to create full releases for debugging changes like this. (no action in this PR needed)

runner/runner.go
eternal-flame-AD: No, just to ensure it doesn't get stuck and eventually would print a message no matter what.. I don't know what is causing the issue but my experience with it is indeed.. sometimes feedback can be confusing

eternal-flame-AD: Anecdotally I was basing this on certbot's default timeout 30 secs for DNS propagation (and my experience that sometime it is just a little too short and need a couple more seconds), but whether it is relevant here I am not sure. I don't think it would hurt to be conservative though as this is a cold code path (once every 2 months) and have no user waiting in real time.

runner/runner.go
eternal-flame-AD: Yeah, good idea and can exercise more code path in CI, broken release CI has bit me 2 times on this..

jmattheis (dismissed):

eternal-flame-AD:
~Holding this for a moment pending if I can find the reason for #836, just so that we don't need to fix/add some other check later~

jmattheis (approved):

jmattheis:
Should these changes be released right now?

eternal-flame-AD:
I am not sure is this a regression (doesn't seem like it, but that could be just because this renewal code path is so cold regression don't show up quickly)? If so then probably. Otherwise up to you I am neutral 

I currently don't have a machine/IP I can test this process end to end, so far my testing only goes the challenge process works and errors work and ALPN happened not that it solved any particular issues, if you have one can you test it?

PR #852, opened 2025-10-01
Optimize uniq function for better performance

Ben1524:
Improve the performance of the unique function by:
1. Pre-allocating map capacity with len(s) to avoid frequent map resizing
2. Pre-allocating result slice capacity with len(s) to reduce append overhead
3. Reducing the number of traversals performs well under the condition of a large number of elements

These changes maintain the original behavior (preserving element order)  while reducing memory allocation operations, especially effective for  large slices (100k+ elements) with benchmark showing ~25% speedup.

No breaking changes, the function signature and output order remain unchanged.

### DIFF
--- a/api/stream/stream.go
+++ b/api/stream/stream.go
@@ -169,13 +169,13 @@ func (a *API) Close() {
 }
 
 func uniq[T comparable](s []T) []T {
-	m := make(map[T]struct{})
+	m := make(map[T]struct{}, len(s))
+	r := make([]T, 0, len(s))
 	for _, v := range s {
-		m[v] = struct{}{}
-	}
-	var r []T
-	for k := range m {
-		r = append(r, k)
+		if _, ok := m[v]; !ok {
+			m[v] = struct{}{}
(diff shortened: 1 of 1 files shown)


Ben1524:
Thanks for your review! I understand your point about the execution frequency and input size. Just wanted to optimize it as a precautionary measure. Glad you think it looks good otherwise. Let me know if there's anything else needed.



---- 回复的原邮件 ----
| 发件人 | 饺子w ***@***.***> |
| 日期 | 2025年10月02日 13:15 |
| 收件人 | ***@***.***> |
| 抄送至 | ***@***.***>***@***.***> |
| 主题 | Re: [gotify/server] Optimize uniq function for better performance (PR #852) |

@eternal-flame-AD approved this pull request.

Hi, thanks for the PR, currently this code is ran very infrequently (every a couple minutes) and input size is constrained by the # of open connections your system can take, so I doubt this is going to be the bottleneck on application performance.

But otherwise lgtm

PR #855, opened 2025-10-06
Delete an application's image

xavier-GitHub76:
I add a delete button (appearing when the mouse hovers over image)
After confirmation, the delete is done (the API 'DELETE /:id/image' is used) and a snack notification is created

### DIFF
--- a/ui/src/application/Applications.tsx
+++ b/ui/src/application/Applications.tsx
@@ -10,6 +10,7 @@ import TableRow from '@mui/material/TableRow';
 import Delete from '@mui/icons-material/Delete';
 import Edit from '@mui/icons-material/Edit';
 import CloudUpload from '@mui/icons-material/CloudUpload';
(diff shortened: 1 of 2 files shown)


xavier-GitHub76:
[Delete image.webm](https://github.com/user-attachments/assets/1a16f99a-c93a-4392-a6bb-d8d32b762eca)

ui/src/application/Applications.tsx
```diff
@@ -159,15 +171,75 @@ const Row = ({
     lastUsed,
     fDelete,
     fUpload,
+    fDeleteImage,
     image,
     fEdit,
 }: IRowProps) => {
+    const [isHovered, setIsHovered] = useState(false);
```
jmattheis: I've changed this to handle the hover effect purely with css and the image always has the onclick handler to delete the image. This make this simpler IMO.

PR #856, opened 2025-10-08
Sort dnd

xavier-GitHub76:
I merged the changes made by
jberlyn (from pull request https://github.com/gotify/server/pull/804 to the updated master branch)

I then added drag 'n' drop sorting support.

[dnd_sortorder.webm](https://github.com/user-attachments/assets/1a13bcbb-5567-454d-9be5-f66c58a3e432)

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -472,3 +472,99 @@ func ValidApplicationImageExt(ext string) bool {
 		return false
 	}
 }
+
+// ApplicationOrderParams represents the new order for applications
+type ApplicationOrderParams struct {
+	// Array of application IDs in the desired order
+	// required: true
+	// example: [3, 1, 5, 2, 4]
+	ApplicationIDs []uint `json:"applicationIds" binding:"required"`
+}
+
+// ReorderApplications updates the sort order of multiple applications based on their position in the array.
+// swagger:operation PUT /application/reorder application reorderApplications
+//
+// Reorder applications.
+//
+//	---
+//	consumes: [application/json]
+//	produces: [application/json]
+//	security: [clientTokenAuthorizationHeader: [], clientTokenHeader: [], clientTokenQuery: [], basicAuth: []]
+//	parameters:
+//	- name: body
+//	  in: body
+//	  description: ordered list of application IDs
+//	  required: true
+//	  schema:
+//	    $ref: "#/definitions/ApplicationOrderParams"
+//	responses:
+//	  200:
+//	    description: Ok
+//	    schema:
+//	      type: array
+//	      items:
+//	        $ref: "#/definitions/Application"
+//	  400:
+//	    description: Bad Request
+//	    schema:
+//	        $ref: "#/definitions/Error"
+//	  401:
+//	    description: Unauthorized
+//	    schema:
+//	        $ref: "#/definitions/Error"
+//	  403:
+//	    description: Forbidden
+//	    schema:
+//	        $ref: "#/definitions/Error"
+//	  500:
+//	    description: Server Error
+//	    schema:
+//	        $ref: "#/definitions/Error"
+func (a *ApplicationAPI) ReorderApplications(ctx *gin.Context) {
+	userID := auth.GetUserID(ctx)
+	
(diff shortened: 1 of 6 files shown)


ui/src/application/Applications.tsx
```diff
@@ -54,6 +88,27 @@ const Applications = observer(() => {
         }
     };
 
+    const handleDragEnd = (event: DragEndEvent) => {
+        const {active, over} = event;
+
+        if (over && active.id !== over.id) {
+            setLocalApps((items) => {
```
jmattheis, line 95: I don't think we need to expose both the sort order integer and the new api with the list of application ids. One should suffice.

If we choose the new reorder api we can hide the sort index in the application model which prevents breaking api changes (new field in the `PUT /application/{id}` endpoint). This would also expose less internals in the public api.

I think there is an argument for the sort index on the application. [Fractional Indexing](https://observablehq.com/@dgreensp/implementing-fractional-indexing) could be used that has some feature for collaborative editing (which we don't need) but this also allows us to move an application to a specific position without having to update the sort order index of all applications. The fractional index would be a text field that's sortable and the client would be able to update the fractionalIndex in the `PUT /application/{id}` endpoint with something like https://www.npmjs.com/package/fractional-indexing.

I don't think I've a clear best solution here. I'm slightly leaning towards the new api as it seems less complex but the fractional index solution only has to update a single application on reordering and doesn't need another endpoint.

@eternal-flame-AD do you have an opinion on this?

matezoltanfarkas:
is there an update on this @eternal-flame-AD, @jmattheis and @xavier-GitHub76 ?
this looks slick!

ui/src/application/Applications.tsx
eternal-flame-AD: Shoot sorry, this drop out of my radar through a mixture of my own fault and a slightly confusing title .. 

Fractional indexing should be fine, I don't really think adding an optional field with a sensible default counts as breaking change.

However my understanding is fractional indexing is not concurrent safe? Let's say you have app with ID 1,2 and 3 and you want to put 3 between 1 and 2. By the time you retrieved the fractional keys for 1 and 2 and generated a new key for 3, you can't say put "newKey" between 1 and 3 (as the keys for them have already changed). I think this might be the "collaborative editing" part. 

Solutions can include:
- Don't worry about it. It's much safer than the current approach (a complete state overwrite) and a botched operation should not cause data loss or anything irreversible.
- Read-Check-Update: you don't say "insert newKey between 1 and 2", you say "insert newKey between oldKeyFor1 and oldKeyFor2" and loop until the server successfully made one single atomic transaction.
- Just design a separate API specifically for "put A between B and C, by ID" and write all the indexing logic on the server side

api/application.go
```diff
@@ -466,3 +472,99 @@ func ValidApplicationImageExt(ext string) bool {
...
+
+	updatedApps := make([]*model.Application, 0, len(orderParams.ApplicationIDs))
+	for i, appID := range orderParams.ApplicationIDs {
+		app, err := a.DB.GetApplicationByID(appID)
+		if success := successOrAbort(ctx, 500, err); !success {
+			return
+		}
+		if app != nil {
+			app.SortOrder = i
+			if success := successOrAbort(ctx, 500, a.DB.UpdateApplication(app)); !success {
```
eternal-flame-AD, line 562: For future reference (after the strategy is settled), please only update the column you need, especially if the operation is not atomic (you are repeating a lot of read/writes in a loop), can happen in a "rapid fire" fasion (a user typically wants to DND a lot at once) and touch critical fields (you overwrote all fields)

What we don't want to happen is to have a stalled DB query (which can happen for a lot of reasons, a contending lock or just packet loss) cause transactions to race with transactions for the next DND request, which can lead to incoherent state or even data corruption.

ui/src/application/Applications.tsx
jmattheis: I think just don't worry about it. We don't have an offline mode, so I think we can just ignore it. I'll take over this PR, and use fractional indexing.

jmattheis:
Will be continued in #909

PR #857, opened 2025-10-11
fix: fix markdown embed image issue

ggqshr:
The new version of **react-markdown** cannot display images in the **embed** format. This issue can be fixed by following the method provided in the corresponding repository. Related issue: [https://github.com/remarkjs/react-markdown/issues/774#issuecomment-1743115270](https://github.com/remarkjs/react-markdown/issues/774#issuecomment-1743115270).

### DIFF
--- a/ui/src/common/Markdown.tsx
+++ b/ui/src/common/Markdown.tsx
@@ -6,7 +6,11 @@ import gfm from 'remark-gfm';
 // Copy from mlflow/server/js/src/shared/web-shared/genai-markdown-renderer/GenAIMarkdownRenderer.tsx
 // Related PR: https://github.com/mlflow/mlflow/pull/16761
 const urlTransform: UrlTransform = (value) => {
-    if (value.startsWith('data:image/png;') || value.startsWith('data:image/jpeg;') || value.startsWith('data:image/gif;')) {
+    if (
+        value.startsWith('data:image/png;') ||
+        value.startsWith('data:image/jpeg;') ||
+        value.startsWith('data:image/gif;')
+    ) {
         return value;
     }
     return defaultUrlTransform(value);


eternal-flame-AD:
Hi, thanks for the contribution.

The security model of markdowns in gotify is it should at least not XSS. This seems to me will allow `javascript:` URLs to be rendered

It seems like the discussion you quoted says you should use defaultUrlTransform by default and explicitly handle other known good cases if needed.

ggqshr:
Understood, this indeed poses some security concerns. So, if we specially handle embedded image URLs and use defaultUrlTransform for non-image URLs, would this approach work?


Related PR: https://github.com/mlflow/mlflow/pull/16761#discussion_r2210268695

eternal-flame-AD:
Nice find, that should do it. I would match a semicolon after that as well just in case.

ggqshr:
Done, and I also agree to include the semicolon in the match.

ui/src/common/Markdown.tsx
```diff
@@ -1,7 +1,17 @@
 import React from 'react';
-import ReactMarkdown from 'react-markdown';
+import ReactMarkdown, {defaultUrlTransform} from 'react-markdown';
+import type {UrlTransform} from 'react-markdown';
 import gfm from 'remark-gfm';
 
+// Copy from mlflow/server/js/src/shared/web-shared/genai-markdown-renderer/GenAIMarkdownRenderer.tsx
+// Related PR: https://github.com/mlflow/mlflow/pull/16761
+const urlTransform: UrlTransform = (value) => {
+    if (value.startsWith('data:image/png;') || value.startsWith('data:image/jpeg;')) {
```
jmattheis: In the gotify/android we support [a few more formats](https://developer.android.com/media/platform/supported-formats#image-formats), maybe we could add image/gif here too? This seems like a commonly used format. For the rest we can wait until someone reports an issue.

ggqshr:
Good point! I've added support for GIFs — that seems like a sensible addition. We can wait for further reports before adding more formats.

Issue #667, opened 2024-07-17
Database has no foreign key


xavier-GitHub76:
Hello,

Database has no foreign key : 
![image](https://github.com/user-attachments/assets/0fe1d789-04f9-4fcc-9f4e-e1c19cdcbca3)

It could be possible to insert bad value on referenced key.

The attached script add foreign keys.
[Foreign Keys for gotify.zip](https://github.com/user-attachments/files/16267329/Foreign.Keys.for.gotify.zip)

After, the database structure will be : 
![image](https://github.com/user-attachments/assets/bf1d3589-ff30-4f8f-aeee-457620229f03)

Best regards

jmattheis:
I agree it would be more robust when there are foreign keys, but this is more of an improvement than a bug.

xavier-GitHub76:
I agree with you. Thanks for your work.

PR #863, opened 2025-10-16
chore: Migrate github.com/jinzhu/gorm to gorm.io/gorm

### LINKED ISSUE #862: Migrate github.com/jinzhu/gorm to gorm.io/gorm

The old repo has moved to gorm.io/gorm 4 years ago and we have not updated since due to renovate not tracking repository moves.

Migrate code to new import path and bump version.

**Additional context**

In the future explore "github.com/glebarez/sqlite" to make builds easier with tradeoff being one more middleman on the supply chain, but we will stay with the C compiled sqlite for now.

eternal-flame-AD:
Closes #862 .

I removed the manual column fixup: The default column type is already bloba and longblob (I tested on the docker container for MariaDB and postgres).

I also dropped a test that is basically if the database returns error then the function has to return an error, feels a little mechanical coverage driven test not really useful.

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -36,7 +36,8 @@ func New(dialect, connection, defaultUser, defaultPass string, strength int, cre
 		Colorful:                  isatty.IsTerminal(os.Stderr.Fd()),
 	})
 	gormConfig := &gorm.Config{
-		Logger: dbLogger,
+		Logger:                                   dbLogger,
+		DisableForeignKeyConstraintWhenMigrating: true,
 	}
 
 	var db *gorm.DB


database/database.go
```diff
@@ -19,40 +20,52 @@ var mkdirAll = os.MkdirAll
...
-	db.DB().SetMaxOpenConns(10)
+	sqldb.SetMaxOpenConns(10)
 
 	if dialect == "sqlite3" {
 		// We use the database connection inside the handlers from the http
 		// framework, therefore concurrent access occurs. Sqlite cannot handle
 		// concurrent writes, so we limit sqlite to one connection.
 		// see https://github.com/mattn/go-sqlite3/issues/274
-		db.DB().SetMaxOpenConns(1)
+		sqldb.SetMaxOpenConns(1)
```
eternal-flame-AD, line 73: It seems we can drop this, it has been fixed a long time a go in https://github.com/mattn/go-sqlite3/commit/acfa60124032040b9f5a9406f5a772ee16fe845e

eternal-flame-AD: Okay it seems to be not that simple .. let's keep it for now

database/database.go
```diff
@@ -19,40 +24,67 @@ var mkdirAll = os.MkdirAll
 func New(dialect, connection, defaultUser, defaultPass string, strength int, createDefaultUserIfNotExist bool) (*GormDatabase, error) {
 	createDirectoryIfSqlite(dialect, connection)
 
-	db, err := gorm.Open(dialect, connection)
+	logLevel := logger.Info
+	if mode.Get() == mode.Prod {
+		logLevel = logger.Warn
+	}
+
+	dbLogger := logger.New(log.New(os.Stderr, "\r\n", log.LstdFlags), logger.Config{
```
jmattheis, line 32: Why the windows line endings??

eternal-flame-AD: I copied it from the official code, didn't ask too many questions:

https://github.com/go-gorm/gorm/blob/b88148363a954f69fa680b152dfd96a94ffea1e1/logger/logger.go#L75

model/application.go
jmattheis: It seems like the unique index doesn't work anymore.

(postgres: master)
```
gotify=# \d clients
                                      Table "public.clients"
  Column   |           Type           | Collation | Nullable |               Default
-----------+--------------------------+-----------+----------+-------------------------------------
 id        | integer                  |           | not null | nextval('clients_id_seq'::regclass)
 token     | character varying(180)   |           |          |
 user_id   | integer                  |           |          |
 name      | text                     |           |          |
 last_used | timestamp with time zone |           |          |
Indexes:
    "clients_pkey" PRIMARY KEY, btree (id)
    "idx_clients_user_id" btree (user_id)
    "uix_clients_id" UNIQUE, btree (id)
    "uix_clients_token" UNIQUE, btree (token)
```
(postgres: this branch without preexisting database)
```
gotify=# \d clients
                                      Table "public.clients"
  Column   |           Type           | Collation | Nullable |               Default
-----------+--------------------------+-----------+----------+-------------------------------------
 id        | bigint                   |           | not null | nextval('clients_id_seq'::regclass)
 token     | character varying(180)   |           |          |
 user_id   | bigint                   |           |          |
 name      | text                     |           |          |
 last_used | timestamp with time zone |           |          |
Indexes:
    "clients_pkey" PRIMARY KEY, btree (id)
    "idx_clients_user_id" btree (user_id)
Foreign-key constraints:
    "fk_users_clients" FOREIGN KEY (user_id) REFERENCES users(id)
```
`UNIQUE, btree (token)` isn't created anymore.

`uix_clients_id` seems weird, it's releated to the unique_index on the ID which is already primary key. We can remove this. But we can do this in a separate PR.

api/application_test.go
```diff
@@ -611,16 +611,16 @@ func upload(values map[string]*os.File) (contentType string, buffer bytes.Buffer
 	for key, r := range values {
```
jmattheis, line 611: Postgres foreign keys will be updated to bigint, but ids stay integer.

(master)
```
gotify=# \d plugin_confs
                                        Table "public.plugin_confs"
     Column     |          Type          | Collation | Nullable |                 Default
----------------+------------------------+-----------+----------+------------------------------------------
 id             | integer                |           | not null | nextval('plugin_confs_id_seq'::regclass)
 user_id        | integer                |           |          |
 module_path    | text                   |           |          |
 token          | character varying(180) |           |          |
 application_id | integer                |           |          |
 enabled        | boolean                |           |          |
 config         | bytea                  |           |          |
 storage        | bytea                  |           |          |
Indexes:
    "plugin_confs_pkey" PRIMARY KEY, btree (id)
    "idx_plugin_confs_id" btree (id)
    "uix_plugin_confs_token" UNIQUE, btree (token)
```
after running this branch:
```
gotify=# \d plugin_confs
                                        Table "public.plugin_confs"
     Column     |          Type          | Collation | Nullable |                 Default
----------------+------------------------+-----------+----------+------------------------------------------
 id             | integer                |           | not null | nextval('plugin_confs_id_seq'::regclass)
 user_id        | bigint                 |           |          |
 module_path    | text                   |           |          |
 token          | character varying(180) |           |          |
 application_id | bigint                 |           |          |
 enabled        | boolean                |           |          |
 config         | bytea                  |           |          |
 storage        | bytea                  |           |          |
Indexes:
    "plugin_confs_pkey" PRIMARY KEY, btree (id)
    "idx_plugin_confs_id" btree (id)
    "uix_plugin_confs_token" UNIQUE, btree (token)
Foreign-key constraints:
    "fk_users_plugins" FOREIGN KEY (user_id) REFERENCES users(id)
```
application_id is migrated to bigint, but id stays integer. It's probably good enough for now, but in the future we should migrate the id to bigint too, otherwise this seems inconsistent.

api/application_test.go
jmattheis, line 609: I've ran the ui e2e tests against postgres + mysql and they seem to work. Only failure is the user e2e test for postgres, but it's also not working on master. It looks like the table changes ordering after the password was changed. Can stay as is.

eternal-flame-AD: Okay, I will fix your previous two problems next week.

model/application.go
jmattheis: unique_index seems to have been renamed to uniqueIndex: https://gorm.io/docs/indexes.html#uniqueIndex

api/application_test.go
eternal-flame-AD: Oh a new DB it gets migrated to bigserial, which aligns with the current bigint type.

If you really want to think about the edge cases they are not the same thing and there is a nonzero chance the migration would hit someone's database with out of range IDs, so I can understand why Gorm refiuse to migrate them implicitly (bigserial starts with 1, Go uint starts with 0, int starts at -2147483648), but for practical purposes I agree with you keeping the primary key int should not pose any short term risk

model/application.go
eternal-flame-AD: Done, and checked the unique constraints are now created.

eternal-flame-AD:
I checked the unique index now generates correctly on postgres. Thanks for the catch. The primary key one doesn't look super actionable right now, the current behavior looks fine to me (migrate the foreign key to a supertype but keep primary key as is on an existing DB)

model/application.go
```diff
@@ -13,13 +13,13 @@ type Application struct {
...
 	// example: 5
-	ID uint `gorm:"primary_key;unique_index;AUTO_INCREMENT" json:"id"`
+	ID uint `gorm:"primary_key;uniqueIndex;AUTO_INCREMENT" json:"id"`
 	// The application token. Can be used as `appToken`. See Authentication.
 	//
 	// read only: true
 	// required: true
 	// example: AWH0wZ5r0Mbac.r
-	Token  string `gorm:"type:varchar(180);unique_index" json:"token"`
+	Token  string `gorm:"type:varchar(180);uniqueIndex" json:"token"`
```
jmattheis: Now unique indexes are duplicated with a database initialized before this change.

(master)
```
gotify=# \d clients
                                      Table "public.clients"
  Column   |           Type           | Collation | Nullable |               Default
-----------+--------------------------+-----------+----------+-------------------------------------
 id        | integer                  |           | not null | nextval('clients_id_seq'::regclass)
 token     | character varying(180)   |           |          |
 user_id   | integer                  |           |          |
 name      | text                     |           |          |
 last_used | timestamp with time zone |           |          |
Indexes:
    "clients_pkey" PRIMARY KEY, btree (id)
    "idx_clients_user_id" btree (user_id)
    "uix_clients_id" UNIQUE, btree (id)
    "uix_clients_token" UNIQUE, btree (token)
```
After running this branch:
```
gotify=# \d clients
                                      Table "public.clients"
  Column   |           Type           | Collation | Nullable |               Default
-----------+--------------------------+-----------+----------+-------------------------------------
 id        | integer                  |           | not null | nextval('clients_id_seq'::regclass)
 token     | character varying(180)   |           |          |
 user_id   | bigint                   |           |          |
 name      | text                     |           |          |
 last_used | timestamp with time zone |           |          |
Indexes:
    "clients_pkey" PRIMARY KEY, btree (id)
    "idx_clients_id" UNIQUE, btree (id)
    "idx_clients_token" UNIQUE, btree (token)
    "idx_clients_user_id" btree (user_id)
    "uix_clients_id" UNIQUE, btree (id)
    "uix_clients_token" UNIQUE, btree (token)
Foreign-key constraints:
    "fk_users_clients" FOREIGN KEY (user_id) REFERENCES users(id)
```
idx_clients_token and uix_clients_token is the same.

Mabye we can explicitly define the name of the unique index, so we can use the old name.

model/application.go
```diff
@@ -13,13 +13,13 @@ type Application struct {
 	// read only: true
 	// required: true
 	// example: 5
-	ID uint `gorm:"primary_key;unique_index;AUTO_INCREMENT" json:"id"`
+	ID uint `gorm:"primary_key;uniqueIndex;AUTO_INCREMENT" json:"id"`
```
jmattheis: While at it, could you remove the uniqueIndex from the primary key columns. Primary key already indicates unique index.

model/application.go
eternal-flame-AD: Hmm, I will try that later today. Seems to be a breaking change that should impact many projects though but I can't find any mention of this issue in the migration guide or GitHub issues in general.

eternal-flame-AD: Done, I checked at least master branch migrates without duplicating indices.

I will leave the existing PK duplicate in place .. haven't figured out a very ergonomic way to fix it up. If we really need a "consistent" schema I think the current way is not the right way to do it, maybe use a migrator with proper schema versioning.

model/application.go
```diff
@@ -13,7 +13,7 @@ type Application struct {
 	// read only: true
 	// required: true
 	// example: 5
-	ID uint `gorm:"primary_key;uniqueIndex;AUTO_INCREMENT" json:"id"`
```
jmattheis: There is another `index` on PluginConf#ID.

model/application.go
```diff
@@ -13,7 +13,7 @@ type Application struct {
...
+	ID uint `gorm:"primary_key;AUTO_INCREMENT" json:"id"`
 	// The application token. Can be used as `appToken`. See Authentication.
```
jmattheis, line 17: The mysql/mariadb upgrade seems to crash while creating a foreign key.
```
$ mariadb -h 127.0.0.1 -u root --password gotify
MariaDB [gotify]> show tables;
Empty set (0.000 sec)

(on master) $ go run .
Starting Gotify version unknown@unknown
....
Started listening for plain connection on tcp [::]:8080
^CShutting down: received signal interrupt
(change to gorm-io-migrate) $ git ch -
Switched to branch 'gorm-io-migrate'
Your branch is up to date with 'origin/gorm-io-migrate'.
(on gorm-io-migrate) $ go run .
Starting Gotify version unknown@unknown

...

2025/10/26 10:13:46 /home/jm/src/gotify/server/database/database.go:83
[0.342ms] [rows:-] SELECT count(*) FROM INFORMATION_SCHEMA.table_constraints WHERE constraint_schema = 'gotify' AND table_name = 'applications' AND constraint_name = 'fk_users_applications'

2025/10/26 10:13:46 /home/jm/src/gotify/server/database/database.go:83 Error 1005 (HY000): Can't create table `gotify`.`applications` (errno: 150 "Foreign key constraint is incorrectly formed")
[12.885ms] [rows:0] ALTER TABLE `applications` ADD CONSTRAINT `fk_users_applications` FOREIGN KEY (`user_id`) REFERENCES `users`(`id`)
panic: Error 1005 (HY000): Can't create table `gotify`.`applications` (errno: 150 "Foreign key constraint is incorrectly formed")

goroutine 1 [running]:
main.main()
        /home/jm/src/gotify/server/app.go:44 +0x33f
exit status 2
```

Tested with the mysql:latest and mariadb:latest docker containers. Both error there but with slightly different error messages. Mysql show this error:

```
2025/10/26 10:22:54 /home/jm/src/gotify/server/database/database.go:83 Error 3780 (HY000): Referencing column 'user_id' and referenced column 'id' in foreign key constraint 'fk_users_applications' are incompatible.
[0.282ms] [rows:0] ALTER TABLE `applications` ADD CONSTRAINT `fk_users_applications` FOREIGN KEY (`user_id`) REFERENCES `users`(`id`)
panic: Error 3780 (HY000): Referencing column 'user_id' and referenced column 'id' in foreign key constraint 'fk_users_applications' are incompatible.

goroutine 1 [running]:
main.main()
        /home/jm/src/gotify/server/app.go:44 +0x33f
exit status 2
```

jmattheis (commented):
postgres and sqlite looks good, but mysql/mariadb has some problems.

model/application.go
eternal-flame-AD: Hmmm, can you give me your config file? I can't replicate it .

Here's my setup and log file

https://gist.github.com/eternal-flame-AD/04928a7bb6cc98bdbacfc534a7011095

jmattheis: I can reproduce it with your config. Have you ran gotify/server with the current master version, and afterwards with the changes from this PR?

If that doesn't work for you, here's a dump from the master state: [init.sql](https://github.com/user-attachments/files/23197094/init.sql) when importing this into a mysql database and running this branch the error occurs for me.

eternal-flame-AD: Okay I know, it's likely the type is inconsistent and MySQL didn't like it..  I will disable the new auto foreign key feature first. I think this is beyond fixing using just AutoMigrate.

If foreign key is really an issue we should use versioned migration because there is already no guarantee existing user's database is consistent and I wouldn't bet just fixing the type issue make it safe for release.

Crossref: https://github.com/gotify/server/issues/667

PR #869, opened 2025-11-01
Fix overlapping audio notifications for high-priority messages

### LINKED ISSUE #850: Throttle notification sound in web UI
Sometime I get a massive flood of notifications. For example I rename a season folder in jellyfin and get a flood of delete and added notifications at once.

This comes with an annoying flood of notification sounds trying to play simultaneously.

A simple solution would be to throttle the notification sound. Maybe no more than 1 sound per 2 seconds?

LaurenceJJones:
fix #850

- Add audioNotification utility to manage single Audio instance reuse
- Debounce audio playback with 300ms cooldown after playback ends
- Skip playback if audio is currently playing or in cooldown window
- Replace per-message Audio creation with shared instance in reactions.ts

### DIFF
--- a/ui/src/reactions.ts
+++ b/ui/src/reactions.ts
@@ -2,6 +2,8 @@ import {reaction} from 'mobx';
 import * as Notifications from './snack/browserNotification';
 import {StoreMapping} from './stores';
 
+const AUDIO_REPEAT_DELAY = 1000;
+
 export const registerReactions = (stores: StoreMapping) => {
     const clearAll = () => {
         stores.messagesStore.clearAll();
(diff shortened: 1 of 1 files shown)


ui/src/audioNotification.ts
```diff
@@ -1,19 +1,22 @@
 let audioInstance: HTMLAudioElement | null = null;
-let isPlaying = false;
 let cooldownEndTime = 0;
```
jmattheis: I dislike global state as we can get weird side-effects from this. The caching state should be local in some function, or be inside the statemanagement.

ui/src/audioNotification.ts
```diff
@@ -1,19 +1,22 @@
...
     const now = Date.now();
 
-    // Skip if currently playing or in cooldown
-    if (isPlaying || now < cooldownEndTime) {
+    // Skip if in cooldown
+    if (now < cooldownEndTime) {
+        return;
+    }
+
+    // Check if audio is currently playing
```
jmattheis: With the comments this looks AI generated, and the solution seems overly complicated to be reusable by passing `src` as parameter.

If we'd call this method with different `src`, only the last one is cached, and the throttling would be global, but yeah, this complexity isn't needed now.

ui/src/audioNotification.ts
```diff
@@ -23,15 +26,15 @@ export function playNotificationAudio(src: string = DEFAULT_SRC): void {
         audioInstance.addEventListener('ended', handleAudioEnd);
```
jmattheis: Given that we know how long the audio notification is, we can omit this dynamic handling and define a countdown that's audio duration + cooldown. This will make this simpler.

jmattheis (approved):

ui/src/audioNotification.ts
LaurenceJJones: 100% makes sense

ui/src/audioNotification.ts
LaurenceJJones: Yes, claude was used to scaffold and I thought, hmmm yeah an outside component could be good if we plan to make apps have customized sounds. However, your right the functionality isnt there and plus I dont know if its ever planned so extracting a function is just unnecessary.

PR #872, opened 2025-11-05
Add file type restriction to image upload input using accept attribute

LaurenceJJones:
# Restrict File Upload Types

## Summary
Adds file type restriction to the application image upload input using the `accept` attribute to improve user experience and prevent invalid file selections.

## Changes
- Added `accept="image/png,image/jpeg,image/gif"` attribute to the file input element in `Applications.tsx`

## Benefits
- **Better UX**: File picker now defaults to showing only supported image types (PNG, JPEG, GIF)
- **Reduces errors**: Users are less likely to select unsupported file types
- **Prevents confusion**: No need to manually filter through all file types

## Technical Details
- The `accept` attribute provides a hint to the browser's file picker to filter files
- Existing validation logic remains in place as a fallback for edge cases
- Backward compatible - no breaking changes

## Testing
- Verify file picker shows only image files on click
- Confirm validation still works if invalid file type is selected (edge case)
- Test on multiple browsers (Chrome, Firefox, Safari)

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 	"github.com/gotify/server/v2/auth"
@@ -459,7 +460,7 @@ func generateNonExistingImageName(imgDir string, gen func() string) string {
 }
 
 func ValidApplicationImageExt(ext string) bool {
-	switch ext {
+	switch strings.ToLower(ext) {
 	case ".gif", ".png", ".jpg", ".jpeg":
 		return true
 	default:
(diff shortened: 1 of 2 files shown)


eternal-flame-AD:
It seems just put the extension explicitly would be great.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/file#unique_file_type_specifiers

LaurenceJJones:
Should we also update the pre check here as it based on mime

https://github.com/gotify/server/blob/f263221133fcbd30ec9a78eaef16f9dc69f86f9c/ui/src/application/Applications.tsx#L74-L76

eternal-flame-AD:
Hmm, honestly if it catches error early enough I would say good enough. I might even say the additional MIME checks are redundant (afaik they don't check the file contents? It's just a static mapping based on extension).

It seems the server side check is the strictest, as long as it isn't stricter than that I would say LGTM.

eternal-flame-AD:
I think I can take it over from now, I will do a cross platform test later today and if needed I will just push directly. Thanks for the PR!

eternal-flame-AD (approved):

Issue #870, opened 2025-11-02
[Bug/Feature Request/] Strip `priority` of quotes/apostrophes to prevent marshal cast error


OdinVex:
**Is your feature request related to a problem? Please describe.**
Several backends I use for API automation (such as Huginn) try passing every JSON parameter and value enclosed in quotes, such as `"priority": "10"`. This causes Gotify to return a 400 because `"10"` is being passed to the API instead of `10`, in this example.

**Describe the solution you'd like**
Allow for some flexibility in the parameter for parsing as string->int or int by removing quotes/apostrophes so either becomes capable.

**Describe alternatives you've considered**
Rewriting various backends/other software, but that's insurmountable.

**Additional context**
`message` API, `title`, `priority`, `message` are all set. Ex:

```
{
  "title": "...",
  "priority": "10",
  "message": "...",
  "body": "{\"error\":\"Bad Request\",\"errorCode\":400,\"errorDescription\":\"json: cannot unmarshal string into Go struct field MessageExternal.priority of type int\"}",
  "status": 400,
  "headers": {
    ... Removed for Brevity ...
  }
}
```

jmattheis:
I'm unsure about this, I like endpoints being strict as this makes the implementation more easy and forces you to use it correctly.

I'd probably be okay with making an exception for the priority field and allowing string numbers there when there are more tools that force number strings in json. If it's just one, then I don't think I want this included.

OdinVex:
I'd normally be entirely in agreement about API endpoints, but I think integer/float parsing from string to int/float is rather common among API handling, in general, because it's rather common and safe. I'll try to raise the issue with Huginn and the other backends for now, thanks for following up.

LaurenceJJones:
referencing the upstream issue https://github.com/huginn/huginn/issues/2661 to help other users find the suggested fix (even though people reported it didnt but linking helps people find it faster 😅 )

OdinVex:
> referencing the upstream issue [huginn/huginn#2661](https://github.com/huginn/huginn/issues/2661) to help other users find the suggested fix (even though people reported it didnt but linking helps people find it faster 😅 )

I forgot to do that, thank you very much. 👍

eternal-flame-AD:
Thanks for the context. I think one way we can potentially do this without hacks is allow priority to be specified on the URL even if the payload is submitted in JSON, I assume you could do something like:

```jsonc
{
  "post_url": "https://my.gotify.box/message?priority={{priority}}",
   // the other fields
}
[...]

Issue #642, opened 2024-03-25
Accept priority as string


lezruk:
**Is your feature request related to a problem? Please describe.**
Home-Assistant notify service with POST_JSON type always sends priority as sting in resulting JSON due to templating engine (similar how ansible jinja templating works), that is not accepted by gotify server with 400 bad requests, as it accepts strict integer value for priority only

**Describe the solution you'd like**
gotify server to accept priority as string, and throw an error only if string value is not castable to integer

**Describe alternatives you've considered**
No real alernativies found so far, only hardcoding particular value/default priority which doesn't serve the purpose of prioirty differentiation within single app messages

**Additional context**
N/A

LaurenceJJones:
Hey 👋🏻 

Looking at [home assistant docs](https://www.home-assistant.io/docs/configuration/templating/#important-template-rules) they provide a link to -> https://jinja.palletsprojects.com/en/latest/templates/#jinja-filters.int

I dont know what the template look like but I guess you should be able to type conversion?

lezruk:
All these filters are fine, but problem is that POST_JSON result is ALWAYS sent in quotes, which is interpreted as string. Thus gotify rejects it.

jmattheis:
You can set the priority like this in home assistent (also included markdown support):
```yml
notify:
  - name: gotify_high
    platform: rest
    resource: https://gotify.example.org/message
    method: POST_JSON
    headers:
      X-Gotify-Key: AsQHYA96LH90.RT
    message_param_name: message
    title_param_name: title
    data:
      priority: 10
      extras:
        client::display:
          contentType: "text/markdown"
```

lezruk:
Appreciate your answer, hardcoding value in data.extras works. Idea though is to avoid hardcoding and send priority as a value for individual message call within same service. Therefore I opened this request for evaluation.

jmattheis:
How do I manually trigger an alert with priority in home assistant? developer-tools/service doesn't have a priority.

lezruk:
> How do I manually trigger an alert with priority in home assistant? developer-tools/service doesn't have a priority.

There is target parameter for notification service, that you can map to priority for gotify call. But again internally that is sent in JSON in quotes and doesn't work. Other option would be sending payload as fields in POST call, yet in this case markdown is lost as gotify doesn't support extras parameters without json. I saw somewhere here request to implement extras fields params i gotify but that was not done

aagit:
[...]

PR #887, opened 2025-12-08
Allow priority to be passed as a string

nienow:
There are two issues surrounding this: #870 and #642, but I've seen mentions of this limitation when trying to use other tools as well. I see there is debate on whether this should be done or not, but I want it for my server, so I was going to implement it anyway.

I'm newer to Go, so I apologize if this could be done simpler, but this was the best way I found. 

I created an intermediate struct so I didn't have to change the type of priority on MessageExternal.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -45,7 +45,7 @@ type pagingParams struct {
 	Since uint `form:"since" binding:"min=0"`
 }
 
-// using custom struct instead of MessageExternal so priority can be a string
+// Using custom struct instead of MessageExternal so priority can be a string.
 type createMessageRequest struct {
 	Message  string                 `form:"message" query:"message" json:"message" binding:"required"`
 	Title    string                 `form:"title" query:"title" json:"title"`


eternal-flame-AD:
Hi, whether this "should be done or not" really needs to be deferred to Jannis but his position is pretty clear and I respect that.

However , You do mention "I've seen mentions of this limitation when trying to use other tools as well", do you know some concrete examples? I think this is the major source of contention so if you have concrete examples I think he might reconsider.

Code wise I think no substantiative change is needed, nitpicks can be deferred later, one potential concern is please verify floating point strings are rejected (NaN, 1.2, 1.2e-3, etc).

Edit: I have to be away for most of next week, consider me green lighted this PR on the condition Jannis is satisfied with the demonstrated need.

nienow:
As for the tools, home assistant is the big one. There is a workaround by setting up multiple notification services with different default priorities, its just not ideal. Some discussion here: https://community.home-assistant.io/t/self-hosted-push-notifications-with-gotify/120307?page=2

huginn is another: https://github.com/huginn/huginn/issues/2661

And I know I've run into this with at least one other tool in the past, and had found discussions about others having the same issue, but honestly don't remember what it was (and its obviously a tool I don't use anymore).

api/message.go
```diff
@@ -45,6 +45,14 @@ type pagingParams struct {
 	Since uint `form:"since" binding:"min=0"`
```
jmattheis, line 45: @nienow would something like a raw message API solve your problem?

E.g. title, extras, title can be optionally sent as header values, and the full message body is the gotify message.
```http
POST /message/raw
Content-Type: text/plain
X-Gotify-Title: my title
X-Gotify-Priority: 10
X-Gotify-Extras: { "client::display": { "contentType": "text/markdown" } }

this is the message **body**....
```

nienow: No, home assistant cannot pass variables into headers

nienow:
After talking with AI for a while, I ended up finding a great solution to my issue with home assistant. Instead of trying to make the notification service work with gotify, I setup a rest_command with a script. The benefits of this is that you can define the priority as a number,  restrict it from 1 to 10, and give it a default. You then call the script in automations. I will be closing this PR because I like this new setup much better. I will suggest it in the various discussions surrounding home assistant and gotify.

```
rest_command:
  gotify_command:
    url: https://my_gotify_server/message
    method: POST
    headers:
      authorization: 'Bearer MY_TOKEN'
      content-type: 'application/json'
    payload: >
      {
        "title": "{{ title }}",
        "message": "{{ message }}",
        "priority": {{ priority }}
      }
          
script:
  gotify:
    alias: Gotify Notification
    fields:
      message:
        name: Message
        required: true
        selector:
          text:
      title:
        name: Title
        required: false
        selector:
          text:
      priority:
        name: Priority
        default: 5
        selector:
          number:
            min: 1
            max: 10
    sequence:
      - service: rest_command.gotify_command
        data:
          message: "{{ message }}"
          title: "{{ title }}"
          priority: "{{ priority }}"
```

Issue #34, opened 2018-03-20
Paginate Message APIs


jmattheis:
Adjust APIs:
* `GET /application/{id}/message`
* `GET /message`

New Query Parameter:
* `limit=uint` the amount of messages to return
* ~`since=date` return all messages before this date (limited by limit)~
* `since=id` return all messages before the given message id

Expected Response:
``` json
{
    "paging": {
       "limit": 50,
       "next": "https://requesturl.com/message?limit=from-request&since=message-id"
    },
    "data": [
          {"id":12183,"appid":1,"message":"mymsg","title":"mytitle","priority":0,"date":"2018-03-20T19:30:51.596354+01:00"}
    ]
}
```

lslah:
`since=id` is probably better as two messages might have the same date, see https://coderwall.com/p/lkcaag/pagination-you-re-probably-doing-it-wrong

jmattheis:
@lslah Using ids for pagination seems to be better. Good point!

naimo:
The name of the parameter seems weird to me : message ids get incremented, so the message received by the server "since" a given id are those with higher numbers.

Keeping the current behavior, the parameter should be renamed "until" or "before". Better though would be to keep "since" but actually return new messages. Seems to be a more frequent use case no ?

jmattheis:
@naimo Naming is hard (:.

I think, I called it this way because the messages are sorted in descending order. Thus, `since` kinda makes sense. The behavior is also [documented](https://gotify.net/api-docs#/message/getMessages)
```
Ascending:
Message 1 |- until 3
Message 2 |
Message 3
Message 4 | 
Message 5 |- since 3

Descending:
Message 5 |- until 3
Message 4 |
Message 3
Message 2 | 
Message 1 |- since 3
```
I don't think this renaming would justify a breaking API change. 

> Better though would be to keep "since" but actually return new messages. Seems to be a more frequent use case no ?

It is actually not because new messages will be transmitted via web socket. Still, the API addition would have some use cases.

naimo:
I see.

In my case I have a client (small esp8366-based display) which polls for
new messages on regular basis, not via websocket. It has limited memory so
only keeps track of the id of the last message it saw.
Being able to get only the messages "since" that last id would be
convenient instead of working it's way back until it reaches that id.

Thanks for the reply, and the great app.

jmattheis:
@naimo  Yeah, our android app does something similar. I've created https://github.com/gotify/server/issues/329 to add this functionality.

PR #890, opened 2025-12-09
feat: allow pagination of message by date

### LINKED ISSUE #889: [Feature Request] API Hook for Date/Time Interval Retrieval of Messages
Hi!

I am one of the developers working on [Winify, a graphical UI client for Windows](https://grimore.org/windows/winify) and one of the features that we implement is the retrieval of past notifications between a certain time-interval.

The problem is that there is no direct API hook to do that directly via Gotify. What we do instead is use the ```/application/{Id}/message``` REST path and recursively retrieve all the messages by using the Gotify pagination. After that, we sort all the messages and only then apply the user-requested date-range filter. I am unaware of any correlation between the ```since``` parameter and the actual date when the notification was sent, is there a correlation?

As you can imagine, if the Gotify server has been running for a while, retrieving all the messages or going through all the messages just to find a date-range is a miserable experience for the user, let alone the fact that loading the messages just means more a/deallocations producing memory spikes unnecessarily (this is something that we can optimize on our end, but not the time it takes to retrieve them all).

Would it be possible for you to add a hook for the API that would allow the retrieval of messages from a given application by applying a date-time filter? Right now there is:

```
/application/{Id}/message
```

so maybe something like:
```
/application/{Id}/message/{date}
```

or part of the rest of the other parameters
```
/application/{Id}/message?date=...
```

By implementing this, time to the amount of linear time complexity ```O(a * m)``` where ```a``` would be the number of applications and ```m``` would be the number of messages could just be reduced to ```O(a)```. In case the application ID is known, which in many cases it is because the option in our software is usually provided from the UI to select an application, then the whole ```O(a *m)``` is reduced to just a single operation ```O(1)``` which is an amazing performance boost at no cost.

I have not looked into Gotify myself, so I do not know the internals and I haven't dabbled in Go yet but the actual date when a notification arrives is already recorded by Gotify internally as part of the message structure such that it should not be too costly. We also filter by other criteria when searching, but that criteria is not to be found in your Gotify message structure so that would be more difficult.

Could you implement this REST API path?

Thank you!

eternal-flame-AD:
Fixes #889

TODO: 
[ ]: check the composite index is created correctly on postgres/mysql.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -17,9 +17,9 @@ import (
 
 // The MessageDatabase interface for encapsulating database access.
 type MessageDatabase interface {
-	GetMessagesByApplicationPaginated(appID uint, limit int, since uint64, after uint64, by string) ([]*model.Message, error)
+	GetMessagesByApplicationPaginated(appID uint, limit int, since, after uint64, by string) ([]*model.Message, error)
 	GetApplicationByID(id uint) (*model.Application, error)
-	GetMessagesByUserPaginated(userID uint, limit int, since uint64, after uint64, by string) ([]*model.Message, error)
+	GetMessagesByUserPaginated(userID uint, limit int, since, after uint64, by string) ([]*model.Message, error)
 	DeleteMessageByID(id uint) error
 	GetMessageByID(id uint) (*model.Message, error)
 	DeleteMessagesByUser(userID uint) error
--- a/api/message.go
+++ b/api/message.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/json"
 	"errors"
+	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -16,9 +17,9 @@ import (
 
 // The MessageDatabase interface for encapsulating database access.
 type MessageDatabase interface {
-	GetMessagesByApplicationSince(appID uint, limit int, since uint) ([]*model.Message, error)
+	GetMessagesByApplicationPaginated(appID uint, limit int, since uint64, after uint64, by string) ([]*model.Message, error)
 	GetApplicationByID(id uint) (*model.Application, error)
-	GetMessagesByUserSince(userID uint, limit int, since uint) ([]*model.Message, error)
+	GetMessagesByUserPaginated(userID uint, limit int, since uint64, after uint64, by string) ([]*model.Message, error)
 	DeleteMessageByID(id uint) error
 	GetMessageByID(id uint) (*model.Message, error)
 	DeleteMessagesByUser(userID uint) error
@@ -41,8 +42,10 @@ type MessageAPI struct {
 }
 
 type pagingParams struct {
-	Limit int  `form:"limit" binding:"min=1,max=200"`
-	Since uint `form:"since" binding:"min=0"`
+	Limit int    `form:"limit" binding:"min=1,max=200"`
+	Since uint64 `form:"since" binding:"min=0"`
+	After uint64 `form:"after" binding:"min=0"`
+	By    string `form:"by" binding:"oneof=id date"`
 }
 
 // GetMessages returns all messages from a user.
@@ -69,6 +72,27 @@ type pagingParams struct {
 //	  required: false
 //	  type: integer
 //	  format: int64
+//	- name: by
+//	  in: query
+//	  description: the field to order by
+//	  required: false
+//	  type: string
+//	  enum: [id, date]
+//	  default: id
+//	- name: after
+//	  in: query
+//	  description: return all messages with an cursor value greater than or equal to this value
+//	  minimum: 0
+//	  required: false
+//	  type: integer
+//	  format: int64
+//	- name: since
+//	  in: query
+//	  description: return all messages with an cursor value less than this value
+//	  minimum: 0
+//	  required: false
+//	  type: integer
+//	  format: int64
 //	responses:
 //	  200:
 //	    description: Ok
@@ -90,7 +114,7 @@ func (a *MessageAPI) GetMessages(ctx *gin.Context) {
 	userID := auth.GetUserID(ctx)
 	withPaging(ctx, func(params *pagingParams) {
 		// the +1 is used to check if there are more messages and will be removed on buildWithPaging
-		messages, err := a.DB.GetMessagesByUserSince(userID, params.Limit+1, params.Since)
+		messages, err := a.DB.GetMessagesByUserPaginated(userID, params.Limit+1, params.Since, params.After, params.By)
 		if success := successOrAbort(ctx, 500, err); !success {
 			return
 		}
@@ -120,8 +144,9 @@ func buildWithPaging(ctx *gin.Context, paging *pagingParams, messages []*model.M
 }
 
 func withPaging(ctx *gin.Context, f func(pagingParams *pagingParams)) {
-	params := &pagingParams{Limit: 100}
+	params := &pagingParams{Limit: 100, By: "id"}
 	if err := ctx.MustBindWith(params, binding.Query); err == nil {
+		log.Printf("Paging params: %+v", params)
 		f(params)
 	}
 }
@@ -151,11 +176,25 @@ func withPaging(ctx *gin.Context, f func(pagingParams *pagingParams)) {
 //	  type: integer
 //	- name: since
 //	  in: query
-//	  description: return all messages with an ID less than this value
+//	  description: return all messages with an cursor value less than this value
 //	  minimum: 0
 //	  required: false
 //	  type: integer
 //	  format: int64
+//	- name: after
+//	  in: query
+//	  description: return all messages with an cursor value greater than or equal to this value
+//	  minimum: 0
+//	  required: false
+//	  type: integer
+//	  format: int64
+//	- name: by
+//	  in: query
+//	  description: the field to order by
+//	  required: false
+//	  type: string
+//	  enum: [id, date]
+//	  default: id
 //	responses:
 //	  200:
 //	    description: Ok
@@ -186,7 +225,7 @@ func (a *MessageAPI) GetMessagesWithApplication(ctx *gin.Context) {
 			}
 			if app != nil && app.UserID == auth.GetUserID(ctx) {
 				// the +1 is used to check if there are more messages and will be removed on buildWithPaging
-				messages, err := a.DB.GetMessagesByApplicationSince(id, params.Limit+1, params.Since)
+				messages, err := a.DB.GetMessagesByApplicationPaginated(id, params.Limit+1, params.Since, params.After, params.By)
 				if success := successOrAbort(ctx, 500, err); !success {
 					return
 				}
--- a/api/message.go
+++ b/api/message.go
@@ -3,7 +3,6 @@ package api
 import (
 	"encoding/json"
 	"errors"
-	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -146,7 +145,6 @@ func buildWithPaging(ctx *gin.Context, paging *pagingParams, messages []*model.M
 func withPaging(ctx *gin.Context, f func(pagingParams *pagingParams)) {
 	params := &pagingParams{Limit: 100, By: "id"}
 	if err := ctx.MustBindWith(params, binding.Query); err == nil {
-		log.Printf("Paging params: %+v", params)
 		f(params)
 	}
 }
--- a/database/message.go
+++ b/database/message.go
@@ -39,7 +39,7 @@ func (d *GormDatabase) GetMessagesByUser(userID uint) ([]*model.Message, error) 
 
 // GetMessagesByUserPaginated returns limited messages from a user.
 // If since is 0 it will be ignored.
-func (d *GormDatabase) GetMessagesByUserPaginated(userID uint, limit int, since uint64, after uint64, by string) ([]*model.Message, error) {
+func (d *GormDatabase) GetMessagesByUserPaginated(userID uint, limit int, since, after uint64, by string) ([]*model.Message, error) {
 	var messages []*model.Message
 	db := d.DB.Joins("JOIN applications ON applications.user_id = ?", userID).
 		Where("messages.application_id = applications.id").Order(clause.OrderBy{Columns: []clause.OrderByColumn{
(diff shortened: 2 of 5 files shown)


eternal-flame-AD:
@jmattheis Is this original "since" parameter semantically backwards?  I would imagine "since yesterday" retrieves messages that are dated _after_ yesterday, however the current implementation returns messages that are _before_ yesterday.

To not introduce breaking changes I named the opposite parameter "after" so it's unambiguous .

jmattheis:
See https://github.com/gotify/server/issues/34#issuecomment-644316623 for the naming.

api/message.go
```diff
@@ -41,8 +42,10 @@ type MessageAPI struct {
 }
 
 type pagingParams struct {
-	Limit int  `form:"limit" binding:"min=1,max=200"`
-	Since uint `form:"since" binding:"min=0"`
+	Limit int    `form:"limit" binding:"min=1,max=200"`
+	Since uint64 `form:"since" binding:"min=0"`
```
jmattheis, line 45: All other date-time related fields use RFC3339 format, this would break the consistency with it being a unix timestamp.

I think I'd prefer either something like an ISO interval e.g. `&interval=2025-10-10T23:00Z/2025-10-15T23:00Z` and then allowing to use the after/before id offset, to page in this daterange to prevent the bugs described in https://github.com/gotify/server/issues/889#issuecomment-3633715123

or have separate sinceTime and afterTime fields which accept an RFC3339 datetime.

What do you think? I don't dislike this current solution, so I'm open for any arguments (:.

api/message.go
```diff
@@ -120,8 +144,9 @@ func buildWithPaging(ctx *gin.Context, paging *pagingParams, messages []*model.M
 }
 
 func withPaging(ctx *gin.Context, f func(pagingParams *pagingParams)) {
-	params := &pagingParams{Limit: 100}
+	params := &pagingParams{Limit: 100, By: "id"}
 	if err := ctx.MustBindWith(params, binding.Query); err == nil {
+		log.Printf("Paging params: %+v", params)
```
jmattheis: This is probably not needed right?

database/message.go
```diff
@@ -34,14 +37,33 @@ func (d *GormDatabase) GetMessagesByUser(userID uint) ([]*model.Message, error)
...
 	var messages []*model.Message
 	db := d.DB.Joins("JOIN applications ON applications.user_id = ?", userID).
-		Where("messages.application_id = applications.id").Order("messages.id desc").Limit(limit)
+		Where("messages.application_id = applications.id").Order(clause.OrderBy{Columns: []clause.OrderByColumn{
+		{
+			Column: clause.Column{
+				Table: "messages",
+				Name:  by,
+			},
+			Desc: since != 0 || after == 0,
```
jmattheis, line 51: Does this mean, the messages are returned in another order based on if since or after is set. I feel like this is unexpected if both parameters are set. It should be another parameter to make it more explicit.

Either way, it should always use sorting by id, as this will be more precise than ordering by date, as there could be messages with the same date.

database/message.go
```diff
@@ -60,13 +82,32 @@ func (d *GormDatabase) GetMessagesByApplication(tokenID uint) ([]*model.Message,
...
 
-// GetMessagesByApplicationSince returns limited messages from an application.
+// GetMessagesByApplicationPaginated returns limited messages from an application.
 // If since is 0 it will be ignored.
-func (d *GormDatabase) GetMessagesByApplicationSince(appID uint, limit int, since uint) ([]*model.Message, error) {
+func (d *GormDatabase) GetMessagesByApplicationPaginated(appID uint, limit int, since, after uint64, by string) ([]*model.Message, error) {
 	var messages []*model.Message
-	db := d.DB.Where("application_id = ?", appID).Order("messages.id desc").Limit(limit)
+	db := d.DB.Where("application_id = ?", appID).Order(clause.OrderBy{Columns: []clause.OrderByColumn{
+		{
```
jmattheis, line 90: Maybe this can be unified between the two *Paginated apis, the only difference is the additional filtering for application id or?

jmattheis (changes requested):

api/message.go
GableTheSnail: >or have separate sinceTime and afterTime fields which accept an RFC3339 datetime.

+1 I prefer ISO dates to unix timestamps

database/message.go
eternal-flame-AD: I agree this is not ideal, the reason I used this scheme is mostly:

The computational complexity of finding inequalities on two keys is O(N) time O(N) space  (or you can build 4 indices, regardless the complexity is squared and as if no indices apply), this is made worse by the fact that the database will not be permitted to assume `id` and `date` are monotonic to each other. 

So I am trying to design the API to prevent making it possible to make queries like `id > 100 and timestamp < 200`, to which the only viable query plan will be traverse `idx_id` upwards and collect 50 messages satisfying `timestamp < 200` which may be expensive depending on the selectivity of the `timestamp < 200` to this query.

eternal-flame-AD: I am open to using distinct keys and just enforce they are mutually exclusive to each other though, but the tradeoff is vocabulary bloat instead of this "switch pagination key" semantics.

database/message.go
eternal-flame-AD: Yes .. I don't think I wrote this part (?) so I am not sure the exact rationale but I assume it's to accommodate it being mounted on a different API path. It can be just one interface at the data model layer.

database/message.go
jmattheis: Sorry, I'm currently a little unresponsive, and probably will continue to be for the rest of the month.

---

Ahh yeah, that's tough.

From a general point of view, do we need the new "after" filter. If we want to query a specific date range e.g. 2025-01-01 to 2025-01-31, I don't think it makes much difference if you start fetching the messages from the newest (2025-01-31) or oldest (2025-01-01 messages.

The user would be responsible to fetch as many pages until all messages in the range were fetched. The only thing to consider would be big date ranges, and wanting to lazy load and starting with the oldest message, but I'm not sure if we need to support this.

I'm probably also okay with just not having an index for the features not used in the official clients. I don't think the difference it performance should be too visible, because the query is still pretty simple, and we're probably not talking about billions of messages in gotify.

Another idea would be for an intermediate api to query the message id boundaries. Something like

```
Get /message/idrange?interval=2025-10-10T23:00Z/2025-10-15T23:00Z
{"begin": 42, "end: 1337}
```
We'd only query messages with the actual ids, so it should make the pagination simpler. Maybe we could do this also internally in our pagination api, if a date is provided we translate it to a message. And if both is provided we do a min(translated date id, provided since message id) or similar.

What do you think about any of that?

eternal-flame-AD: Yeah honestly I don't really care about the perf degredation for my own use case -  I am just worried about it becoming a hard to resolve issue once somebody opens a ticket saying they have DB timeout for what looks like a simple query why (had this problem in another project I am engaged in)

This final proposal looks best to me because it inherits existing schemes , but we also need to implement the same thing again in the /application path, probably needs some refactoring, but the API presentations looks okay to me.

@cxtal Do you have any input on this? It should work for your case and respect your preference for long timestamps.

GableTheSnail: hi again @eternal-flame-AD while you were at it, I added some stuff to #889 but I think the point is that my request was more along the lines of "filters" than pagination so there shouldn't be any conflicts with the current way of paginating.

Even if the algorithm is a dumb linear-search when returning data, having the ability to filter the data sent should prevent cases where the application would stall because it is waiting for too much data or that it receives too much data. 

I guess it would be possible to optimize it like having buckets for date-ranges and so on but just the ability to only return matching messages is already a huge performance gain.

RFC-dates are more elegant to work with and also more stable given that Unix time is sometimes measured in milliseconds and other times measured in seconds otherwise it does not matter much :-)

Thank you for adding :-) Looking forward to rip a recursive function apart.

eternal-flame-AD: Thanks. I think I will still go with the pagination, after I come home from this week's vacation. 

If by "filter" you mean "give me the messages on this date", I have to give a hard no on that because time is a complex but continuous measure, the only viable way to implement it on server side is a range query (i.e. pagination). 

The problem is If let's say I live in UTC-8 you live in UTC+2 , we use the same gotify instance to maintain a project together. What is `/message/2025-12-03` supposed to do?

But how you can present it on the frontend is much more freedom, you don't have to literally implement it like a date interval picket: you can just use one date picker, if the user say today you just transform it into the correct range, server returns the first 50 messages, if there are more you can show a "next page" button , it feels natural to me.

GableTheSnail: hi @eternal-flame-AD!

>after I come home from this week's vacation.

Sounds cool, having a good one?

>If by "filter" you mean "give me the messages on this date", I have to give a hard no on that because time is a complex but continuous measure, the only viable way to implement it on server side is a range query (i.e. pagination).

Imagine it like this. Let's say I have a server with 100 messages. The user asks me for messages between two dates. I write a client like Winify that queries the 1000 messages. Because there is no way to filter the responses, I have to pull all the 1000 messages from the server, put them in a collection or array, then I need to sort them by their date field and only then I can offer the result to the user by matching the date-range.

This blows up quickly, let's say on an Android app. Imagine I am across the planet and I query my server back home with a date-range. Imagine the network bloat just to pull those messages and imagine the RAM or CPU my Android device would just waste on sorting useless messages that could have been pre-filled by the server.

The actual way of retrieving these messages would not matter and the current way that you are doing things are just fine using IDs and pagination. When Gotify receives a query like:
```
/application/10/message?limit=5&since=10&dateRange=S,T
```

Gotiy could retrieve all those messages using ```limit``` and ```since``` and only after apply the filter ```dateRange```  with ```S``` being a start date and ```T```.

> The problem is If let's say I live in UTC-8 you live in UTC+2 , we use the same gotify instance to maintain a project together. What is `/message/2025-12-03` supposed to do?

RFC3339 dates that @jmattheis suggested carry the timezone so it should then be:

/message/2025-10-15T23:00Z

where ```Z``` stands for Zulu time which is military for UTC, which is why I chimed in as a preference over timestamps.

With RFC/ISO times, you would know the server timezone so you could even pass your local timezone and Gotify would know what that date represents.

> But how you can present it on the frontend is much more freedom

You're right, pagination is great for UI-oriented frontends but they're not too great for data. Clearly, it would feel awkward to display on the same page 10000000s of results like on a Google search result page. Imagine the scrolling, lol.

Instead pagination is awkward when accessed via programmer's hook because I have to send stuff like this:
```
/application/10/message?limit=5&since=15
/application/10/message?limit=5&since=10
/application/10/message?limit=5&since=5
...
```
till I retrieve all the messages for application ```10```. It would even be cool to be able to retrieve all the messages at once.

When I couple Gotify with, say Amazon Alexa, and I tell Amazon Alexa "Alexa, show all notifications last Tuesday", there is no UI, pages, calendar selector or anything of the sort, not even conceptually.

IoT transforms "last Tuesday" into a date and then a program like Winify will query Gotify by performing calls similar to the above in order to retrieve all the messages from "last Tuesday" and then send a message back to IoT that will make Alexa talk back, perhaps, the notification titles of the messages matching the date "last Tuesday". 

There is no user-selection here, well, except the voice, and Gotify has to be queried programmatically by Winify to process the messages itself.

(By the way, this scenario is real, Winify can do that because it connects with MQTT to brokers for automation.)

eternal-flame-AD: Thanks, I get what you are talking about .. let me think about it  (probably after I come back).. in short I think this debate cannot end unless we literally redesign away from REST principles (which , by construction, enforces some kind of hierarchical ordering of keys that isn't universally beneficial).

eternal-flame-AD: If we want to be ambitious I mean we can have a graphQL endpoint for ultimate freedom on how to lookup messages .. I am open to it but depends on Jannis's appetite for complexity.

The thing is, I totally get your case, if you put this ticket into my own fringe project I will make an API for you specifically. But this is a multi person project where I have to be responsible for a lot of user's API stability and long term scope bloat (like magic query parameters, application specific endpoints, etc)  control. I am not quite convinced any solution proposed here isn't just "hey this is a way to modify API so my specific case of a structural problem is fixed"

GableTheSnail: @eternal-flame-AD Great idea, I mean, I see why you stay close to UI because Gotify indeed has an UI that is used to setup applications but in principle, not all REST endpoints have to be user-endpoints and they can be programmer endpoints and hooks. 

Then, retrieving "all" messages is something an end-user would not do but a programmer that wants to query Gotify would. It's a bit of a cheat because then the API becomes a wrapper around the database but that's the only way it would work because some backends like SQLite are not really concurrent safe so querying the database while Gotify uses it would not work reliably.

Like if you use Sonarr, Radarr, etc, they all have an API token where you can perform various actions that are more data-oriented than what their respective UIs offers.

I was about to write that SQLite has [a pretty powerful FTS engine](https://sqlite.org/fts5.html), so at least for SQLite it would even be easy to offer a "fitler" to search message text itself and it's "free" because this is already built-in to SQLite via plugin.

eternal-flame-AD: I think speaking as a user I never treated Gotify as some kind of data source or sink .. If I lost my database I won't be losing a night's sleep on it. That's why I am as you said "stay close to UI" because I think that's the primary use case.

Whether we should go for expose the entire database as long as it isn't a security vuln kind of thinking .. I can accept the validity but I would say would be a niche use case. If gotify were a paid product, the selling point would be a simple no frills API to get stuff on your phone notification , not how extensible the message query system is. 

About FTS .. yes we could have that but we need to have coverage on MySQL and Postgres as well. I'm sure it's a resolvable problem but let's keep it to a different issue.

GableTheSnail: @eternal-flame-AD hey! back from holiday? Given [the current documentation](https://gotify.net/api-docs#/) does the endpoint used to retrieve messages ```/application/{id}/message``` guarantee that the messages are returned in chronological order? In fact, for any of these endpoints that return messages, like the more general ```/message```, is there any guarantee that the messages will be sorted by date?

I have actually tried this myself and it does seem that the messages are returned in chronological order but I do not know the rest of the Gotify internals to conclude that "chronological order" cannot be changed to, say, "alphabetical order by application", etc.

Similar for returning "all clients" via ```/client```, or "all users" via ```/user``` - just as a question though, what exactly is the sort order and by what criteria is everything sorted when stuff like "clients" or "messages" are retrieved?

I am asking because it might be a good idea to spell out the idea in the documentation. You can do pagination by any criteria, alphabetically, chronologically, etc. It's somewhat related to the ongoing problem because you can flip any table or page depending on the criteria. I just asked for a hook for date-ranges, but I mean, what about endpoints like ```/user``` that return users and you can probably return that structure ordered by any property within that structure.

cheers!

eternal-flame-AD: > I have actually tried this myself and it does seem that the messages are returned in chronological order.

Yeah that is because the primary key is generally self incrementing. In most cases I think you can assume Id and date is monotonic. No guarantee though, if you send a message, change the clock of the server back and send another, ID of the second message will still be increasing.

> you can probably return that structure ordered by any property within that structure.

That's exactly what this PR is already trying to implement, I also agree this is a good pattern in terms of design, but I don't want to overdo it just yet and promise everything can be reordered by every key.

GableTheSnail: > if you send a message, change the clock of the server back and send another, ID of the second message will still be increasing

Thanks. Then I guess that is not an avenue.
I'll wait till the patch is merged.

PR #894, opened 2026-01-01
fix: always use English for TimeAgo

### LINKED ISSUE #893: Why messages timestamps are negative
**Have you read the documentation?**
- [ ] Yes, but it does not include related information regarding my question.

**You are setting up gotify in**
- [ ] Gotify server: kubernetes cluster
- [ ] Gotify client: raspberry pi/debian 13 and macbook/macos 26

**Describe your problem**

I have deployed gotify server in my kubernetes cluster, with a simple helm chart.
Chart is using gotify default configuration, only admin name/pass are updated at first start.

I have created two apps in gotify, one for my raspberry pi, one for my macbook.
I have installed gotify-cli on my my raspberry pi, but not on my macbook.

I have sent messages from raspberry and macbook to gotify-server.
Messages are displayed with negative timestamps:

Image

This is weird, I don't understand why timestamps are displayed like this.

Server, raspberry pi and macbook are all using timesync/ntp and "Europe/Paris" as timezone.

Do someone what is happening here?

**Any errors, logs, or other information that might help us identify your problem**

No log available.

eternal-flame-AD:
Fixes #893 

Also lifted the config out to a global object per the docs for performance.

### DIFF
--- a/ui/src/common/TimeAgoConfig.ts
+++ b/ui/src/common/TimeAgoConfig.ts
@@ -1,7 +1,0 @@
-import {Formatter} from 'react-timeago';
-import {makeIntlFormatter} from 'react-timeago/defaultFormatter';
-
-export const TimeAgoFormatter: Map<'long' | 'narrow', Formatter> = new Map([
-    ['long', makeIntlFormatter({style: 'long', locale: 'en'})],
-    ['narrow', makeIntlFormatter({style: 'narrow', locale: 'en'})],
-]);
--- a/ui/src/typedef/react-timeago.d.ts
+++ b/ui/src/typedef/react-timeago.d.ts
@@ -3,7 +3,7 @@ declare module 'react-timeago' {
 
     export type FormatterOptions = {
         style?: 'long' | 'short' | 'narrow';
-        locale?: void | string;
+        locale?: string;
     };
     export type Formatter = (options: FormatterOptions) => React.ReactNode;
 
--- a/ui/src/common/LastUsedCell.tsx
+++ b/ui/src/common/LastUsedCell.tsx
@@ -1,7 +1,7 @@
 import {Typography} from '@mui/material';
 import React from 'react';
 import TimeAgo from 'react-timeago';
-import {TimeAgoFormatter} from './TimeAgoConfig';
+import {TimeAgoFormatter} from './TimeAgoFormatter';
 
 export const LastUsedCell: React.FC<{lastUsed: string | null}> = ({lastUsed}) => {
     if (lastUsed === null) {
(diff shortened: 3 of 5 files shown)


ui/src/common/TimeAgoConfig.ts
```diff
@@ -0,0 +1,7 @@
+import {Formatter} from 'react-timeago';
+import {makeIntlFormatter} from 'react-timeago/defaultFormatter';
+
+export const TimeAgoFormatter: Map<'long' | 'narrow', Formatter> = new Map([
+    ['long', makeIntlFormatter({style: 'long', locale: 'en'})],
+    ['narrow', makeIntlFormatter({style: 'narrow', locale: 'en'})],
+]);
```
jmattheis: In this case a normal record should be enough. This then can be accessed with TimeAgoFormatter.long
```suggestion
export const TimeAgoFormatter: Record<'long' | 'narrow', Formatter> = {
    long: makeIntlFormatter({style: 'long', locale: 'en'}),
    narrow: makeIntlFormatter({style: 'narrow', locale: 'en'}),
};
```

ui/src/common/TimeAgoConfig.ts
```diff
@@ -0,0 +1,7 @@
+import {Formatter} from 'react-timeago';
+import {makeIntlFormatter} from 'react-timeago/defaultFormatter';
+
+export const TimeAgoFormatter: Map<'long' | 'narrow', Formatter> = new Map([
```
jmattheis: If there is only one export in a file, the file should be named exactly like the export.

ui/src/typedef/react-timeago.d.ts
```diff
@@ -3,6 +3,7 @@ declare module 'react-timeago' {
 
     export type FormatterOptions = {
         style?: 'long' | 'short' | 'narrow';
+        locale?: void | string;
```
jmattheis:
```suggestion
        locale?: string;
```

jmattheis (changes requested):

ui/src/typedef/react-timeago.d.ts
eternal-flame-AD: I copied this verbatim from upstream, but I agree this doesn't have to be this confusing...

https://github.com/nmn/react-timeago/blob/8883dba903c47d6e93e03c4f5efaef60c94ccad4/src/defaultFormatter.js#L17

jmattheis: timeago is using [flow](https://flow.org/en/docs/types/primitives/). This is a typesystem for JavaScript. We use TypeScript, and according to [the docs](https://flow.org/en/docs/types/primitives/) `void` is `undefined` and in typescript this is already covered by `?:`.

Issue #791, opened 2025-04-23
Confirmation Dialog when trying to delete a message


Fratt:
Hello,

First, thanks for Gotify, I love it!

**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I accidentally delete a message in the Gotify web interface. There's no confirmation dialog, so it's very easy to remove messages unintentionally.

**Describe the solution you'd like**
I'd like to see a confirmation dialog appear when a user clicks the delete button on a message. This dialog would ask the user to confirm the deletion action before the message is permanently removed. A simple "Are you sure you want to delete this message?" prompt with "Yes" and "Cancel" buttons would be sufficient.

**Describe alternatives you've considered**
I could write a Tampermonkey script that intercepts the click and do just that, but it would be better if it was part of Gotify.

Thanks in avance!

jmattheis:
Sounds good, we already have an undo functionality in gotify/android added in https://github.com/gotify/android/pull/99. I'd probably prefer an confirmation dialog over the undo feature as it's less error prone, see e.g. https://github.com/gotify/android/issues/382.

The confirmation dialog should have a "remember decision" checkbox that's stored locally on the client which skips the confirmation dialog.

PR #901, opened 2026-01-13
feat: Add undoable message delete with UI-layer snackbar actions and   cleaner store boundaries

AlliotTech:
- Added an undo flow for single-message deletion so users can revert
    accidental deletes within a short window.
  - Moved undo-specific UI behavior (snackbar actions, timers,
    restore) into the Messages page to keep stores UI-agnostic and
    aligned with existing architecture.

image

related issue: https://github.com/gotify/server/issues/791

### DIFF
--- a/ui/src/layout/Header.tsx
+++ b/ui/src/layout/Header.tsx
@@ -9,7 +9,9 @@ import AccountCircle from '@mui/icons-material/AccountCircle';
 import Chat from '@mui/icons-material/Chat';
 import DevicesOther from '@mui/icons-material/DevicesOther';
 import ExitToApp from '@mui/icons-material/ExitToApp';
-import Highlight from '@mui/icons-material/Highlight';
+import Brightness4 from '@mui/icons-material/Brightness4';
+import Brightness7 from '@mui/icons-material/Brightness7';
+import BrightnessAuto from '@mui/icons-material/BrightnessAuto';
 import GitHubIcon from '@mui/icons-material/GitHub';
 import MenuIcon from '@mui/icons-material/Menu';
 import Apps from '@mui/icons-material/Apps';
@@ -17,6 +19,13 @@ import SupervisorAccount from '@mui/icons-material/SupervisorAccount';
 import React, {CSSProperties} from 'react';
 import {Link} from 'react-router-dom';
 import {useMediaQuery} from '@mui/material';
+import {ThemeKey} from './theme';
+
+const themeIcons: Record<ThemeKey, React.ReactElement> = {
+    dark: <Brightness4 />,
+    light: <Brightness7 />,
+    system: <BrightnessAuto />,
+};
 
 const useStyles = makeStyles()((theme: Theme) => ({
     appBar: {
@@ -67,6 +76,7 @@ interface IProps {
     name: string;
     admin: boolean;
     version: string;
+    themeMode: ThemeKey;
     toggleTheme: VoidFunction;
     showSettings: VoidFunction;
     logout: VoidFunction;
@@ -84,9 +94,11 @@ const Header = ({
     style,
     setNavOpen,
     showSettings,
+    themeMode,
 }: IProps) => {
     const {classes} = useStyles();
-
+    const themeLabel = `Toggle theme (current: ${themeMode})`;
+    const themeIcon = themeIcons[themeMode];
     return (
         <AppBar
             sx={{position: {xs: 'sticky', sm: 'fixed'}}}
@@ -117,8 +129,13 @@ const Header = ({
                     />
                 )}
                 <div>
-                    <IconButton onClick={toggleTheme} color="inherit" size="large">
-                        <Highlight />
+                    <IconButton
+                        onClick={toggleTheme}
+                        color="inherit"
+                        size="large"
+                        title={themeLabel}
+                        aria-label={themeLabel}>
+                        {themeIcon}
                     </IconButton>
 
                     <a
--- a/ui/src/layout/Layout.tsx
+++ b/ui/src/layout/Layout.tsx
@@ -1,4 +1,10 @@
-import {createTheme, ThemeProvider, StyledEngineProvider, Theme} from '@mui/material';
+import {
+    createTheme,
+    ThemeProvider,
+    StyledEngineProvider,
+    Theme,
+    useMediaQuery,
+} from '@mui/material';
 import {makeStyles} from 'tss-react/mui';
 import CssBaseline from '@mui/material/CssBaseline';
 import * as React from 'react';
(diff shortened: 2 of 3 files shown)


ui/src/message/Messages.tsx
```diff
@@ -12,6 +12,17 @@ import LoadingSpinner from '../common/LoadingSpinner';
 import {useStores} from '../stores';
 import {Virtuoso} from 'react-virtuoso';
 import {PushMessageDialog} from './PushMessageDialog';
+import {closeSnackbar, SnackbarKey} from 'notistack';
+
+const UndoTimeoutMs = 5000;
```
eternal-flame-AD: Please use onClose or onExited property of notistack instead of using a hardcoded timeout, this won't work if the user mouseover the "undo" button, hesitates for 5+ secs, then click it the request would already gone out.

ui/src/message/Messages.tsx
```diff
@@ -20,22 +31,94 @@ const Messages = observer(() => {
...
-    const deleteMessage = (message: IMessage) => () => messagesStore.removeSingle(message);
+    const deleteMessage = (message: IMessage) => () => startDelete(message);
 
     React.useEffect(() => {
         if (!messagesStore.loaded(appId)) {
             messagesStore.loadMore(appId);
         }
     }, [appId]);
 
+    React.useEffect(() => () => clearPendingDeletes(), []);
```
eternal-flame-AD: Why is this cleanup behavior cancelling pending deletes instead of executing them? Showing a notification saying "<...> (has been) deleted" with an "undo" button and then instead of clicking it, the user switches to another tab, that seems like a confirmation to me.

ui/src/message/Messages.tsx
```diff
@@ -20,22 +31,94 @@ const Messages = observer(() => {
...
     const [pushMessageOpen, setPushMessageOpen] = React.useState(false);
     const [isLoadingMore, setLoadingMore] = React.useState(false);
-    const {messagesStore, appStore} = useStores();
+    const {messagesStore, appStore, snackManager} = useStores();
     const messages = messagesStore.get(appId);
     const hasMore = messagesStore.canLoadMore(appId);
     const name = appStore.getName(appId);
     const hasMessages = messages.length !== 0;
     const expandedState = React.useRef<Record<number, boolean>>({});
+    const pendingDeletesRef = React.useRef<Map<number, PendingDelete>>(new Map());
```
eternal-flame-AD: I think this should be lifted out of the Component, reason is (1) ideally deleting clients and apps should also have this feature (2) it couples the lifetime of the delete transaction, the lifetime of the snackbar, and the lifetime of the component, which leads to difficulties getting the behavior right

ui/src/message/Messages.tsx
```diff
@@ -20,22 +31,94 @@ const Messages = observer(() => {
...
+        messagesStore.markPendingDelete(message.id);
+        const snackKey = snackManager.snack('Message deleted', {
+            action: (key) => (
+                <Button color="inherit" size="small" onClick={() => undoDelete(message.id, key)}>
+                    Undo
+                </Button>
+            ),
+            autoHideDuration: UndoTimeoutMs,
+        });
+        const timeoutId = window.setTimeout(async () => {
```
eternal-flame-AD: Good to have: Since we said the message is "deleted" we should try to fulfill that promise as much as possible. try to use fetch.keepalive option and get the request out even if the page is closed.

AlliotTech: That's a good idea.

ui/src/message/Messages.tsx
AlliotTech: fixed

ui/src/message/Messages.tsx
AlliotTech: The original intention was to give users as many opportunities as possible to reconsider, but you're right—from a UI perspective, it's essential to ensure that the actual behavior matches what the UI displays.

ui/src/message/MessagesStore.ts
```diff
@@ -161,6 +184,27 @@ export class MessagesStore {
         }
     };
 
+    public restoreSingleLocal = (
+        message: IMessage,
+        allIndex: false | number,
+        appIndex: false | number
+    ) => {
+        if (allIndex !== false && this.exists(AllMessages)) {
+            this.state[AllMessages].messages.splice(allIndex, 0, message);
```
eternal-flame-AD, line 193: Good to have: Don't make this assumption here. there is an edge case where if you delete the message, send a new one in and then try to undo the order can get swapped.

No need if it's too complicated

ui/src/message/MessagesStore.ts
```diff
@@ -93,15 +101,30 @@ export class MessagesStore {
...
+                keepalive: true,
+                credentials: 'include',
+            });
+            if (!response.ok) {
+                throw new Error('Failed to delete message');
+            }
+            return;
         }
-        this.snack('Message deleted');
+        await axios.delete(url);
```
eternal-flame-AD, line 117:
```suggestion
        await axios.delete(url, {
            fetchOptions: {
                keepalive: options?.keepalive,
                credentials: 'include',
            },
        });
```

eternal-flame-AD (changes requested):
Are there any specific difficulties making the delete all button also undoable ? A cursory look it seems like it's just deleting with a magic number -1.

ui/src/message/MessagesStore.ts
AlliotTech: it seems like  Axios is still using the default XHR adapter (no adapter configured), so fetchOptions is ignored and keepalive won’t take effect.  👀

ui/src/message/messageDeleteQueue.tsx
```diff
@@ -0,0 +1,139 @@
+import Button from '@mui/material/Button';
```
jmattheis, line 1: I feel like the current implementation is too complex, the storing of indexes seems error prone, as the index changes with each delete and each new message.

I think it should be possible to implement this functionality without modifying the message state for pending deletions and only modify it when there is an actual delete.

I've an idea in mind and will draft something today or later this week.

ui/src/message/MessagesStore.ts
jmattheis: Seems to work with
```
        await axios.delete(config.get('url') + 'message/' + message.id, {
            adapter: 'fetch',
            fetchOptions: {keepalive: true},
        });
```

ui/src/message/messageDeleteQueue.tsx
jmattheis: Could you have a look at #903

Issue #851, opened 2025-09-28
Detect system theme in UI


SuchithSridhar:
**Is your feature request related to a problem? Please describe.**
Right now the Headlight button just changes themes between dark and light - it would be nice to have an option to detect dark theme using the `prefers-color-scheme` option in browsers. This is especially useful in apps like Ferdium where you can't install the dark-reader extension.

**Describe the solution you'd like**
Provide the user an option to use system theme and then detect the state of `prefers-color-scheme`. In my personal opinion, this should be the default option as well.

**Describe alternatives you've considered**
One can use dark-reader but that's a rather clunky work-around.

**Additional context**
None.

jmattheis:
I thought dark is our default. But yeah, I've nothing against using the preferred color scheme for selecting the default scheme in the gotify ui.

PR #902, opened 2026-01-13
feat: implement theme switching with system preference support

AlliotTech:
- Show explicit theme state with system/sun/moon icons in the header
  toggle.
- Add a subtle transition animation when switching theme modes.
- Support “system” theme preference in the layout logic.  

To maintain a consistent user experience, the dark theme is still used as the default.  

related issue: https://github.com/gotify/server/issues/851

Preview: 

image
image
image

### DIFF
--- a/ui/src/layout/Layout.tsx
+++ b/ui/src/layout/Layout.tsx
@@ -57,7 +57,7 @@ const Layout = observer(() => {
     const {classes} = useStyles();
     const [currentTheme, setCurrentTheme] = React.useState<ThemeKey>(() => {
         const stored = window.localStorage.getItem(localStorageThemeKey);
-        return isThemeKey(stored) ? stored : 'dark';
+        return isThemeKey(stored) ? stored : 'system';
     });
     const prefersDark = useMediaQuery('(prefers-color-scheme: dark)');
     const paletteMode = currentTheme === 'system' ? (prefersDark ? 'dark' : 'light') : currentTheme;


ui/src/layout/Layout.tsx
```diff
@@ -68,13 +62,28 @@ const Layout = observer(() => {
         const stored = window.localStorage.getItem(localStorageThemeKey);
         return isThemeKey(stored) ? stored : 'dark';
     });
-    const theme = themeMap[currentTheme];
+    const prefersDark = useMediaQuery('(prefers-color-scheme: dark)');
+    const paletteMode = currentTheme === 'system' ? (prefersDark ? 'dark' : 'light') : currentTheme;
```
eternal-flame-AD, line 63: Why does this start with one of the manual options instead of "system", it looks like you are checking the system settings?

I think we can either 

(1) Don't put "system" as a setting and only use the media query as a fall back: if the user decide to switch the local storage will have a record so the next time they open the app it will respect their choice. If the user doesn't choose there shouldn't be a local storage entry, so you can detect the system theme and use that.

(2) Keep the original behavior just add "system" as a third coice: always start with dark, and only use the system them when the user explicitly sets it to the "system" setting.

 wdyt?

eternal-flame-AD (commented):
Thanks for the PR, see comments please

ui/src/layout/Layout.tsx
AlliotTech: Hi, thanks for your comment, the reason it wasn't set to start from the 'system' is to maintain consistency with the previous default option (dark).  
The current logic should already meet your second condition.

eternal-flame-AD: Thanks, on second thought it looks okay , I just don't want another angry user suddenly come out and say how dare you change my theme .. 

It looks okay on my part .. I will give it some more testing tomorrow and also wait to see if Jannis has additional inputs, thanks!

jmattheis: I prefer system as default, otherwise this feature will probably stay unused for 80% of the usecases because it's not visible that it's even possible to configure. And system is a sane default.

ui/src/layout/Header.tsx
```diff
@@ -60,13 +62,30 @@ const useStyles = makeStyles()((theme: Theme) => ({
...
+        opacity: 1,
+    },
+    themeIconDark: {
+        transform: 'rotate(-20deg) scale(0.95)',
+        opacity: 0.9,
+    },
+    themeIconLight: {
+        transform: 'rotate(20deg) scale(1.05)',
+        opacity: 1,
+    },
```
jmattheis: Can we remove all additional custom css + the span in the icon button. I'd rather keep it simple and it IMO doesn't make much difference.

ui/src/layout/Header.tsx
```diff
@@ -60,13 +62,30 @@ const useStyles = makeStyles()((theme: Theme) => ({
...
 }));
 
 interface IProps {
     loggedIn: boolean;
     name: string;
     admin: boolean;
     version: string;
+    themeMode: 'dark' | 'light' | 'system';
```
jmattheis: Reuse the `ThemeKey`-type here.

ui/src/layout/Header.tsx
```diff
@@ -84,8 +103,24 @@ const Header = ({
     style,
     setNavOpen,
     showSettings,
+    themeMode,
 }: IProps) => {
     const {classes} = useStyles();
+    const themeLabel = `Toggle theme (current: ${themeMode})`;
+    const themeIcon =
```
jmattheis: Similar to how you've done it in Layout, use a Record instead of nested shortif
```
const icons: Record<ThemeKey, React.ReactElement> = {
    dark: <Brightness4/>,
    light: <Brightness7/>,
    system: <BrightnessAuto/>,
}
```

ui/src/layout/Header.tsx
```diff
@@ -117,8 +152,15 @@ const Header = ({
                     />
                 )}
                 <div>
-                    <IconButton onClick={toggleTheme} color="inherit" size="large">
-                        <Highlight />
+                    <IconButton
```
jmattheis, line 132: I think this is good enough for now, but this probably should be on a separate config page with a select with named options to make it more clear what the actual functionality is.

ui/src/layout/Header.tsx
AlliotTech: fixed

ui/src/layout/Header.tsx
AlliotTech: fixed

ui/src/layout/Layout.tsx
AlliotTech: Haha, it's hard to please everyone, so initially, like @eternal-flame-AD  thought, we avoided changing the default behavior for existing users and stuck with the previous dark mode.  But it's okay, the default has now been changed to `system`.

PR #903, opened 2026-01-14
feat: add delete message undo

jmattheis:


eternal-flame-AD:
Should we implement this for delete all messages as well? That seems like a much easier (and higher impact) mistake to make?

ui/src/message/Messages.tsx
```diff
@@ -70,6 +92,8 @@ const Messages = observer(() => {
     };
 
     const renderMessages = () => (
+        // Virtuoso logs errors when elements have 0px size as this normally means it's miscalculated.
```
eternal-flame-AD: I think that error meant when that element is needed for scrollbar positioning it will not work.

If you send more than a page of messages, scroll to the bottom and start clicking delete the scrollbar has a lot of "jumping" glitch/error that weren't there on master branch.

Reproduction: 

1. shell script send 100 messages, call them "message $i"
2. scroll to the bottom
3. click the delete button on the message and the end of the viewport on master you should be clicking "1, 2, 3, 4, ..." sequentially, on this branch it is very random

eternal-flame-AD: I think just .filter() should be fine here no need to render an empty component, I am pretty sure Virtuoso is O(n) anyways.

jmattheis: Good catch, the filter is the better solution, and we already have caching for "get messages" as we merge the messages with the application images, so this should work quite nicely.

Could you test this again, I had some weird occurrences when deleting a message, closing the tab before the undo timer was done, and then reopen the tab. But cannot reproduce it now, maybe it was some weird caching in dev mode.

> Should we implement this for delete all messages as well? That seems like a much easier (and higher impact) mistake to make?

We have a confirmation dialog there already, I'd say this is enough.

eternal-flame-AD: > Could you test this again, I had some weird occurrences when deleting a message, closing the tab before the undo timer was done, and then reopen the tab

I gave it some more testing it seems fine to me. I've seen some repeated firing of DELETE on the same ID (seems to be onExited and the clean process firing at the same time), but since deleting a single message is idempotent I don't think this is related to messages _not_ being deleted/

off topic: now that node also has native fetch support maybe the axios abstraction is no longer needed?

jmattheis: Yeah, agree. Created https://github.com/gotify/server/issues/904

PR #906, opened 2026-01-21
fix: MobX V6 actions

eternal-flame-AD:
Gets rid of warnings like:

```
[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: CurrentUser@6.loggedIn
```

I don't find any behavioral breakage but wrapping them in actions should be more conformant.

doc: https://mobx.js.org/actions.html#asynchronous-actions

### DIFF
--- a/ui/src/message/MessagesStore.ts
+++ b/ui/src/message/MessagesStore.ts
@@ -1,5 +1,5 @@
 import {BaseStore} from '../common/BaseStore';
-import {action, IObservableArray, observable, reaction, makeObservable, runInAction} from 'mobx';
+import {action, IObservableArray, observable, reaction, runInAction} from 'mobx';
 import axios, {AxiosResponse} from 'axios';
 import * as config from '../config';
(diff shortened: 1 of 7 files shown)


ui/src/message/MessagesStore.ts
```diff
@@ -170,6 +137,7 @@ export class MessagesStore {
 
     public exists = (id: number) => this.stateOf(id).loaded;
 
+    @action
```
jmattheis, line 170: The other actions are defined in the constructor via makeObservable. The stores should either use this or the decorator syntax and not mix between them.

eternal-flame-AD: Ah yes .. I think let's just migrate to the new syntax, iirc I did this because the traditional makeObserver can't monkeypatch a private method into an action so these two have to be written like this. Decorator should not have this visibility problem

PR #909, opened 2026-01-24
Allow sorting applications

### LINKED ISSUE #358: Reorganize the list of "Apps"

jmattheis:
Supercedes #856 
See #804
Fixes #358

image

### DIFF
--- a/model/application.go
+++ b/model/application.go
@@ -20,7 +20,7 @@ type Application struct {
 	// required: true
 	// example: AWH0wZ5r0Mbac.r
 	Token  string `gorm:"type:varchar(180);uniqueIndex:uix_applications_token" json:"token"`
-	UserID uint   `gorm:"index" json:"-"`
+	UserID uint   `gorm:"index;uniqueIndex:uix_application_user_id_sort_key,priority:1" json:"-"`
 	// The application name. This is how the application should be displayed to the user.
 	//
 	// required: true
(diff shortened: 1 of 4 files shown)


model/application.go
```diff
@@ -20,7 +20,7 @@ type Application struct {
 	// required: true
 	// example: AWH0wZ5r0Mbac.r
 	Token  string `gorm:"type:varchar(180);uniqueIndex:uix_applications_token" json:"token"`
-	UserID uint   `gorm:"index" json:"-"`
+	UserID uint   `gorm:"index;uniqueIndex:idx_id_sortkey,priority:1" json:"-"`
```
eternal-flame-AD: maybe use the existing name convention (uix_application_user_id_sortkey)

jmattheis: I'll change this to `uix_application_user_id_sort_key`.

matezoltanfarkas:
yaaay, thank you guys your are the best @jmattheis @eternal-flame-AD !

truthsword:
So 2.9.0? Maybe...

jmattheis:
@truthsword yes soon, probably this week

PR #911, opened 2026-02-06
enhance(backend): allow disabling HTTP port completely

### LINKED ISSUE #910: HTTPS only mode (disable HTTP Permanently)
**Have you read the documentation?**
- [x] Yes, but it does not include related information regarding my question.
- [ ] Yes, but the steps described in the documentation do not work on my machine.
- [ ] Yes, but I am having difficulty understanding it and want clarification.

**You are setting up gotify in**
- [ ] Docker
- [x] Linux native platform
- [ ] Windows native platform

**Describe your problem**

I want to disable plain HTTP completely and only use HTTPS only but don't know how i can disable it completely
Can someone help me a bit

**Any errors, logs, or other information that might help us identify your problem**

Ex: `docker-compose.yml`, `nginx.conf`, android logcat, browser requests, etc.

<details><summary>Name of the information here</summary><br><pre>

contents here

</pre></details>

eternal-flame-AD:
Resolves #910

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -22,13 +22,23 @@ func Run(router http.Handler, conf *config.Configuration) error {
 	shutdown := make(chan error)
 	go doShutdownOnSignal(shutdown)
 
-	httpListener, err := startListening("plain connection", conf.Server.ListenAddr, conf.Server.Port, conf.Server.KeepAlivePeriodSeconds)
-	if err != nil {
-		return err
-	}
-	defer httpListener.Close()
-
 	s := &http.Server{Handler: router}
+
+	hasListener := false
+
+	if conf.Server.Port >= 0 {
+		httpListener, err := startListening("plain connection", conf.Server.ListenAddr, conf.Server.Port, conf.Server.KeepAlivePeriodSeconds)
+		if err != nil {
+			return err
+		}
+		hasListener = true
+		defer httpListener.Close()
+		go func() {
+			err := s.Serve(httpListener)
+			doShutdown(shutdown, err)
+		}()
+	}
+
 	if conf.Server.SSL.Enabled {
 		if conf.Server.SSL.LetsEncrypt.Enabled {
 			applyLetsEncrypt(s, conf)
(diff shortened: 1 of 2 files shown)


runner/runner.go
```diff
@@ -22,13 +22,23 @@ func Run(router http.Handler, conf *config.Configuration) error {
...
 	go doShutdownOnSignal(shutdown)
 
-	httpListener, err := startListening("plain connection", conf.Server.ListenAddr, conf.Server.Port, conf.Server.KeepAlivePeriodSeconds)
-	if err != nil {
-		return err
+	s := &http.Server{Handler: router}
+
+	hasListener := false
+
+	if conf.Server.Port >= 0 {
```
jmattheis, line 29: Could this instead be a `enabled` setting similar to the one for TLS? Should be default true.

eternal-flame-AD: I guess so, but it would be called GOTIFY_SERVER_ENABLED=false which semantically sounds funny.

IzumiSenaSora: `GOTIFY_SERVER_ENABLED=false` it really looks confusing... Some may even think that "huh? My server disabled? :p"
But we can use like
`GOTIFY_PLAIN_SERVER_ENABLED=false`
Or
`GOTIFY_HTTP_SERVER_ENABLED=false`

eternal-flame-AD: What I am trying to say is the name of these keys are tied to the yaml hierarchy by the configor, so if we add a key side by side with the current "port" key it would be called GOTIFY_SERVER_SOMETHING, which I don't like.

But I guess we can use GOTIFY_SERVER_DISABLE_HTTP or GOTIFY_SERVER_HTTP_FUNC=on/off/redirect , both should be good options forward

jmattheis: I'm okay with GOTIFY_SERVER_DISABLE_HTTP but I'm questioning if we really need this functionality or can prevent the problems via a different solution. See https://github.com/gotify/server/issues/910#issuecomment-3864372652

eternal-flame-AD: If it is acceptable for this particular ticket I won't object (I guess in that case you would also set port to 0 just to get out of the way for potential conflicts). 

If more people want this maybe a real disable is more reliable though, we can defer this until then.

jmattheis: Okay, then let's wait for more feedback.

PR #917, opened 2026-02-12
chore: refactor out axios

### LINKED ISSUE #904: Migrate from axios to fetch in UI
We should use the browser defaults and use fetch instead of axios in the UI and the UI tests.

eternal-flame-AD:
Fixes #904


ui/src/application/AppStore.ts
```diff
@@ -89,11 +115,11 @@ export class AppStore extends BaseStore<IApplication> {
...
     ): Promise<void> => {
-        await axios.post(`${config.get('url')}application`, {
-            name,
-            description,
-            defaultPriority,
-        });
+        await this.currentUser.authenticatedFetch(
+            config.get('url') + 'application',
+            jsonBody({name, description, defaultPriority}),
+            jsonTransform
```
jmattheis, line 121: Could you explain the abstraction for the Response transformer? Versus directly calling .then(resp => resp.json()) where needed? I think I'd rather be more aligned with the api from fetch vs doing the custom.

ui/src/CurrentUser.ts
```diff
@@ -67,17 +110,17 @@ export class CurrentUser {
...
-                url: config.get('url') + 'client',
-                method: 'POST',
-                data: {name},
-                headers: {Authorization: 'Basic ' + btoa(username + ':' + password)},
-            })
-            .then((resp: AxiosResponse<IClient>) => {
+        const fetchInit = jsonBody({name});
+        fetchInit.headers = new Headers(fetchInit.headers);
+        fetchInit.headers.set('Authorization', 'Basic ' + btoa(username + ':' + password));
+        return this.authenticatedFetch(
```
jmattheis, line 116: Can this use fetch directly, it's weird if we use authenticatedFetch to authenticate.

ui/src/application/AppStore.ts
```diff
@@ -78,7 +100,11 @@ export class AppStore extends BaseStore<IApplication> {
         IApplication,
         'id' | 'name' | 'description' | 'defaultPriority' | 'sortKey'
     >): Promise<void> => {
-        await axios.put(`${config.get('url')}application/${id}`, app);
+        await this.currentUser.authenticatedFetch(
+            config.get('url') + 'application/' + id,
+            {...jsonBody(app), method: 'PUT'},
```
jmattheis, line 105: The method could accept a second parameter which overrides stuff defined in json body. E.g.
```
export const jsonBody = (body: unknown, {headers, ...rest}: RequestInit = {}) => ({
    method: 'POST',
    headers: {
        'Content-Type': 'application/json',
        ...headers,
    },
    body: JSON.stringify(body),
    ...rest,
});
```

```
            jsonBody(app, {method: 'PUT'}),
```

ui/src/message/MessagesStore.ts
```diff
@@ -83,12 +83,28 @@ export class MessagesStore {
...
+                .authenticatedFetch(
+                    config.get('url') + 'application/' + appId + '/message',
+                    {
+                        method: 'DELETE',
+                    },
+                    identityTransform
+                )
+                .then(() =>
+                    this.snack(`Deleted all messages from ${this.appStore.getByID(appId).name}`)
+                );
```
jmattheis, line 107: What's the reason for moving the snack into the then? This will be shorter outside.
```
            await this.currentUser.authenticatedFetch(
                config.get('url') + 'application/' + appId + '/message',
                {method: 'DELETE'},
                identityTransform
            );
            this.snack(`Deleted all messages from ${this.appStore.getByID(appId).name}`);
```

ui/src/message/MessagesStore.ts
```diff
@@ -147,10 +169,12 @@ export class MessagesStore {
...
         };
 
-        await axios.post(`${config.get('url')}message`, payload, {
-            headers: {'X-Gotify-Key': app.token},
-        });
-        this.snack(`Message sent to ${app.name}`);
+        const fetchInit = jsonBody(payload);
+        fetchInit.headers = new Headers(fetchInit.headers);
+        fetchInit.headers.set('X-Gotify-Key', app.token);
+        await this.currentUser
```
jmattheis, line 175: I'd say this should use fetch directly or use the jsonBody second parameter mentioned in a previous comment:
```
        await this.currentUser.authenticatedFetch(
            config.get('url') + 'message',
            jsonBody(payload, {headers: {'X-Gotify-Key': app.token}}),
            jsonTransform
        );
        this.snack(`Message sent to ${app.name}`);
```

ui/src/tests/setup.ts
```diff
@@ -37,7 +35,7 @@ export const newTest = async (pluginsDir = ''): Promise<GotifyTest> => {
     const gotifyInstance = startGotify(gotifyFile, port, pluginsDir);
 
     const gotifyURL = 'http://localhost:' + port;
-    await waitForGotify('http-get://localhost:' + port);
+    await waitForGotify(gotifyURL);
```
jmattheis, line 38: Next time please create a separate PR for this.

ui/src/CurrentUser.ts
```diff
@@ -92,59 +135,58 @@ export class CurrentUser {
...
-            .then((resp: AxiosResponse<IClient[]>) => {
-                resp.data
-                    .filter((client) => client.token === this.tokenCache)
-                    .forEach((client) => axios.delete(config.get('url') + 'client/' + client.id));
+        await this.authenticatedFetch(config.get('url') + 'client', {}, jsonTransform<IClient[]>)
+            .then((resp) => {
+                resp.filter((client) => client.token === this.tokenCache).forEach((client) =>
+                    this.authenticatedFetch(
+                        config.get('url') + 'client/' + client.id,
+                        {},
```
jmattheis, line 186: This should have method: 'DELETE'. Please check the other, and always make this explicit, even if it's not needed for method: "GET".

ui/src/fetchUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const jsonBody: (body: any) => RequestInit = (body: any) => ({
+    method: 'POST',
+    headers: {
+        'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(body),
+});
+
+export const yamlBody: (text: string) => RequestInit = (text: string) => ({
```
jmattheis, line 19: Either define the type on the variable / constant, or in the lambda, but not both.

```suggestion
export const yamlBody = (text: string): RequestInit => ({
```

or

```suggestion
export const yamlBody: (text: string) => RequestInit = (text) => ({
```

I prefer the first one.

ui/src/fetchUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const identityTransform: ResponseTransformer<Response> = (response: Response) =>
+    Promise.resolve(response);
+
+export const jsonTransform = <T>(response: Response): Promise<T> => response.json();
+
+export const textTransform = (response: Response): Promise<string> => response.text();
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const jsonBody: (body: any) => RequestInit = (body: any) => ({
```
jmattheis, line 11: Can we call them postJsonBody, or something that makes POST more obvious?

jmattheis, line 11: Using `unknown` instead of `any` removes the need for disabling the linting.

jmattheis (changes requested):

ui/src/tests/setup.ts
eternal-flame-AD: There is a common dep on axios from wait-on so I figured it's good to also take this one out so axios no longer appear in the SBOM.

PR #923, opened 2026-02-21
Sort key fixes

### LINKED ISSUE #920: Error 1170 (42000): BLOB/TEXT column 'sort_key' used in key specification without a key length

- [x] Docker
- [ ] Linux machine
- [ ] Windows machine

```
services:
gotify:
image: gotify/server:latest
container_name: gotify
ports:
- 8082:80
environment:
- GOTIFY_DEFAULTUSER_NAME=user
- GOTIFY_DEFAULTUSER_PASS=password
- GOTIFY_DATABASE_DIALECT=mysql
- GOTIFY_DATABASE_CONNECTION=gotify:password@tcp(nsys-mysql:3306)/gotify?charset=utf8mb4&parseTime=True&loc=Local
- TZ=Europe/Rome
volumes:
- "./gotify-data:/app/data"
networks:
- nsys-net
restart: always
networks:
nsys-net:
external: true
volumes:
gotify-data:
```
</details>

- [ ] None
- [ ] Nginx
- [ ] Apache
- [x] Caddy

```

```
</details>

- [ ] WebUI
- [ ] gotify-cli
- [ ] Android Client 
- [x] Docker

**What did you do?**
Updated from 2.8.0 to 2.9.0

Starting Gotify version 2.9.0@2026-02-13-15:29:00

2026/02/16 12:02:08 /src/gotify/database/database.go:83 Error 1170 (42000): BLOB/TEXT column 'sort_key' used in key specification without a key length
[0.960ms] [rows:0] CREATE UNIQUE INDEX `uix_application_user_id_sort_key` ON `applications`(`user_id`,`sort_key`)
panic: Error 1170 (42000): BLOB/TEXT column 'sort_key' used in key specification without a key length

goroutine 1 [running]:
main.main()
        /src/gotify/app.go:44 +0x33f

jmattheis:
This PR fixes the sort_key index creation reported in #920 and sorting in general, as it was broken for MySQL and Postgres. ORDER BY isn't consistently implemented by Sqlite, Postgres and Mysql. For Postgres and Mysql it's case insensitive which breaks fractional sorting.

Fixes #920

### DIFF
--- a/model/application.go
+++ b/model/application.go
@@ -58,5 +58,5 @@ type Application struct {
 	//
 	// required: true
 	// example: a1
-	SortKey string `gorm:"uniqueIndex:uix_application_user_id_sort_key,priority:2,length:255" form:"sortKey" query:"sortKey" json:"sortKey"`
+	SortKey string `gorm:"type:bytes;uniqueIndex:uix_application_user_id_sort_key,priority:2,length:255" form:"sortKey" query:"sortKey" json:"sortKey"`
 }


database/database.go
```diff
@@ -108,10 +108,10 @@ func fillMissingSortKeys(db *gorm.DB) error {
 	}
 
 	var apps []*model.Application
-	if err := db.Order("user_id, sort_key, id ASC").Find(&apps).Error; err != nil && err != gorm.ErrRecordNotFound {
```
eternal-flame-AD, line 111: Does this not work on MySQL ?

jmattheis: It should work, but given that there is a check before that the sort_key is null, it shouldn't be needed, and as the sort_key sorting can be inconsistent between database engines, I'd rather remove this.

database/application.go
```diff
@@ -40,10 +42,14 @@ func (d *GormDatabase) CreateApplication(application *model.Application) error {
 	return d.DB.Transaction(func(tx *gorm.DB) error {
 		if application.SortKey == "" {
```
jmattheis, line 41: > That doesn't sound right, I tried on postgres and it seems to be case sensitive, am I missing something?

Your example seems sorted case insensitive. Here a more explicit example:
```
psql (18.1, server 18.2 (Debian 18.2-1.pgdg13+1))
Type "help" for help.

gotify=# create table test (sort_key text not null); insert into test values ('a0V'), ('a0l'), ('a1'), ('Zz'), ('a0A');
CREATE TABLE
INSERT 0 5
gotify=# select * from test order by sort_key asc;
 sort_key
----------
 a0A
 a0l
 a0V
 a1
 Zz
(5 rows)
```
`a0l` is between `a0A` and `a0V`, but should actually be sorted like this:
```
$ python
Python 3.14.2 (main, Jan  2 2026, 14:27:39) [GCC 15.2.1 20251112] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = ['a0V', 'a0l', 'a1', 'a0A', 'a2', 'Zz']
>>> a.sort()
>>> a
['Zz', 'a0A', 'a0V', 'a0l', 'a1', 'a2']
```

(using docker.io/postgres:18)

eternal-flame-AD: Huhh, my example did sort A0V < A0l, your output is the opposite, .. maybe it's some collation or global options? I am also on Postgres 18 but I'm using a system install, I will try if I can reproduce it on docker.

It's probably collation options again ( utf8ci is case insensitive) , can we just store these as []byte instead (so the type is no longer text)?

jmattheis: Nice idea, this works.

PR #927, opened 2026-03-05
enhance: LABEL docker builds

### LINKED ISSUE #924: add OCI labels to Docker image
It would be helpful if OpenContainer Initiative (OCI) labels (like org.opencontainers.image.source) were added to the Docker image. Tools like Renovate use these labels to provide changelogs with updates.

Add OCI labels to the gotify/server Docker image. The following specifically would be most helpful:
- org.opencontainers.image.source
- org.opencontainers.image.version
- org.opencontainers.image.created

For Renovate I can manually provide a changelog URL, but it's not as embedded as using an image label

**Additional context**
[OCI Pre-defined labels](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys)
[Renovate docs](https://docs.renovatebot.com/modules/datasource/docker/#description)

eternal-flame-AD:
Fixes #924


jmattheis (approved):

docker/Dockerfile
```diff
@@ -54,6 +54,9 @@ RUN cd /src/gotify && \
 
 FROM debian:${DEBIAN}
 
+LABEL org.opencontainers.image.documentation https://gotify.net/
```
jmattheis: This is the deprecated format could you use the new format?
```
 2 warnings found (use docker --debug to expand):
 - LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 57)
 - LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 58)
```

PR #939, opened 2026-03-28
enhance: bump token keyspace to >128 bits

### LINKED ISSUE #936: Feature Request: Add GOTIFY_TOKEN_LENGTH configuration option
When Gotify generates tokens (for applications, clients, etc.), they are always 15 characters long. This length is hardcoded and cannot be changed either through the UI or any configuration option. For environments with stricter security requirements, 15 characters may not be sufficient, and there is currently no way for an administrator to enforce a longer, more secure token length.

A new application configuration option `GOTIFY_TOKEN_LENGTH` should be introduced to allow administrators to configure the length of generated tokens. The option should:

- Default to `15` if not provided, preserving the current behavior and ensuring backward compatibility.
- Accept any integer value greater than or equal to `15` to ensure a minimum level of security.
- Be documented alongside the other existing configuration options.

Example configuration:
```yml
server:
  [...]
  tokenlength: 15
```
Or via environment variable:
```bash
GOTIFY_TOKEN_LENGTH=32
```

Manually editing tokens after generation: Users could theoretically update tokens manually in the database, but this is error-prone, not user-friendly, and not a scalable solution for administrators.

Enforcing a fixed longer token length: Hardcoding a longer token length (e.g., 32 characters) instead of 15 would improve security but would not give administrators the flexibility to define their own required length based on their specific security policies.

**Additional context**
Allowing configurable token lengths is a common best practice in security-conscious applications, giving administrators the ability to align token generation with their organization's security policies or compliance requirements (e.g., NIST, OWASP recommendations). This change is fully backward compatible since the default value remains `15`.

eternal-flame-AD:
Previous keyspace: 2^84.3131 (2.40e25) , we are moving to 2^132.5 keyspace (7.65e+39)

Closes #936 

The original length is indeed on the legacy end of the spectrum and slightly behind OWASP recommendations for a password without MFA , therefore I think it's good to bump it to >128 bits to push it well into cryptographic secret strength that TLS/SSH accepts and collision resistance on SHA256 is targeted at. Maybe this can also simplify some workflows if some users want's to built ad hoc E2EE systems. 

Regardng the proposal in #936 I don't see the need to add a token length option because the "ceiling" of security (TLS) is clear and achievable by simply bumping the constant. Creating a knob can lead to downstream problems like tokens too long causing them to not fit in certain webhook forms, etc. 

The increase in character count is modest, so I don't expect any issues with URL or webhook form length not fitting, but after the increase we can confidently say at least regarding the brute force attack vector TLS would be the weakest link by every metric.

### DIFF
--- a/auth/token.go
+++ b/auth/token.go
@@ -7,7 +7,7 @@ import (
 
 var (
 	tokenCharacters   = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-_")
-	randomTokenLength = 14
+	randomTokenLength = 22 // ~2^132.5 keyspace (7.65e+39)
 	applicationPrefix = "A"
 	clientPrefix      = "C"
 	pluginPrefix      = "P"


jmattheis (approved):
👍

Issue #433, opened 2021-09-17
Option to disable authentication


tpansino:
**Have you read the documentation?**
- [x] Yes, but it does not include related information regarding my question.
- [ ] Yes, but the steps described in the documentation do not work on my machine.
- [ ] Yes, but I am having difficulty understanding it and wants clarification.

**You are setting up gotify in**
- [x] Docker
- [ ] Linux native platform
- [ ] Windows native platform

**Describe your problem**
Related to #203, #20.

I've noticed that authentication seems to be a hot topic for self-hostable services, like this one. There are many different standards people want supported (OIDC, SAML, AD, LDAP), all of which can be difficult to implement correctly and support. Libraries similar to [passport.js](https://www.passportjs.org/) can help, but generally have a learning curve to integrating them.

Looking through the Issues, another mode I have not seen suggested yet is to simply allow disabling of Authentication entirely. This has the benefits of:
- No additional libraries or knowledge necessary for the developer to implement
- Users can bring their own auth, typically in the form of a proxy like [Oauth2-Proxy](https://oauth2-proxy.github.io/oauth2-proxy/), or Traefik's [ForwardAuth](https://doc.traefik.io/traefik/v2.3/middlewares/forwardauth/) feature paired with an external Identity Provider like [KeyCloak](https://www.keycloak.org/) or [Authelia](https://github.com/authelia/authelia)
- Users can choose to have no auth and only secure their systems behind a firewall instead (great for experimentation and local dev)

Given the reluctance on #20. and the ongoing discussion about an Auth plugin system, this seemed worth mentioning. I've seen projects avoid Auth plugins and ONLY offer built-in Auth or none, I've also seen projects that offer support for every method under the sun. Just depends on what the devs feel comfortable supporting.

I think most folks _would_ like to see support for an SSO strategy, whatever that is. I personally use OIDC, but mostly via Auth proxies since not a lot of projects have native OIDC support, which I think is fair.

jmattheis:
Hey, just disabling auth isn't enough because gotify needs to know which user should be used. Sure, it should be possible to add a setting like preauthenticed_user to the config, which auto logs on an existing user, but this feels a little hacky.

Generally I'm open to natively support another auth system, but as you say it's hard to choose (:.

tpansino:
Ok, I've spent slightly more than 2 minutes playing around with Gotify now. 😅  When I wrote the above I didn't realize there were multiple users, that does complicate things.

[...]

PR #941, opened 2026-03-29
OIDC Server

jmattheis:
This MR:
* Refactors the auth handling by replacing the old `requireToken` / callback structure with a simpler approach.
* Adds cookie-based token reading and `/auth/local/{login,logout}` endpoints for web UI authentication.
  * Without this, we'd have to redirect to an url `https://gotify-ui.net/#/oidc/callback?token=...` to store the token in the local storage. Using cookies seems to be more easy than securing this against CSRF and stuff. Also with HttpOnly cookies we guard against pontential token exposures.
  * Secure cookies are supported but disabled by default to avoid breaking existing HTTP installations. We could detect HTTPS via the `Referer` or `X-Forwarded-Proto` headers in the future.
* Add /gotifyinfo endpoint. In the android app we need to know if oidc is enabled, and the android app also displays the version. I didn't want to add this information to the /version endpoint so it's a new endpoint. I don't really like the naming, but I wanted something unique so that the android app can be sure it's a gotify/server instance.

### Web UI: Standard OAuth2 Authorization Code + PKCE

This is the standard OIDC implementation with a library.

* A "Login with OIDC" button appears when OIDC is enabled **(This is only visible in the production build, when the UI is served by gotify/server)**
* Clicking it redirects to `GET /auth/oidc/login`, which then redirects to the IdP.
* After authentication the IdP redirects to `/auth/oidc/callback`, which:
  1. Exchanges the authorization code for tokens.
  2. Resolves the user via the userinfo endpoint (creating the user if configured).
  3. Creates a Gotify client, sets the session cookie, and redirects to the UI.

### Android App flow

Android's recommended [App Links](https://developer.android.com/training/app-links/create-deeplinks) (verified domains) aren't usable here because the server is self-hosted at a user-chosen domain.

I guess most of the android app, are doing their own OIDC flow against the IdP server to then obtain a token from the IdP server to then access the resources with the token. In our case, we need a client token from gotify.

We could let the android app do the whole OIDC flow, but this would require us to have a public oauth2 client, as we don't want to store the client_secret in the android app. Requesting the client secret from the server seems like a bad idea.

Another simple implementation would be to do the same flow as the Web-UI, but in the end redirecting to a custom URI scheme like gotify://oidc/callback?token=C123123123 and then using this token for authentication. I've decided against this because the custom URI scheme can be set by any app, and therefore intercept the token.

PKCE was exactly made to prevent this, so I've searched for a way to still get the benefits from PKCE, but not having to reimplement a OIDC like handshake.

This flow is non standard but it's basically stolen from Vaultwarden. I'm not that deep in security, but I'd expect they spend some time designing this, and ensuring it's secure.

It works like this.

1. The app generates a PKCE `code_verifier` / `code_challenge` pair.
2. The app `POST /auth/oidc/external/authorize` with `code_challenge`, a custom-scheme `redirect_uri`, and a client name. The server stores a pending session keyed by a random `state` token, and then reponds with the the authorization URL + `state`. [Vaultwarden 1](https://github.com/dani-garcia/vaultwarden/blob/3a1378f469c66dc4cd92a32add3c47594c40604f/src/sso.rs#L206), [Vaultwarden 2](https://github.com/dani-garcia/vaultwarden/blob/3a1378f469c66dc4cd92a32add3c47594c40604f/src/sso_client.rs#L116),
3. The app opens the authorization URL in a browser. The IdP authenticates the user and redirects to the app's custom URI (gotify://oidc/callback) with `code` and `state`.
4. `POST /auth/oidc/external/token` with `code`, `state`, and `code_verifier`. The server gets the pending session, relays the `code_verifier` to the IdP for PKCE validation, exchanges the code for tokens, resolves the user, and returns a Gotify client token + user info. [Vaultwarden 1](https://github.com/dani-garcia/vaultwarden/blob/3a1378f469c66dc4cd92a32add3c47594c40604f/src/sso.rs#L241), [Vaultwarden 2](https://github.com/dani-garcia/vaultwarden/blob/3a1378f469c66dc4cd92a32add3c47594c40604f/src/sso_client.rs#L163)

PKCE is required, and only verified by the IdP. We could support non PKCE IdP by manually validating the challange on the /auth/oidc/external/token endpoint, simliar to [Vaultwarden Impl](https://github.com/dani-garcia/vaultwarden/blob/3a1378f469c66dc4cd92a32add3c47594c40604f/src/sso_client.rs#L171), but I think most of the Idp will support PKCE, so it's required until someone requests it.

For testing I've added Dex and Authelia config in [./test/oidc](https://github.com/gotify/server/tree/oidc/test/oidc).

Related MRs:
- https://github.com/gotify/website/pull/100
- https://github.com/gotify/android/pull/444

I've manually tested this with:
- Web-UI: Dex (on root, and subpath /gotify/ with a reverse proxy) and authelia
- Android with Dex and Authelia

Disclaimer: this PR was created with AI assistance.


api/health.go
```diff
@@ -16,7 +16,7 @@ type HealthAPI struct {
 }
```
jmattheis, line 16: @eternal-flame-AD sorry this is kinda big, but would you mind having a look at this? Don't worry about declining if you don't want to review this, or don't have the time. I don't want to burden you with this :D.

Anyway, my plan is to create a snapshot docker build in the coming days and then let the people in #433 try out the functionality to improve the user-guide and find possible bugs with special IdP servers. So there is no rush reviewing this.

eternal-flame-AD: Sure, I can look at this tonight first to see if there are large architectural/security changes needed and we can do a final round for final check and nitpicks after the code settles. Probably won't review Android though because I'm not very familar with android dev.

api/oidc.go
```diff
@@ -0,0 +1,362 @@
...
+	SecureCookie       bool
+	AutoRegister       bool
+	pendingSessions    map[string]*pendingOIDCSession
+	pendingSessionsMu  sync.Mutex
+}
+
+func (a *OIDCAPI) storePendingSession(state string, session *pendingOIDCSession) {
+	a.pendingSessionsMu.Lock()
+	defer a.pendingSessionsMu.Unlock()
+	for s, sess := range a.pendingSessions {
```
eternal-flame-AD: I feel this could be done better, this has O(n^2) complexity over number of users who are authenticating at the same time.

If we are unconditionally iterating through the entire map as the first action of the flow anyways, might as well just use an array (with swap-delete).

A more complex solution is bulk cleanup like this: https://gist.github.com/eternal-flame-AD/e4a47a8a7e6b26bba365a7d24ab84221

api/oidc.go
```diff
@@ -0,0 +1,362 @@
...
+//	    schema:
+//	        $ref: "#/definitions/Error"
+func (a *OIDCAPI) CallbackHandler() gin.HandlerFunc {
+	callback := func(w http.ResponseWriter, r *http.Request, tokens *oidc.Tokens[*oidc.IDTokenClaims], state string, provider rp.RelyingParty, info *oidc.UserInfo) {
+		user, status, err := a.resolveUser(info)
+		if err != nil {
+			http.Error(w, err.Error(), status)
+			return
+		}
+		clientName, _, _ := strings.Cut(state, ":")
```
eternal-flame-AD: Can the name itself already have a colon in it?

api/oidc.go
```diff
@@ -0,0 +1,362 @@
...
+//	    description: Error
+//	    schema:
+//	        $ref: "#/definitions/Error"
+func (a *OIDCAPI) ExternalAuthorizeHandler(ctx *gin.Context) {
+	var req model.OIDCExternalAuthorizeRequest
+	if err := ctx.ShouldBindJSON(&req); err != nil {
+		ctx.AbortWithError(http.StatusBadRequest, err)
+		return
+	}
+	state, err := a.generateState(req.Name)
```
eternal-flame-AD: Is this how it works for VaultWarden (server generates state) ? I'm not super sure what this is for. From my cursory exploration it seems like in VaultWarden state is simply for the client's convenience (might be wrong..)

docs/spec.json
```diff
@@ -958,13 +1105,33 @@
         }
       }
     },
+    "/gotifyinfo": {
```
eternal-flame-AD, line 1108: I guess if we want to be standard we can use /.well-known/appspecific which is IANA registered 

https://github.com/Vroo/well-known-uri-appspecific/blob/main/well-known-uri-for-application-specific-purposes.txt

config/config.go
```diff
@@ -55,6 +56,16 @@ type Configuration struct {
 	UploadedImagesDir string `default:"data/images"`
 	PluginsDir        string `default:"data/plugins"`
 	Registration      bool   `default:"false"`
+	OIDC              struct {
+		Enabled       bool   `default:"false"`
+		Issuer        string `default:""`
+		ClientID      string `default:""`
+		ClientSecret  string `default:""`
```
eternal-flame-AD, line 63: I think it would be great if this can be read from a file as well for things like docker secrets

auth/cookie.go
```diff
@@ -0,0 +1,22 @@
+package auth
+
+import (
+	"net/http"
+)
+
+// CookieMaxAge is the lifetime of the session cookie in seconds (7 days).
+const CookieMaxAge = 7 * 24 * 60 * 60
```
eternal-flame-AD, line 8: Should we actually enforce this (delete or invalidate clients an expiration date in DB)? 

I feel like we should otherwise I'm not very certain of the benefits of adding a lifetime. It's also useful for things like orgs who may want to automatically deprovision users after they leave.

api/oidc.go
eternal-flame-AD: Okay this seems to be documented in RFC6749, I got it messed up. The flow is right although the  bookkeeping ClientName could be clearer (don't keep them both in the server memory and in the token text at the same time).

auth/cookie.go
jmattheis: I'd say there is still a benefit, I'd say if a user hasn't used the cookie in a longer time, it's better to just discard it, even if it's still valid.

I'll add a "source" string field to the client table which can be "user, local-login, oidc-login". And then periodically prune clients with type *-login that didn't had activity in the last 7 days.

Do you think 7 days is a good enough default for now?

config/config.go
jmattheis: Yeah, I agree, but let's do this in a separate MR.

I'm thinking about dropping the YAML support and just support env vars. The current config library isn't really great for array values about which we have many issues.

Besides env vars, the user could specify a env file. Similar to this https://screego.net/#/config

docs/spec.json
jmattheis: I'm a little worried if users have badly configured reverse proxies and e.g. serve certbot/acme stuff directly under .well-known instead of .well-known/acme-challenge/ which then could block gotify/android from working.

What do you think?

api/oidc.go
jmattheis: No, this is different from vaultwarden, but as I understood is this for CSRF protection, so basically the android app has to check the initial state against the state that will be received on the redirect uri. So I don't think it matters.

Do you mean removing the client name from the pending session struct and then directly reading it from the state similar to how it's done in the web ui flow?

api/oidc.go
jmattheis: True, I'll reverse the order of the random nonce and the client name and then split by the first colon as the nonce is hex encoded and never includes a colon.

api/oidc.go
jmattheis: My plan was to have a simple implementation as we probably don't have many parallel logins anyway.

Your decay map doesn't seem to complex, I'll use it, as it's not that much code overhead.

api/health.go
jmattheis: Awesome thanks!

api/oidc.go
eternal-flame-AD: > Do you mean removing the client name from the pending session struct and then directly reading it from the state similar to how it's done in the web ui flow?

Yes, actually I would prefer the inverse, I think either of these are better:

- remove the "plaintext" name from the state as it's only a CSRF token but store it in the pending session struct as we are already doing in L72), OR
- make the content of state actually semantically strong (JSON, optionally signed, etc)

docs/spec.json
eternal-flame-AD: That makes sense, it's up to you :D, I was just commenting this because you said you feel like this is a little out of place, but I don't see /gotifyinfo being bad.

auth/cookie.go
eternal-flame-AD: That seems okay, I would prefer if the cookie can be auto-renewed if the user visits say within 3 days of the expiration, since the current behavior is not really real expiration and even if we don't use it for actual security, the amount of client entry churn that user must manually delete can be annoying. Not hard blocker though.

api/oidc.go
eternal-flame-AD: See below, but for a simple fix I would say just base64 URL safe encode it.

auth/cookie.go
jmattheis: The expiration time is already updated in the same frequency as the last used token timestamp -> https://github.com/gotify/server/blob/oidc/auth/authentication.go#L152

api/oidc.go
jmattheis: Currently, the pending sessions struct is only used with the mobile flow. The webui flow, doesn't need it, as it internally already stores the state in cookies.

To make it consistent, we coulfd fill the pending session also for the web flow and then only store the client name in the pending session struct. This would duplicate some validations that are already done in the standard webui oidc flow. But I don't think it's much of a problem. I'll change it to be consistent between the flows.

api/oidc.go
jmattheis: Regarding the implementation, of the decay map. I'll change it so that Get also removes the item from the maps, as it should only be possible to do the token exchange once.

Generally, replaying isn't possible because the IdP also validates this, and only allows the token exchange once. But it seems cleaner to enforce this in the gotify implementation.

jmattheis: Additionally, it seems to not check expiry on Get, I feel like this should be enforced.

eternal-flame-AD: If we want the 10 minute expiration to be exact we have to double check (like currently where we store the exact value in the value field) anyways because the map has a granularity of 10 minutes. There are more checks you can do inside decaymap (specifically, verify whether each half of the structure is unexpired), but will only guarantee the age of all inserted keys are between 10-20 minutes, should be okay from a business logic perspective (OIDC flow expiration isn't a hard security parameter) but an exact solution still require a double check. (I can edit the gist to add this check if you want to just lmk

I think the simplest solution is just double check the CreatedAt field like you are already doing in the current implementation and maybe change a read-only Get to a Pop instead for the atomicity.

jmattheis:
Stuff I changed:
- Removed the client name from the state
- Used the decay map (Splitted into two commits, my modifications are separate)
- Added a client source to the api / ui

For inactive client deleting I've created #943, it's probably a bigger change, when all models get a createdAt field. So I didn't want to pollute this PR.

jmattheis:
@eternal-flame-AD Can you recheck the review fixes? I merge this then, and do the step up auth in a separate MR to not have a big MR with all changes in it.

eternal-flame-AD:
Sure, will take a look tomorrow , sorry have been busy

eternal-flame-AD:
Wait why do we still need this source field? Seems a little arbitrary .. If there is a planned behavior that depends on this field we should have a fixed list of permitted values, if not I would say just remove it or use a more generic "comment" field.

lgtm otherwise, I'm not going to do a functional test as I'm unfamiliar with the technology so likely can't give advice better than it works or it doesn't. Your plan of putting it on docker for the proponents to test it seems like a good next step.

jmattheis:
> Wait why do we still need this source field? Seems a little arbitrary .. If there is a planned behavior that depends on this field we should have a fixed list of permitted values, if not I would say just remove it or use a more generic "comment" field.

I intended it to be used for the step up auth. When doing the session elevation, we need to know if we should present oidc or password authentication to the user. This could be based on the client source.

We could let the user decide for session elevation which authentication method to use, as it doesn't really matter for the gotify side if the same authentication was used for creating the client.

I think I'd prefer letting the user chose on session elevation, as this makes the logic simpler. What do you think?

eternal-flame-AD:
> I intended it to be used for the step up auth. When doing the session elevation, we need to know if we should present oidc or password authentication to the user. This could be based on the client source.

I would say let the user decide and don't worry about it. You can even reuse the login UI component.

Memorizing the original auth method feels to be achieving very little to me at least in term of security, might as well give the user the option to use either at the time of use. If you meant it as a convenience (so user don't have to select the auth method at the step up prompt) I don't think it's worth the complexity, and we still must have a list of permitted value instead of a freestyle field.

I can't think of a scenario where you would want to have a user to have client A (that must be stepped up with the _user_ password) but client B which must be stepped up with OIDC instead, especially considering all clients are equal. 

If the goal is to prevent the user from switching authentication methods: just mark the user with a login_method field or bitmask seems to be most straightforward and standard to me. (I.e. each user can be userpass only, OIDC only, oe both)

jmattheis:
Okay, removed the client source commit(s).

eternal-flame-AD (approved):
Lgtm

PR #951, opened 2026-04-16
enhance(security): add body size and pluginConf/image entity size limits

### LINKED ISSUE #950: Limit body size for all requests
We should limit the body size for all requests. It's currently possible to send big messages to /message or big yaml configs to /plugin/{id}/config.

I guess a good default could be something like 10MB, but it should be configurable.

<details><summary>Email Report</summary>
<p>

**Summary**

Gotify Server's plugin configuration update feature accepts
attacker-controlled YAML bodies at ` POST /plugin/{id}/config`.
Confirmed in ` v2.9.1`, when ` registration=true` and at least one
installed plugin supports ` Configurer`, an unauthenticated remote
attacker can self-register and submit an oversized configuration body to
trigger severe memory pressure and persistent configuration storage
growth.

**Details**

`UpdateConfig` verifies plugin ownership and ` Configurer` support, but
then immediately calls ` io.ReadAll(ctx.Request.Body)` with no size
limit. The full request body is buffered before YAML parsing or
plugin-side validation, so any reachable caller can force memory
allocation proportional to the submitted body size. After validation,
the handler stores the original bytes in ` conf.Config`, which turns the
same endpoint into a persistence amplifier rather than only a transient
parse-time sink.

``` elementToProof
conf, err := c.DB.GetPluginConfByID(id)
...
if aborted := supportOrAbort(ctx, instance, compat.Configurer); aborted { return }
newconfBytes, err := io.ReadAll(ctx.Request.Body)
...
if err := yaml.Unmarshal(newconfBytes, newConf); err != nil { ... }
if err := instance.ValidateAndSetConfig(newConf); err != nil { ... }
conf.Config = newconfBytes
```

When ` registration=true`, ` POST /user` is reachable under
` authentication.Optional()` and allows unauthenticated non-admin
account creation. User creation triggers ` fireUserAdded`, which
initializes per-user instances for all installed plugins, and
` RequireClient` accepts Basic Auth directly. The attacker can then call
` GET /plugin` to enumerate a plugin instance whose ` capabilities`
include ` configurer`, fetch the current YAML from
` GET /plugin/{id}/config`, and reuse that ` id` against
` POST /plugin/{id}/config`.

Because the handler stores raw YAML bytes rather than a normalized
representation, the attacker can append large YAML comment blocks to an
otherwise valid configuration. ` yaml.Unmarshal` ignores comments, but
the oversized original document is still persisted as
` PluginConf.Config`. I verified the unbounded body read and raw-byte
persistence in ` api/plugin.go`, and the self-registration plus Basic
Auth reachability in ` api/user.go`, ` router/router.go`,
` auth/authentication.go`, and ` plugin/manager.go`. I did not identify
a fixed version from the current materials.

**PoC**

1.  Ensure the target runs Gotify Server ` v2.9.1` with
    ` registration=true` and at least one installed plugin whose
    ` /plugin` entry lists ` configurer` in ` capabilities`.

2.  Create a normal account:

``` elementToProof
POST /user

{"name":"pocuser","pass":"PocPassw0rd!"}
```

3.  Use Basic Auth for that account to call ` GET /plugin`, select an
    entry whose ` capabilities` array contains ` configurer`, then fetch
    its current YAML from ` GET /plugin/{id}/config`.

4.  Append a large number of YAML comment lines to the returned document
    and submit it back:

``` elementToProof
POST /plugin/{id}/config
Authorization: Basic <base64(pocuser:PocPassw0rd!)>
Content-Type: application/x-yaml

<original valid YAML>
# padding 000001
# padding 000002
...
```

5.  Observe high memory pressure during the request. If the YAML remains
    otherwise valid, a subsequent ` GET /plugin/{id}/config` returns the
    enlarged document, confirming that the oversized raw body was
    persisted.

**Impact**

Observed impact is a conditional unauthenticated network denial of
service in deployments that allow self-registration and have at least
one installed ` Configurer` plugin: a remote attacker can consume
process memory with a single oversized request and can repeatedly
enlarge persisted plugin configuration data. Where registration is
disabled, the same bug remains reachable to a low-privilege
authenticated user. I did not confirm confidentiality or integrity
impact from current evidence.

</p>
</details>

eternal-flame-AD:
This fixes #950 .

Added:

- [X] Limit for blob and files (32MB)
- [X] Limit for all request body content combined (64MB)

Do we need:

- [ ] Limit for total applications/clients? 
- [ ] Make the limits configurable?

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -3,7 +3,6 @@ package api
 import (
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -11,6 +10,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/gotify/server/v2/auth"
+	"github.com/gotify/server/v2/database"
 	"github.com/gotify/server/v2/model"
 	"github.com/h2non/filetype"
 	"gorm.io/gorm"
@@ -18,7 +18,7 @@ import (
 
 // The ApplicationDatabase interface for encapsulating database access.
 type ApplicationDatabase interface {
-	CreateApplication(application *model.Application) error
+	CreateApplication(application *model.Application, quota uint32) error
 	GetApplicationByToken(token string) (*model.Application, error)
 	GetApplicationByID(id uint) (*model.Application, error)
 	GetApplicationsByUser(userID uint) ([]*model.Application, error)
@@ -30,6 +30,7 @@ type ApplicationDatabase interface {
 type ApplicationAPI struct {
 	DB       ApplicationDatabase
 	ImageDir string
+	Quota    uint32
 }
 
 // Application Params Model
@@ -55,6 +56,10 @@ type ApplicationParams struct {
 	//
 	// example: a1
 	SortKey string `form:"sortKey" query:"sortKey" json:"sortKey"`
+}
+
+func (p *ApplicationParams) EffectiveSize() uint64 {
+	return uint64(len(p.Name)) + uint64(len(p.Description)) + uint64(len(p.SortKey))
 }
 
 // CreateApplication creates an application and returns the access token.
@@ -90,9 +95,17 @@ type ApplicationParams struct {
 //	    description: Forbidden
 //	    schema:
 //	        $ref: "#/definitions/Error"
+//	  422:
+//	    description: Unprocessable Entity
+//	    schema:
+//	        $ref: "#/definitions/Error"
 func (a *ApplicationAPI) CreateApplication(ctx *gin.Context) {
 	applicationParams := ApplicationParams{}
 	if err := ctx.Bind(&applicationParams); err == nil {
+		if applicationParams.EffectiveSize() > MaxApplicationClientEntrySize {
+			ctx.AbortWithError(http.StatusUnprocessableEntity, fmt.Errorf("application entry too large (max: %d bytes)", MaxApplicationClientEntrySize))
+			return
+		}
 		app := model.Application{
 			Name:            applicationParams.Name,
 			Description:     applicationParams.Description,
@@ -103,7 +116,7 @@ func (a *ApplicationAPI) CreateApplication(ctx *gin.Context) {
 			Internal:        false,
 		}
 
-		if err := a.DB.CreateApplication(&app); err != nil {
+		if err := a.DB.CreateApplication(&app, a.Quota); err != nil {
 			handleApplicationError(ctx, err)
 			return
 		}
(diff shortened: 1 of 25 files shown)


api/application.go
jmattheis, line 349: Instead of manually calling ParseMultiPartForm it's possible to handle the error below, E.g

```go
file, err := ctx.FormFile("file")
if err == http.ErrMissingFile {
	ctx.AbortWithError(400, errors.New("file with key 'file' must be present"))
	return
} else if _, ok := err.(*http.MaxBytesError); ok {
	ctx.AbortWithError(http.StatusRequestEntityTooLarge, fmt.Errorf("file too large (max: %d bytes)", MaxUploadSize))
	return
} else if err != nil {
	ctx.AbortWithError(http.StatusBadRequest, err)
	return
}
```
This makes it shorter and easier to understand IMO, but I'm okay with the current solution, as it's the recommendation from their docs.

api/limits.go
```diff
@@ -0,0 +1,8 @@
+package api
+
+const (
+	// Maximum upload size of 32MB for blobs and files.
+	MaxUploadSize = 32 << 20
```
jmattheis, line 5: Why have the different limits?

api/application.go
```diff
@@ -3,22 +3,22 @@ package api
 import (
```
jmattheis, line 3: I think I rather not have the quota feature for now. I think most gotify instances are shared with trusted users and not open to all.

If you think this feature is worth it, then I'm okay with including it, but it would need UI support (remaining quota) and messages need to be included too.

eternal-flame-AD: I think my stance is as as expressed in the original issue, we should resolve the threat not resolve the checklist.

If we are committing to the threat model of Gotify being used between relatively trusted users (who won't DOS the server by filling it up with data), which I actually agree with, then we should not bother with either. I suggest we close the report as risk accepted and maybe add documentation if needed. (After all, a body size limit is standard feature for all common reverse proxies).

If we decide we should bother about insider threats DOSing the server, then we need systematic measure because body size limit alone won't fix any exposure with the client and application loophole wide open.

jmattheis: Okay, then lets close this a s risk accepted. 

Do you know of any other open source project that has such a risk documentation? I've looked a little bit but found nothing.

eternal-flame-AD: Security.md is a good place to get started, we can be specific about our threat model, and for this particular point we can declare things like resource exhaustion attacks by authenticated users as out of scope.

eternal-flame-AD:
I will close this and cherry pick limits that may be good for convienience later.

PR #952, opened 2026-04-18
Client elevation

### LINKED ISSUE #944: Step-Up auth
Actions including
* deleting clients,applications
* creating/deleting/modifying users
* changing the password of the current user

Should require step-up auth. Meaning the user has to re authenticate with
* their user password if it's a local user
* the IdP server, if it's a oidc user

This is to ensure that potentially unrecoverable actions are a more secured. 

We probably have to add another field to the user to define the last used login method, so we can use this to determine how to re authenticate.

Todos:
* Local user step-up: Add /auth/step-up api which accepts a client-id, and step-up duration e.g. (15m). This endpoint should require basic auth.
* Oidc step up
  * Add step up boolean flag to  /auth/oidc/login endpoint, and store it in pending session
  * On /auth/oidc/callback we see the stepup flag and can read the existing session cookie and then do the step up. We'll have to additionally validate that the user of the session and the oidc callback matches.
  * Redirect the user back to the previously open page.

jmattheis:
[step-up.webm](https://github.com/user-attachments/assets/31967ccd-fce5-442a-8163-e38d4f6a6746)

Fixes #944


eternal-flame-AD:
I am traveling rn, will take a look probably Tuesday!

ui/src/client/ClientStore.ts
```diff
@@ -38,4 +38,11 @@ export class ClientStore extends BaseStore<IClient> {
         await this.createNoNotifcation(name);
         this.snack('Client added');
     };
+
+    @action
+    public elevate = async (id: number, durationSeconds: number): Promise<void> => {
+        await axios.post(`${config.get('url')}client:elevate`, {id, durationSeconds});
+        await this.refresh();
+        this.snack('Client elevated');
```
eternal-flame-AD: ux: when the selection is "cancel elevation" (if durationSeconds <0) it should show a different msg

auth/authentication.go
```diff
@@ -39,16 +42,20 @@ type Auth struct {
...
 
 // RequireClient returns a gin middleware which requires a client token or basic authentication header to be supplied
 // with the request.
 func (a *Auth) RequireClient(ctx *gin.Context) {
-	a.evaluateOr401(ctx, a.user(false), a.client(false))
+	a.evaluateOr401(ctx, a.user(false), a.client(false, false))
+}
+
+// RequireElevatedClient requires an elevated client token or basic auth.
+func (a *Auth) RequireElevatedClient(ctx *gin.Context) {
```
eternal-flame-AD, line 57: nitpick: the helper function in auth package is getting more and more numerous and with cryptic boolean flag, maybe consider use some pattern (functional, etc) to compose things like "RequireAdmin", "RequireClient", "RequireElevatedClient", etc

auth/authentication.go
```diff
@@ -39,16 +42,20 @@ type Auth struct {
 	SecureCookie bool
 }
 
-// RequireAdmin returns a gin middleware which requires a client token or basic authentication header to be supplied
-// with the request. Also the authenticated user must be an administrator.
+// RequireElevatedAdmin requires an elevated client token or basic auth, the user must be an admin.
```
eternal-flame-AD: The function name does not match doc name

router/router.go
```diff
@@ -185,64 +186,52 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
...
-
-		clientAuth.POST("current/user/password", userHandler.ChangePassword)
-
 		clientAuth.POST("/auth/logout", sessionHandler.Logout)
 	}
 
+	clientElevated := g.Group("")
+	{
+		clientElevated.Use(authentication.RequireElevatedClient)
+		clientElevated.POST("/client:elevate", clientHandler.ElevateClient)
```
eternal-flame-AD: Maybe /client/:id/elevate is more REST-y?

api/user.go
```diff
@@ -126,7 +130,17 @@ func (a *UserAPI) GetCurrentUser(ctx *gin.Context) {
...
-	ctx.JSON(200, toExternalUser(user))
+	result := &model.CurrentUserExternal{
+		ID:    user.ID,
+		Name:  user.Name,
+		Admin: user.Admin,
+	}
+	client := auth.GetClient(ctx)
+	if client != nil {
+		result.ClientID = client.ID
+		result.ElevatedUntil = client.ElevatedUntil
```
eternal-flame-AD: I think it would make more sense and be more useful to hide this field if the server believes it is in the past. It gives a concreate yes or no for the client whether they are elevated right now irrespective of time differences.

ui/src/client/Clients.tsx
```diff
@@ -113,6 +134,20 @@ const Row = ({name, value, lastUsed, fEdit, fDelete}: IRowProps) => (
         <TableCell>
             <LastUsedCell lastUsed={lastUsed} />
         </TableCell>
+        <TableCell>
+            {elevatedUntil && Date.parse(elevatedUntil) > Date.now() ? (
+                <TimeAgo date={elevatedUntil} formatter={TimeAgoFormatter.long} />
```
eternal-flame-AD: Same concern as above, I would cap the granularity of this field to one minute because the "in 15 seconds" or "5 seconds ago" are false precision that may not be true depending on clock synchronization.

seems to be doable by setting a custom unit for react-timeago between minutes, hours and days

router/router.go
jmattheis: I've got this from here: https://google.aip.dev/136 though I'm not sure how common this actually is. I don't really have a preference, do you have one?

PS: Thanks for the review!

eternal-flame-AD: It's okay, I prefer using REST as that matches as the current "vibe" of designs, but I'm not going to insist on it.

eternal-flame-AD (dismissed):

router/router.go
jmattheis: I've changed this to /client/:id/elevate, you're right, this is more consistent with other apis e.g. /application/:id/image.

PR #959, opened 2026-05-09
Delete inactive clients/sessions

### LINKED ISSUE #943: Periodically delete inactive client sessions
From https://github.com/gotify/server/pull/941#discussion_r3015537823

With the changes in #941, we have a session cookie for both local login and OIDC login. The cookie expires after 7 days of inactivity. gotify should periodically delete clients of type session when there weren't used in the last 7 days.

LastUsed from the client is only set on use, so it may be nil when the client token was never used. We probably have to add a createdAt field to the client (and to be consistent to all other models). So it's possible check `createdAt < now - 7 days && (lastUsed is null or lastUsed < now - 7 days)` for the delete query.

jmattheis:
This MR includes:

* Add createdAt fields to all models
* Add expiresAfterInactivitySeconds to client
* Clean up inactive clients

fixes #943


router/router.go
```diff
@@ -71,7 +72,16 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 		for range ticker.C {
 			connectedTokens := streamHandler.CollectConnectedClientTokens()
 			now := time.Now()
-			db.UpdateClientTokensLastUsed(connectedTokens, &now)
+			if err := db.UpdateClientTokensLastUsed(connectedTokens, &now); err != nil {
+				log.Println("Error updating last used", err)
+			}
+			if expired, err := db.CleanupExpiredClients(now); err == nil {
```
eternal-flame-AD, line 78: I think as with last time we shouldn't have significant unusable precision on the schema.

If we can only process this every minute let use expiresAfter...Minutes or hours instead

ui/src/client/Clients.tsx
```diff
@@ -141,6 +162,16 @@ const Row = ({name, value, lastUsed, elevatedUntil, fEdit, fDelete, fElevate}: I
                 '-'
             )}
         </TableCell>
+        <TableCell>
+            <TimeAgo date={createdAt} formatter={TimeAgoFormatter.long} />
+        </TableCell>
+        <TableCell className="expires-in">
+            {expiresAt ? (
+                <TimeAgo date={expiresAt} formatter={TimeAgoFormatter.longMinutes} />
```
eternal-flame-AD: Same thing, this format should be unified as minutes or hours, minutes.

I think the ability to glance through this column is more important than the exact timestamp represented in a changing formatting.

eternal-flame-AD:
Another option for this to work precisely is to memoize an expiresAt field updated along with lastUsed, and predicate all client selections with a condition.

ui/src/client/Clients.tsx
jmattheis: Do you mean that the units for all rows are the same, so only hours and minutes are displayed even if the client expires in days or months?

eternal-flame-AD: Yes, like all X hours, Y minutes or some other format that makes sense for scanning and glancing I think.

router/router.go
jmattheis: I'll use your proposal with the memoized expires_at, then we can keep the second precision.

ui/src/client/Clients.tsx
jmattheis: It now looks like this. I've changed created at to absolute time and elevation/expiry to the short relative time.

image

eternal-flame-AD (approved):
lgtm!

PR #960, opened 2026-05-12
fix(ui): make messages layout responsive

### LINKED ISSUE #744: Request to add option to fill page or make reactive
I have a few notifications that are long and wide and I have to scroll to get to the horizontal scroll bar to see the wider notifications.
image

It would be nice to have the notifications window fill the available area to allow for viewing the wider notifications either with an option to "fill" or have the notification area be reactive to the size of the screen or window viewing them. 

not using Gotify as there are others that offer this but aren't as easy to configure if there is no native Gotify support.

**Additional context**
none at the moment as the above should suffice, but let me know if you have questions.

Victor-root:
## ✨ Summary

This improves the responsiveness of the WebUI Messages page.

The Messages page was previously constrained by the default `DefaultPage` width, which made the message list feel unnecessarily narrow on desktop screens and left a lot of available horizontal space unused.

This PR keeps the change scoped to the Messages page and preserves the existing layout for the other pages.

Fixes #744

## 🖥️ Desktop behavior

On larger desktop screens, the Messages page now uses more of the available horizontal space while still keeping a reasonable maximum width.

This makes long notifications, logs, Markdown content, and code blocks much easier to read without changing the overall Gotify layout.

## 🔧 Changes

- Set a wider `maxWidth` only for the Messages page.
- Size individual message cards based on their content.
- Keep short text-only messages compact.
- Allow messages with wide content, such as logs or code blocks, to expand up to the available page width.
- Clamp `<pre>` blocks to the message card width while preserving internal horizontal scrolling.
- Leave the other pages unchanged.

### ✅ Tested

I tested the change locally on:

- 🖥️ Desktop
- 💻 Laptop
- 📱 Mobile web viewport

Everything works as expected:
- the Messages page uses much more available width on desktop;
- the layout remains clean on laptop/tablet-sized screens;
- mobile web keeps message cards within the viewport while preserving code block horizontal scrolling;
- the Android app is not affected;
- existing messages, apps, clients, and WebSocket updates continue to work.

### Desktop :
image

### Laptop :

laptop

### Mobile web :

image

### DIFF
--- a/ui/src/message/Message.tsx
+++ b/ui/src/message/Message.tsx
@@ -80,6 +80,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
         },
         '& pre': {
             overflow: 'auto',
+            maxWidth: '100%',
             borderRadius: '0.25em',
             backgroundColor:
                 theme.palette.mode === 'dark' ? 'rgba(255,255,255,0.05)' : 'rgba(0,0,0,0.05)',
@@ -168,6 +169,10 @@ const Message = ({
                 style={{
                     display: 'flex',
                     flexWrap: 'wrap',
+                    width: 'fit-content',
+                    minWidth: 'min(700px, 100%)',
+                    maxWidth: '100%',
+                    margin: '0 auto',
                     borderLeftColor: priorityColor(priority),
                     borderLeftWidth: 6,
                     borderLeftStyle: 'solid',
--- a/ui/src/message/Messages.tsx
+++ b/ui/src/message/Messages.tsx
@@ -115,6 +115,7 @@ const Messages = observer(() => {
     return (
         <DefaultPage
             title={name}
+            maxWidth={1600}
             rightControl={
                 <div>
                     {app && (


ui/src/message/Message.tsx
```diff
@@ -88,6 +89,13 @@ const useStyles = makeStyles()((theme: Theme) => ({
         '& img': {
             maxWidth: '100%',
         },
+        [theme.breakpoints.down('sm')]: {
+            '& pre': {
+                whiteSpace: 'pre-wrap',
```
jmattheis: pre must continue to have _no_ word wrap. It's basically the feature of code formatting to keep it as it. I'd say almost every markdown renderer doesn't do this. E.g. 

```
REALLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLY LOOOOOOOOOOOOOOOOOOOOOOOOOOOOONG LINNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNE
```

ui/src/message/Messages.tsx
```diff
@@ -115,6 +115,7 @@ const Messages = observer(() => {
     return (
         <DefaultPage
             title={name}
+            maxWidth={1600}
```
jmattheis, line 118: While this works for wide messages, shorter, text only messages now look bad and aren't that readable anymore.

The width of the message should depend on the content. If there is a wide \<code> block inside, the message can be wider, but smaller text only messages should continue having the small width.

Also mentioned here: https://github.com/gotify/server/issues/744#issuecomment-2509749549

jmattheis (changes requested):
Thanks for the contribution.

Victor-root:
Thanks for the review.

I updated the PR to keep `<pre>` content unwrapped with internal horizontal scrolling.

I also changed the message card sizing so short text-only messages keep the compact width, while wider content such as logs or code blocks can expand up to the available page width.

I retested it on desktop, laptop, and mobile web viewports.

Unrelated to this PR: would you be open to a separate PR adding an optional WebUI settings/preferences section?

The idea would be to keep all current defaults unchanged, but allow users to customize some UI preferences explicitly from the WebUI. For example, message layout/display preferences could live there instead of changing the default behavior for everyone.

I’m asking before working on it, since I’d rather avoid spending time on a larger UI change if this is not something you would want in Gotify.

ui/src/message/Message.tsx
```diff
@@ -168,6 +169,10 @@ const Message = ({
                 style={{
                     display: 'flex',
                     flexWrap: 'wrap',
+                    width: 'fit-content',
+                    minWidth: 'min(700px, 100%)',
```
jmattheis, line 173: This doesn't seem to work sometimes. E.g. with this messages

```
{
  "title": "hello",
  "message": "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.",
  "priority": 5,
  "extras": {
      "client::display": { "contentType": "text/markdown" }
  }
}

{
  "title": "hello",
  "message": "![](https://placecats.com/600/400)",
  "priority": 5,
  "extras": {
      "client::display": { "contentType": "text/markdown" }
  }
}
```

Both messages should have 700px width. as the width is for better readability of text.

Image

ui/src/message/Messages.tsx
```diff
@@ -115,6 +115,7 @@ const Messages = observer(() => {
     return (
```
jmattheis, line 115: > Unrelated to this PR: would you be open to a separate PR adding an optional WebUI settings/preferences section?

> The idea would be to keep all current defaults unchanged, but allow users to customize some UI preferences explicitly from the WebUI. For example, message layout/display preferences could live there instead of changing the default behavior for everyone.

Yes, we're open to that, but generally it's better to first create a ticket for proposed changes.

I prefer good defaults instead of settings but for some things like: Date format, message layout, relative/absolute time a setting would be okay. The discussion about possible changes could result in just changing the behavior instead creating a toggle for the feature.

When creating PRs please create separate PRs for each feature, so that we don't have a big changes, and can merge the subsets that don't require much discussion faster.

Victor-root:
Thanks for the feedback and the review time.

I’m closing this PR.

After testing the latest requested approach, it no longer solves the original issue. The more constrained/content-specific behavior keeps many real-world log-like notifications at the compact width, so it does not address the actual problem reported in #744: making these messages easier to read on wider screens.

This is not just my own use case; the issue was already reported by other users before I opened this PR.

I understand the concern about preserving the default readability/layout for regular text messages, but continuing in this direction would no longer fix the issue this PR was meant to solve.

I’ll maintain this behavior in my own fork/local build instead.

Thanks.

PR #962, opened 2026-05-17
fix: use zerolog

jmattheis:
This changes the logging to zerolog. What do you think @eternal-flame-AD? we could use log.* everywhere, but I kinda like zerolog more.

### DIFF
--- a/app.go
+++ b/app.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 	"time"
 
@@ -11,6 +10,9 @@ import (
 	"github.com/gotify/server/v2/model"
 	"github.com/gotify/server/v2/router"
 	"github.com/gotify/server/v2/runner"
+	"github.com/mattn/go-isatty"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


app.go
```diff
@@ -24,10 +26,12 @@ var (
 )
 
 func main() {
+	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.RFC3339})
```
eternal-flame-AD: This need to be guarded by a isTerminal check. A lot of people run Gotify with a supervisor (systemd, etc) or a pipe and will want the log clear of ANSI escape sequences.

eternal-flame-AD:
I'm good for getting better logging. I'm enjoying structured logging in my Rust projects so I like it as well.

app.go
eternal-flame-AD: Another thing is it doesn't look like zerolog has an environment variable for changing log levels? do we have to manually code that or?

jmattheis: Yeah, I'll add the log level in a separate PR with a rework of the overall config.

I've added terminal check + NO_COLOR check. I'll probably move the NO_COLOR env var later to the config package.

I also changed it to stdout, makes more sense as we don't output to stdout anywhere else.

eternal-flame-AD (dismissed):

Issue #325, opened 2020-06-02
Plain text api token


gougeon-s:
**Is your feature request related to a problem? Please describe.**
Tokens shouldn't be stored in plain text. (There isn't security issue to fill so I'm publishing here)

**Describe the solution you'd like**
The easiest to do is to store them hashed, like passwords. But unlike passwords, they need to be unique. Also, they aren't user input, that means that we can have a really long token and we can use fast hashing algorithm (long to crack but fast with the token)
So the main solution is to use a non-salted hash (e.g sha256) with a longer API token (e.g. 128 chars instead of 14).

**Describe alternatives you've considered**
An other solution would be to have an unique token_id with a salted hash.

**Additional context**
nothing

jmattheis:
Yeah, this is kinda intentional. Otherwise, tokens can't be viewed at any time inside the webui. 
I'll ask in our community chat if anyone uses this feature to decide if we can change this feature to be more secure. As this is a breaking change it will probably take some time before deciding.

The bcrypt default can be changed.

mdnix:
I was testing gotify today and I also noticed that the tokens are not hashed. I have seen that only the passwords are hashed and stored with bcrypt.

Is there any news about this?

It's not a big deal but just something I noticed.

jmattheis:
Nope, nothing happened yet.

eternal-flame-AD:
Why should token be hashed? I am a little confused.

The reason passwords are derived is that they have patterns and grant additional scope if the user do not use unique passwords:

- A brute force or heuristic brute force on the actual password that be used to compromise *other services* is difficult even if the hashed password is leaked. If you can get your hands on this hash it is likely the service is compromised already.
- It completely scrambles any visible pattern on the password.
- If two users use the same passwords they do not look the same in the database.

Tokens are randomized to begin with and grants no additional scope than the session originally has, if you can get your hand on the DB entry of that token chances are the whole service is already compromised, session integrity is already not a concern now.

The only thing that is useful I think is to avoid showing tokens back to the user once it is viewed to prevent session hijacking or impersonation that is caused by vulnerability on the client side. Hashing on the server side has nothing to do with this.

@jmattheis Do you think this is worth patching? If not I think we can close this.

p1gp1g:
[...]

PR #964, opened 2026-05-25
feat: allow sending message with client/basic path

jmattheis:
This PR removes the need for using application tokens for sending messages. This would allow us to hash our tokens as described in https://github.com/gotify/server/issues/325.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -325,17 +325,20 @@ func (a *MessageAPI) DeleteMessage(ctx *gin.Context) {
 	})
 }
 
-// CreateMessage creates a message, authentication via application-token is required.
+// CreateMessage creates a message, authentication via application token, client token, or basic auth is required.
 // swagger:operation POST /message message createMessage
 //
 // Create a message.
 //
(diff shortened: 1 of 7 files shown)


eternal-flame-AD:
It's too late where I am right now but I will try it out tomorrow. It looks good to me, are you committed to using hashed tokens? I can update the previous hashed token PR on top of this so we can get this feature done.

jmattheis:
Yeah, but let's do this the token hashing in a separate PR, as it probably needs a little more discussion.

I'd say we make the token hashing backwards compatible, so that we add a new token_hashed column (or do it via a different token format) which is filled for newly created client/apps/plugins while the old tokens continue to work without any database migration. If we define a new token format we could check when authenticating which column to check against.

The new token format could look like gtfya_ID_SECRET (GoTiFY App). We then parse ID, and then hash SECRET and compare it against the database value.

But let's discuss this in a different PR/ticket.

eternal-flame-AD (approved):
lgtm!

Issue #966, opened 2026-05-26
New Token Format Proposal


eternal-flame-AD:
The original threat model for this should be: attacker gets a read-only copy of the database, we should not allow them to make authenticated requests. In other words, BOTH the client AND the backend database record should be prevented from recovering or forging valid tokens.

The properties I think we should aim for:

- Contain data secret from the server, ideally in a way that can be used without ever being revealed back to the server
- Contain some authenticated or verifiable plaintext claims (like token type, application id, etc)
- Rotation and revocation should be possible without changing the claim itself
- Backwards compatible with verbatim tokens
- Not too long for webhook forms with length limits
- Not too rigidly bound to the entity ontology (i.e. a token doesn't necessarily need to have an associated user) for future features like sharing messages

A new token could look something like this: 

- A literal G (or gtfy), to distinguish it with old tokens.
- Underscore.
- A sequence of claims separared by underscore, a single letter denoting the next claim type, like "U1" means user ID 1, "A2" means application ID 2, "P3" means accessing resources for plugin 3. This is already similar to the property based authorization system in the auth package.
- An underscore.
- The "key" encoded in the token, randomly generated by the server at creation time. The server only records response of HMAC(key=key,salt=claims) and never commits the key to the database, so someone with read-only access to the database still have an HMAC preimage problem to solve to find a valid "key". (HKDF might be better in order to protect against confused deputy scenarios and gives us more degrees of freedom on binding more necessary context without canonicalization headaches, but we can decide that as we write the PR)
- Optional/for future: A period (.) then a counter , IIRC way back in the original discussion I put forward some reservations on the practical security of merely hashing a token and yet demand the client to send the "answer" to you in every request. This part of the token being present causes the database to update the max counter recorded in the system for each token, and the "key" part being double hashed with a further message containing the counter instead of the key itself. We can detail this up in another discussion.
- For future proofing we might want to generate "keys" that are in addition to being unpredictable, but also valid private keys in a public key crypto system (Curve25519, etc)

Currently we will only issue tokens mapping back to the current system, like "gtfy_U1_A2_0123456789abc" 

Footnote: the HMAC output can be simply stored by reusing the token field, we just need to make sure they don't start with "ACP" so they can't be confused with legacy verbatim tokens

References:

#325 
https://github.com/gotify/server/pull/964#issuecomment-4536081607
An old PR that is finally unblocked with #964: #707 

@jmattheis

jmattheis:
[...]

PR #971, opened 2026-06-02
feat: EdDSA token for database leakage/index mitigation

### LINKED ISSUE #325: Plain text api token

eternal-flame-AD:
Fixes #325 , supercedes #707 

Backend should be working with tests now, I'm currently settling on a timestamp based signature scheme like TOTP with more relaxed window (+-15 minutes for now).

also refactored some tests that used to rely on hijacking the token generation function to be consistency/property based.


jmattheis:
Could you answer the questions in https://github.com/gotify/server/issues/966#issuecomment-4573971138?

api/stream/client.go
```diff
@@ -16,7 +16,7 @@ var ping = func(conn *websocket.Conn) error {
 	return conn.WriteMessage(websocket.PingMessage, nil)
```
jmattheis, line 16: Do we want to include this PR in the Gotify 3 release?

eternal-flame-AD: I don't think it's a breaking change so maybe not necessary?

I can work on it a bit more next week and hopefully get a review Friday or so.

eternal-flame-AD: Oh actually probably yes we should merge it before v3.. we should do it since it will hide all the new tokens after generation and that might break workflows.

eternal-flame-AD:
1. I think we could leave the timestamp in but undocumented for now, I like the current shape allows us to adapt to future needs like signed shared message links. It's strictly gated behind a valid verify so it shouldn't open up additional security problems 
2. We probably should have some support for token rotation at least for applications (an elevated endpoint for regenerating tokens). I can create an issue for that later.

ui/src/client/AddClientDialog.tsx
```diff
@@ -7,59 +7,95 @@ import DialogTitle from '@mui/material/DialogTitle';
...
-                />
+                {returnToken ? (
+                    <>
+                        <DialogContentText>Your token will only be shown once.</DialogContentText>
+
+                        <span
+                            style={{padding: 16}}
+                            onClick={(e) => {
+                                window.getSelection()?.selectAllChildren(e.currentTarget);
+                            }}>
```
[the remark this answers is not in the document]
eternal-flame-AD: I tried it with Caret browsing it looks fine to me - I can move the cursor and select all or part of the token itself just fine.

auth/authentication.go
```diff
@@ -166,7 +174,7 @@ func (a *Auth) handleClient(checks ...func(*model.Client) (authState, error)) fu
 				return authStateSkip, err
 			}
 			if isCookie {
-				SetCookie(ctx.Writer, client.Token, CookieMaxAge, a.SecureCookie)
+				SetCookie(ctx.Writer, originalToken, CookieMaxAge, a.SecureCookie)
```
jmattheis, line 181: FYI: I don't think I'd want the timestamp based authentication for the UI in the future. This would require the UI to access the private key and as it doesn't have access as it's a HTTP only cookie. I'd say it logging the cookies isn't done normally.

We could add auto rotation for cookie based authentication, so the token is refreshed after some time. But yeah, that for another discussion.

auth/authentication.go
```diff
@@ -182,9 +190,17 @@ func (a *Auth) handleClient(checks ...func(*model.Client) (authState, error)) fu
...
 func (a *Auth) handleApplication(ctx *gin.Context) (authState, error) {
 	token, isCookie := a.readTokenFromRequest(ctx)
+	originalToken := token
 	if token == "" {
 		return authStateSkip, nil
 	}
+	if strings.HasPrefix(token, enhancedTokenPrefix) {
+		complexToken, err := ParseEnhancedToken(token)
+		if err != nil || !complexToken.ValidateTimestamp(timeNow().Unix()) {
+			return authStateSkip, nil
```
jmattheis: I think we should return errors when the token is malformed and abort authentication with 400.

If a client later signs the timestamp, it's possible that they client creates a malformed token, so I think the users should get a good error message here.

auth/token.go
```diff
@@ -1,20 +1,167 @@
...
+	tokenCharacters     = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_")
+	pluginPrefix        = "P"
+	enhancedTokenPrefix = "gtfy_"
 
 	randReader = rand.Reader
 )
 
+type EnhancedToken struct {
+	ident        string // a shared identifier, in formats like A12 for application ID 12
+	pubOrPrivKey []byte // public key
```
jmattheis:
```suggestion
	pubOrPrivKey []byte
```

auth/token.go
```diff
@@ -1,20 +1,167 @@
...
+}
+
+// Sign signs the timestamp with the private key and returns a new EnhancedToken.
+func (c *EnhancedToken) Sign(timestamp int64) (*EnhancedToken, error) {
+	if c.timestamp != 0 || len(c.signature) != 0 || len(c.pubOrPrivKey) != ed25519.SeedSize {
+		return nil, errNoPrivateKey
+	}
+	privKey := ed25519.NewKeyFromSeed(c.pubOrPrivKey)
+	sha512 := sha512.New()
+	sha512.Write([]byte("iss="))
```
jmattheis: iss is normally the shorthand for issuer. maybe use `iat` for issued at?

auth/token.go
```diff
@@ -1,20 +1,167 @@
...
-	tokenCharacters   = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-_")
-	randomTokenLength = 22 // ~2^132.5 keyspace (7.65e+39)
-	applicationPrefix = "A"
-	clientPrefix      = "C"
-	pluginPrefix      = "P"
+	errInvalidToken     = errors.New("invalid token")
+	errNoPrivateKey     = errors.New("no private key")
+	tokenCharacters     = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_")
+	pluginPrefix        = "P"
+	enhancedTokenPrefix = "gtfy_"
```
jmattheis: Any reason why it's gtfy_app and not gtfy.app? Maybe we could make it just `gtfya` (for app) and `gtfyc` (for client). So it's not too long. GitHub does this similar. https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#githubs-token-formats

auth/authentication.go
eternal-flame-AD: Makes sense, beyond the key name the current logic should be fine?

jmattheis: Yes

auth/authentication.go
eternal-flame-AD: I did this based on a wrapped error filter now so detailed reasoning would be returned if there is a parsing rejection. I decided to return 401 because it's still fundamentally lacking valid credentials, and 400 could be confusing to automated clients (say someone made a gotify GUI client that might consider a 400 an unrecoverable error, when an improperly formatted token should result in the GUI resulting reauth)

api/tokens_test.go
```diff
@@ -8,7 +8,12 @@ import (
 )
```
jmattheis, line 8: > We probably should have some support for token rotation at least for applications (an elevated endpoint for regenerating tokens). I can create an issue for that later.

Yeah, if the tokens aren't viewable in the UI anymore, this will be necessary for the 3.0 release. Currently, the token column is removed from the table. For existing users this may seem like a bug if there is no replacement. Can be done in a separate issue.

auth/authentication.go
```diff
@@ -148,9 +152,17 @@ func (a *Auth) handleUser(checks ...func(*model.User) (authState, error)) func(c
...
 	return func(ctx *gin.Context) (authState, error) {
 		token, isCookie := a.readTokenFromRequest(ctx)
+		originalToken := token
 		if token == "" {
 			return authStateSkip, nil
 		}
+		if strings.HasPrefix(token, enhancedTokenPrefix) {
+			complexToken, err := ParseEnhancedToken(token)
+			if err != nil || !complexToken.ValidateTimestamp(timeNow().Unix()) {
+				return authStateSkip, nil
```
jmattheis: Return the error here too.

auth/authentication.go
```diff
@@ -197,7 +201,7 @@ func (a *Auth) handleApplication(ctx *gin.Context) (authState, error) {
 	if strings.HasPrefix(token, enhancedTokenPrefix) {
```
jmattheis, line 201: In the database the public hey is stored in the same format as the private key. When a user inspects the database, it seems like it could be used for authentication, but in reality it's a different key with the same prefix.

Maybe we can differentiate the pub / private key format prefix.

But given that the public key is transferred when using signing it may be okay. What do you think?

model/application.go
```diff
@@ -17,7 +17,6 @@ type Application struct {
 	// The application token. Can be used as `appToken`. See Authentication.
 	//
 	// read only: true
-	// required: true
 	// example: AWH0wZ5r0Mbac.r
 	Token  string `gorm:"type:varchar(180);uniqueIndex:uix_applications_token" json:"token"`
```
jmattheis: maybe add a omitempty to client/app token. Now it's returned as empty string.

auth/token_test.go
```diff
@@ -3,32 +3,45 @@ package auth
...
-func TestTokenHavePrefix(t *testing.T) {
-	for i := 0; i < 50; i++ {
-		assert.True(t, strings.HasPrefix(GenerateApplicationToken(), "A"))
-		assert.True(t, strings.HasPrefix(GenerateClientToken(), "C"))
-		assert.True(t, strings.HasPrefix(GeneratePluginToken(), "P"))
-		assert.NotEmpty(t, GenerateImageName())
-	}
-}
+func TestNewComplexToken(t *testing.T) {
+	token := NewEnhancedToken("A12")
```
jmattheis:
```suggestion
	token := NewEnhancedToken("a")
```

auth/authentication.go
eternal-flame-AD: I'm not super sure what you are trying to say. Do you mean we shouldn't use the length of the key alone to distinguish private and public keys, or the database shouldn't look confusing to someone trying to recover the credential frm inside the database and would be blocked because of this? For the latter one I would think it doesn't matter that much - the logic is pretty clear to me: if it starts with gtfy then it's an asymetric credential. 

If you simply meant the private form could look different - sure ,what do you think would be good? I thought gtfy is a good thing because some users might have already modified the token field to something else, so I wanted some magic string that is not one character.

api/tokens_test.go
eternal-flame-AD: sounds good, let's do it in a separate issue/PR

auth/authentication.go
jmattheis: Yeah, I meant that the private key and public key from could look different. But having using the format for asymmetric creds sounds good.

jmattheis (approved):
👍

PR #976, opened 2026-06-13
Env config migrate

jmattheis:
This PR:
* Adds a cli with gotify serve, migrate-config, version
* Config migration with the yaml format to env file format
  * I've decided against filtering out default values, because we don't have inheritance anymore, and maybe a user wanted to explicitly set a default value, so that if the default is later changed, their config stays the same.

<details><summary>Migration example</summary>
<p>

```
# Example configuration file for the server.
# Save it to `config.yml` when edited

server:
  keepaliveperiodseconds: 0 # 0 = use Go default (15s); -1 = disable keepalive; set the interval in which keepalive packets will be sent. Only change this value if you know what you are doing.
  listenaddr: 'adrr' # the address to bind on, leave empty to bind on all addresses. Prefix with "unix:" to create a unix socket. Example: "unix:/tmp/gotify.sock".
  port: 80 # the port the HTTP server will listen on

  ssl:
    listenaddr: 'ddrr' # the address to bind on, leave empty to bind on all addresses. Prefix with "unix:" to create a unix socket. Example: "unix:/tmp/gotify.sock".
    port: 443 # the https port
    certfile: certfile # the cert file (leave empty when using letsencrypt)
    certkey: certkey # the cert key (leave empty when using letsencrypt)
    letsencrypt:
      enabled: false # if the certificate should be requested from letsencrypt
      accepttos: false # if you accept the tos from letsencrypt
      cache: data/certs # the directory of the cache from letsencrypt
      directoryurl: # override the directory url of the ACME server
      hosts: # the hosts for which letsencrypt should request certificates
      - mydomain.tld
      - myotherdomain.tld
  responseheaders: # response headers are added to every response (default: none)
    X-Custom-Header: "custom value"

  trustedproxies: # IPs or IP ranges of trusted proxies. Used to obtain the remote ip via the X-Forwarded-For header. (configure 127.0.0.1 to trust sockets)
  - 127.0.0.1/32
  - ::1
  securecookie: false # If the secure flag should be set on cookies. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure

  cors: # Sets cors headers only when needed and provides support for multiple allowed origins. Overrides Access-Control-* Headers in response headers.
    alloworigins:
    - '.+.example.com'
    - 'otherdomain.com'
    allowmethods:
    - "GET"
    - "POST"
    allowheaders:
    - "Authorization"
    - "content-type"
  stream:
    pingperiodseconds: 45 # the interval in which websocket pings will be sent. Only change this value if you know what you are doing.
    allowedorigins: # allowed origins for websocket connections (same origin is always allowed)
    - ".+.example.com"
    - "otherdomain.com"
oidc:
  enabled: false # Enable OpenID Connect login, allowing users to authenticate via an external identity provider (e.g. Keycloak, Authelia, Google).
  issuer: issue # The OIDC issuer URL. This is the base URL of your identity provider, used to discover endpoints. Example: "https://auth.example.com/realms/myrealm"
  clientid: client # The client ID registered with your identity provider for this application.
  clientsecret: secret # The client secret for the registered client.
  redirecturl: http://gotify.example.org/auth/oidc/callback # The callback URL that the identity provider redirects to after authentication. Must match exactly what is configured in your identity provider.
  autoregister: true # If true, automatically create a new user on first OIDC login. If false, only existing users can log in via OIDC.
  usernameclaim: preferred_username # The OIDC claim used to determine the username. Common values: "preferred_username" or "email".

database: # for database see (configure database section)
  dialect: sqlite3
  connection: data/gotify.db
defaultuser: # on database creation, gotify creates an admin user (these values will only be used for the first start, if you want to edit the user after the first start use the WebUI)
  name: admin # the username of the default user
  pass: admin # the password of the default user
pluginsdir: data/plugins # the directory where plugin resides (leave empty to disable plugins)
registration: false # enable registrations
```

```
$ go run . migrate-config config.example.yml
GOTIFY_DATABASE_CONNECTION="data/gotify.db"
GOTIFY_DATABASE_DIALECT="sqlite3"
GOTIFY_DEFAULTUSER_NAME="admin"
GOTIFY_DEFAULTUSER_PASS="admin"
GOTIFY_OIDC_AUTOREGISTER="true"
GOTIFY_OIDC_CLIENTID="client"
GOTIFY_OIDC_CLIENTSECRET="secret"
GOTIFY_OIDC_ENABLED="false"
GOTIFY_OIDC_ISSUER="issue"
GOTIFY_OIDC_REDIRECTURL="http://gotify.example.org/auth/oidc/callback"
GOTIFY_OIDC_USERNAMECLAIM="preferred_username"
GOTIFY_PASSSTRENGTH=10
GOTIFY_PLUGINSDIR="data/plugins"
GOTIFY_REGISTRATION="false"
GOTIFY_SERVER_CORS_ALLOWHEADERS="Authorization,content-type"
GOTIFY_SERVER_CORS_ALLOWMETHODS="GET,POST"
GOTIFY_SERVER_CORS_ALLOWORIGINS=".+.example.com,otherdomain.com"
GOTIFY_SERVER_KEEPALIVEPERIODSECONDS=0
GOTIFY_SERVER_LISTENADDR="adrr"
GOTIFY_SERVER_PORT=80
GOTIFY_SERVER_RESPONSEHEADERS="{\"X-Custom-Header\":\"custom value\"}"
GOTIFY_SERVER_SECURECOOKIE="false"
GOTIFY_SERVER_SSL_CERTFILE="certfile"
GOTIFY_SERVER_SSL_CERTKEY="certkey"
GOTIFY_SERVER_SSL_ENABLED="false"
GOTIFY_SERVER_SSL_LETSENCRYPT_ACCEPTTOS="false"
GOTIFY_SERVER_SSL_LETSENCRYPT_CACHE="data/certs"
GOTIFY_SERVER_SSL_LETSENCRYPT_ENABLED="false"
GOTIFY_SERVER_SSL_LETSENCRYPT_HOSTS="mydomain.tld,myotherdomain.tld"
GOTIFY_SERVER_SSL_LISTENADDR="ddrr"
GOTIFY_SERVER_SSL_PORT=443
GOTIFY_SERVER_SSL_REDIRECTTOHTTPS="true"
GOTIFY_SERVER_STREAM_ALLOWEDORIGINS=".+.example.com,otherdomain.com"
GOTIFY_SERVER_STREAM_PINGPERIODSECONDS=45
GOTIFY_SERVER_TRUSTEDPROXIES="127.0.0.1/32,::1"
GOTIFY_UPLOADEDIMAGESDIR="data/images"
```

</p>
</details>

### DIFF
--- a/app.go
+++ b/app.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/gotify/server/v2/config"
+	"github.com/gotify/server/v2/config/migrate"
 	"github.com/gotify/server/v2/database"
 	"github.com/gotify/server/v2/mode"
 	"github.com/gotify/server/v2/model"
@@ -60,6 +61,14 @@ func run(args []string, stdout, stderr io.Writer) int {
 		if ok {
 			fmt.Fprintln(stdout, b)
 		}
+		return 0
+	case "migrate-config":
+		content, err := migrate.Config(fs.Arg(1))
+		if err != nil {
+			fmt.Fprintln(stderr, err)
+			return 1
+		}
+		fmt.Fprintln(stdout, content)
 		return 0
 	default:
 		if command != "" {
--- a/app.go
+++ b/app.go
@@ -1,7 +1,12 @@
 package main
 
 import (
+	"errors"
+	"flag"
+	"fmt"
+	"io"
 	"os"
+	"runtime/debug"
 	"time"
 
 	"github.com/gotify/server/v2/config"
@@ -27,7 +32,45 @@ var (
 )
 
 func main() {
+	os.Exit(run(os.Args[1:], os.Stdout, os.Stderr))
+}
+
+func run(args []string, stdout, stderr io.Writer) int {
 	vInfo := &model.VersionInfo{Version: Version, Commit: Commit, BuildDate: BuildDate}
+	fs := flag.NewFlagSet("gotify", flag.ContinueOnError)
+	fs.SetOutput(stderr)
+	fs.Usage = func() { printUsage(stderr) }
+	if err := fs.Parse(args); err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			return 0
+		}
+		return 2
+	}
+
+	command := fs.Arg(0)
+	switch command {
+	case "serve", "":
+		return serve(vInfo)
+	case "version":
+		fmt.Fprintln(stdout, "Version:", vInfo.Version)
+		fmt.Fprintln(stdout, "Commit:", vInfo.Commit)
+		fmt.Fprintln(stdout, "Build Date:", vInfo.BuildDate)
+		fmt.Fprintln(stdout, "Go Build Info:")
+		b, ok := debug.ReadBuildInfo()
+		if ok {
+			fmt.Fprintln(stdout, b)
+		}
+		return 0
+	default:
+		if command != "" {
+			fmt.Fprintf(stderr, "gotify: unknown command %q\n\n", command)
+		}
+		printUsage(stderr)
+		return 2
+	}
+}
+
+func serve(vInfo *model.VersionInfo) int {
 	mode.Set(Mode)
 
 	conf, futureLogs := config.Get()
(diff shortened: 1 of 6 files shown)


app.go
```diff
@@ -27,7 +33,53 @@ var (
...
+	if err := fs.Parse(args); err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			return 0
+		}
+		return 2
+	}
+
+	command := fs.Arg(0)
+	switch command {
+	case "serve":
```
jmattheis: Maybe we could do this non-breaking and execute serve when gotify was execute without a command.

eternal-flame-AD (approved):

config/migrate/migrate.go
```diff
@@ -0,0 +1,185 @@
...
+		return ""
+	}
+	data, err := json.Marshal(m)
+	if err != nil {
+		return ""
+	}
+	return string(data)
+}
+
+func marshalList(values []string) string {
```
jmattheis, line 176: I changed this to use csv writing too, similar to the parsing. The escaping looks bad in the tests, but normally these values doesn't include `",` so it should only happen in the edge-cases.

eternal-flame-AD (approved):

jmattheis:
manually cherry picked to target branch.

PR #977, opened 2026-06-14
OIDC link user by iss/sub

jmattheis:
This PR:
* prevents login in as oidc user if a local user with the same username already exists
  * this can be enabled via a setting
* identifies oidc user by iss/sub, so even if the username claim changes, they still log into the same account
* prevents auto registration, if another oidc user with the same username exists
* adds a oidc e2e test

Sorry for the PR spam @eternal-flame-AD (:.

Issue reported by @KovachVL.

### DIFF
--- a/api/oidc.go
+++ b/api/oidc.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/gin-gonic/gin"
@@ -56,6 +58,7 @@ func NewOIDC(conf *config.Configuration, db *database.GormDatabase, userChangeNo
 		PasswordStrength:   conf.PassStrength,
 		SecureCookie:       conf.Server.SecureCookie,
 		AutoRegister:       conf.OIDC.AutoRegister,
+		LinkByUsername:     conf.OIDC.LinkByUsername,
 		pendingSessions:    decaymap.NewDecayMap[string, *pendingOIDCSession](time.Now(), pendingSessionMaxAge),
 	}
 }
@@ -83,6 +86,7 @@ type OIDCAPI struct {
 	PasswordStrength   int
 	SecureCookie       bool
 	AutoRegister       bool
+	LinkByUsername     bool
 	pendingSessions    *decaymap.DecayMap[string, *pendingOIDCSession]
 }
 
@@ -196,7 +200,7 @@ func (a *OIDCAPI) ElevateHandler(ctx *gin.Context) {
 //	        $ref: "#/definitions/Error"
 func (a *OIDCAPI) CallbackHandler() gin.HandlerFunc {
 	callback := func(w http.ResponseWriter, r *http.Request, tokens *oidc.Tokens[*oidc.IDTokenClaims], state string, provider rp.RelyingParty, info *oidc.UserInfo) {
-		user, status, err := a.resolveUser(info)
+		user, status, err := a.resolveUser(tokens.IDTokenClaims.GetIssuer(), info)
 		if err != nil {
 			http.Error(w, err.Error(), status)
 			return
@@ -362,7 +366,7 @@ func (a *OIDCAPI) ExternalTokenHandler(ctx *gin.Context) {
 		ctx.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to get user info: %w", err))
 		return
 	}
-	user, status, resolveErr := a.resolveUser(info)
+	user, status, resolveErr := a.resolveUser(tokens.IDTokenClaims.GetIssuer(), info)
 	if resolveErr != nil {
 		ctx.AbortWithError(status, resolveErr)
 		return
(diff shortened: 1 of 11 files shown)


api/oidc.go
```diff
@@ -386,32 +388,77 @@ func (a *OIDCAPI) generateState() (string, error) {
...
+//  3. Otherwise auto-register a new user, which requires GOTIFY_OIDC_AUTOREGISTER.
+func (a *OIDCAPI) resolveUser(issuer string, info *oidc.UserInfo) (*model.User, int, error) {
+	if issuer == "" {
+		return nil, http.StatusInternalServerError, errors.New("issuer claim was empty")
+	}
+	subject := info.GetSubject()
+	if subject == "" {
+		return nil, http.StatusInternalServerError, errors.New("subject claim was empty")
+	}
+	oidcID := issuer + "#" + subject
```
eternal-flame-AD, line 414: This should be safe on a compliant IdP but it's nice to check issuer is actually a valid URL without a fragment.

api/oidc.go
```diff
@@ -386,32 +388,77 @@ func (a *OIDCAPI) generateState() (string, error) {
...
+	}
+	user.OIDCID = &oidcID
+	if err := a.DB.UpdateUser(user); err != nil {
+		return nil, http.StatusInternalServerError, fmt.Errorf("failed to bind user to OIDC identity: %w", err)
+	}
+	return user, 0, nil
+}
+
+func (a *OIDCAPI) registerUser(username, oidcID string) (*model.User, int, error) {
+	if !a.AutoRegister {
```
eternal-flame-AD, line 460: Can we make GOTIFY_REGISTRATION imply this? I can't think of a reason why you want to disable auto register on OIDC but allow general registration.

eternal-flame-AD (commented):
Is there no way for an admin to create user for a specific OIDC identity without turning on link by username in general? 

I think an alternative implementation that covers more use cases is always make a user for an OIDC identity but make it locked by default - then give admin users a toggle to approve and unlock these users. This is basically what Gitea/Forgejo does with public idP logins.

api/oidc.go
```diff
@@ -56,6 +56,7 @@ func NewOIDC(conf *config.Configuration, db *database.GormDatabase, userChangeNo
 		PasswordStrength:   conf.PassStrength,
```
jmattheis, line 58: > Is there no way for an admin to create user for a specific OIDC identity without turning on link by username in general?
> 
> I think an alternative implementation that covers more use cases is always make a user for an OIDC identity but make it locked by default - then give admin users a toggle to approve and unlock these users. This is basically what Gitea/Forgejo does with public idP logins.

Is your usecase for that the admin doesn't want to enable autoregister, and manually whitelists users by either creating them manually or via unlocking?

For me this seems more of an addition than an alternative. The linking is the guard that an oidc user can take over the local account, so this has to be kept regardless, as if you have an local admin user, you probably don't want that a oidc user can take over that account.

I'd see this as extra toggle for autoregister. eg with values `false`, `true`, `locked`. This could be supported also for the normal registration.

api/oidc.go
jmattheis: A usecase could be if you have linking by username enabled, and don't want oidc users to take over an account, but still allow registrations with local accounts.

jmattheis: maybe we can combine this in one setting GOTIFY_REGISTRATION=false|oidc|local|true?

api/oidc.go
eternal-flame-AD: > For me this seems more of an addition than an alternative. The linking is the guard that an oidc user can take over the local account, so this has to be kept regardless, 

That's correct, when I said "alternative" I didn't mean alternative for the whole PR I meant alternative for the quite complicated (and IMO rigid) user linking logic. 

I was trying to say while this is a feature not really a bug, this seems like a must have feature for this to work in a real world capacity. In the previous implementation onboarding can simply be done by creating a user with the same username, but now there is no way to actually onboard an OIDC user without turning on auto register in general.

So in practice I am proposing instead of failing outright when autoregister is not opened, register it anyways but lock it by default and say need admin approval.

api/oidc.go
eternal-flame-AD: > A usecase could be if you have linking by username enabled, and don't want oidc users to take over an account, but still allow registrations with local accounts.

Isn't this problem already solved by this PR? you can't take over accounts by crossing over registration methods.

> maybe we can combine this in one setting GOTIFY_REGISTRATION=false|oidc|local|true

Yeah I would say merge it into one, or call it GOTIFY_REGISTRATION_METHOD=list. The current system is not bad I just prefer config that don't relate to each other despite being in completely different places.

api/oidc.go
jmattheis: Ahh, you mean the general logic of this, but it's not specific to the change in this PR (as this PR doesn't really removes functionality, it just changes to a safer default).

To summarize, you're use-case is using a public IdP like GitHub where you don't want to enable registrations, and you don't want to link by username.

Can we track this as a separate issue for "User Locking for OIDC/local registration". Are you having this use-case? otherwise I'd probably wait for user feedback on this. I'd say it's very likely when you are self-hosting gotify, that you're also selfhosting the IdP.

api/oidc.go
jmattheis: > Isn't this problem already solved by this PR? you can't take over accounts by crossing over registration methods.

Nevermind, misunderstood.

---

The list is a good idea so GOTIFY_REGISTRATION=oidc,local.

> Can we make GOTIFY_REGISTRATION imply this? I can't think of a reason why you want to disable auto register on OIDC but allow general registration.

But we want it the other way around, disable general registration, and allow OIDC registration. I think it's valid to allow to configure them separately.

With the current approach, I like that we have a namespace for all oidc related settings. I'd say "showing a registration form" and "automatically creating a user on oidc login" is somewhat different 

Now the oidc setting is called autoregister which makes it IMO a little more expressive. E.g.

```ini
GOTIFY_OIDC_ENABLED=true
GOTIFY_OIDC_ISSUER=https://auth.example.org
GOTIFY_OIDC_REDIRECTURL=https://gotify.example.org/auth/oidc/callback

GOTIFY_OIDC_AUTOREGISTER=true
# vs
GOTIFY_REGISTRATION=oidc
```
Maybe we could rename GOTIFY_OIDC_AUTOREGISTER to GOTIFY_OIDC_CREATE_USER to be more explicit.

api/oidc.go
eternal-flame-AD: Sure, I don't have use case for this personally but I know this is a standard feature on Gitea/Forgejo and a lot of systems in mixed organizations (like universities) are set up this way to whitelist only authorized personnel to certain applications.

eternal-flame-AD (approved):
Apart from the concern of lacking an approval based workflow this looks good to me

@KovachVL if you have time could you double check this?

KovachVL:
Hey! Yes, I'll check and comment a bit later, I just saw the notification now :)

KovachVL:
I reviewed the PR.

From a code perspective, this appears to address the original issue. The login flow now binds identities using a stable OIDC identifier derived from issuer + subject, instead of implicitly matching existing local users only by a mutable username claim. Linking by username is now an explicit opt-in via GOTIFY_OIDC_LINK_BY_USERNAME, with the default being disabled.

I do not currently see the original account-takeover path in the updated flow.

The only non-blocking point I would mention is migration behavior for any users that may have been created on earlier unreleased OIDC code paths without an oidc_id binding. Those cases may require a one-time manual bind or a temporary migration approach if you care about pre-release instances.

Other than that, the fix looks good to me.

P.s Sorry for replying so late

jmattheis:
@KovachVL thanks for the feedback!

PR #986, opened 2026-06-30
feat(security): application token refresh

### LINKED ISSUE #985: Credential Management API

An API or group of APIs that perform credential management features such as credential rotation, account lock/block, bind/rebind/unbind OIDC, etc.

This can probably simply be a virtual resources under /application/:id/security, the name is tentative, and then you post a descriptor of security update actions, such as:

```jsonc
{
    "rotateCredentials": true, // we should have this, require elevation
    "accountLock": { "locked": false }, // future proposal, this requires admin and not current account
    "oidc": {  .... }
}
```

Which the server will return with a matching result like:

```jsonc
{
    "rotateCredentials": { "token": "gtfya_..." }, 
}
````

Split these up into separate endpoints, but I want these to be in a single transaction as our current update coding pattern has atomicity problems, when applied to parallel security updates (not an uncommon thing for users to do), we could have data races and lead to critical security bugs.

**Additional context**

Let's only do the credential rotation endpoint, and I would say only implement it on application first - there are not many reasons why one might wish to keep their client ID so I would say once application tokens can be rotated this is no longer a release blocker. 

Additionally, exactly how the OIDC sessions should interact with token regenerations without creating unchecked session renewal vulnerabilities needs to be thought further.

### PR DESCRIPTION
**Pull request by eternal-flame-AD**
Fixes #985

### DIFF
--- a/api/application_test.go
+++ b/api/application_test.go
@@ -155,9 +155,9 @@ func (s *ApplicationSuite) Test_UpdateApplicationSecurity_regenerateToken() {
 	s.a.UpdateApplicationSecurity(s.ctx)
 	assert.Equal(s.T(), 200, s.recorder.Code)
 	bodyBytes, err := io.ReadAll(s.recorder.Body)
-	assert.Nil(s.T(), err)
+	assert.NoError(s.T(), err)
 	var got model.SecurityUpdateActionResponse
-	assert.Nil(s.T(), json.Unmarshal(bodyBytes, &got))
+	assert.NoError(s.T(), json.Unmarshal(bodyBytes, &got))
 	assert.Equal(s.T(), &model.SecurityUpdateActionResponse{
 		RegenerateToken: &model.RegenerateTokenResponse{
 			Token: got.RegenerateToken.Token,
@@ -180,7 +180,7 @@ func (s *ApplicationSuite) Test_UpdateApplicationSecurity_isNoOpIfNilAction() {
 	s.a.UpdateApplicationSecurity(s.ctx)
 	assert.Equal(s.T(), 200, s.recorder.Code)
 	bodyBytes, err := io.ReadAll(s.recorder.Body)
-	assert.Nil(s.T(), err)
+	assert.NoError(s.T(), err)
 	assert.Equal(s.T(), "{}", string(bodyBytes))
 	newToken, err := s.db.GetApplicationByID(1)
 	assert.NoError(s.T(), err)
--- a/api/application_test.go
+++ b/api/application_test.go
@@ -143,6 +143,50 @@ func (s *ApplicationSuite) Test_CreateApplication_ignoresReadOnlyPropertiesInPar
 	}
 }
 
+func (s *ApplicationSuite) Test_UpdateApplicationSecurity_regenerateToken() {
+	s.db.User(5).App(1)
+	test.WithUser(s.ctx, 5)
+
+	oldToken, err := s.db.GetApplicationByID(1)
+	assert.NoError(s.T(), err)
+	s.ctx.Request = httptest.NewRequest("PUT", "/application/1/security", bytes.NewBufferString(`{"regenerateToken": true}`))
+	s.ctx.Request.Header.Set("Content-Type", "application/json")
+	s.ctx.Params = gin.Params{{Key: "id", Value: "1"}}
+	s.a.UpdateApplicationSecurity(s.ctx)
+	assert.Equal(s.T(), 200, s.recorder.Code)
+	bodyBytes, err := io.ReadAll(s.recorder.Body)
+	assert.Nil(s.T(), err)
+	var got model.SecurityUpdateActionResponse
+	assert.Nil(s.T(), json.Unmarshal(bodyBytes, &got))
+	assert.Equal(s.T(), &model.SecurityUpdateActionResponse{
+		RegenerateToken: &model.RegenerateTokenResponse{
+			Token: got.RegenerateToken.Token,
+		},
+	}, &got)
+	newToken, err := s.db.GetApplicationByID(1)
+	assert.NoError(s.T(), err)
+	assert.NotEqual(s.T(), oldToken.Token, newToken.Token)
+}
+
+func (s *ApplicationSuite) Test_UpdateApplicationSecurity_isNoOpIfNilAction() {
+	s.db.User(5).App(1)
+	test.WithUser(s.ctx, 5)
+
+	oldToken, err := s.db.GetApplicationByID(1)
+	assert.NoError(s.T(), err)
+	s.ctx.Request = httptest.NewRequest("PUT", "/application/1/security", bytes.NewBufferString(`{}`))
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: api/application_test.go [hunk 1]
```diff
@@ -143,6 +143,50 @@ func (s *ApplicationSuite) Test_CreateApplication_ignoresReadOnlyPropertiesInPar
...
+
+	oldToken, err := s.db.GetApplicationByID(1)
+	assert.NoError(s.T(), err)
+	s.ctx.Request = httptest.NewRequest("PUT", "/application/1/security", bytes.NewBufferString(`{"regenerateToken": true}`))
+	s.ctx.Request.Header.Set("Content-Type", "application/json")
+	s.ctx.Params = gin.Params{{Key: "id", Value: "1"}}
+	s.a.UpdateApplicationSecurity(s.ctx)
+	assert.Equal(s.T(), 200, s.recorder.Code)
+	bodyBytes, err := io.ReadAll(s.recorder.Body)
+	assert.Nil(s.T(), err)
```
jmattheis: Use `assert.NoError` here. Same for the json.unmarshal and the other test method.

ui/src/application/AppStore.ts
```diff
@@ -36,6 +36,13 @@ export class AppStore extends BaseStore<IApplication> {
         this.snack('Application image updated');
     };
 
+    public async rekey(id: number): Promise<string> {
```
jmattheis: Can this be called regenerateToken, so we use a single name for this feature?

```
$ rg rekey
ui/src/tests/application.test.ts
111:            await page.click($table.cell(1, Col.EditRekey, '.rekey'));

ui/src/application/Applications.tsx
195:                            .rekey(toRekeyApp.id)
296:                    <IconButton onClick={fRekey} className="rekey">

ui/src/application/AppStore.ts
39:    public async rekey(id: number): Promise<string> {
```

ui/src/application/Applications.tsx
```diff
@@ -259,16 +291,27 @@ const Row = ({app, fDelete, fUpload, fDeleteImage, fEdit}: IRowProps) => {
...
+                    </IconButton>
+                </TableCell>
+            )}
+            {fDelete && (
+                <TableCell align="right" padding="none">
+                    <IconButton onClick={fDelete} className="delete" disabled={app.internal}>
+                        <Delete />
+                    </IconButton>
+                </TableCell>
+            )}
```
jmattheis: The fEdit + fDelete conditionals doesn't seem necessary, as the parameters aren't nullable.

ui/src/application/Applications.tsx
```diff
@@ -173,6 +181,29 @@ const Applications = observer(() => {
...
+                    fOnSubmit={() =>
+                        appStore
+                            .rekey(toRekeyApp.id)
+                            .then((token) => {
+                                setCreateDialogKnownToken(token);
+                                setCreateDialog(true);
+                            })
+                            .catch((error) => {
+                                snackManager.snack(error.message);
+                            })
```
jmattheis: I think the catch isn't necessary, due to this global handler: https://github.com/gotify/server/blob/master/ui/src/apiAuth.ts#L6

ui/src/application/AddApplicationDialog.tsx
```diff
@@ -13,12 +13,13 @@ import {copyToClipboard} from '../clipboard';
...
 
 interface IProps {
+    fKnownToken?: string;
     fClose: VoidFunction;
     fOnSubmit: (name: string, description: string, defaultPriority: number) => Promise<string>;
 }
 
-export const AddApplicationDialog = ({fClose, fOnSubmit}: IProps) => {
-    const [returnToken, setReturnToken] = useState('');
+export const AddApplicationDialog = ({fClose, fOnSubmit, fKnownToken}: IProps) => {
```
jmattheis: Can we create another `ViewTokenDialog` that is used by the application creation and regeneration features? It took me some time, why the token regeneration has to use the AddApplicationDialog, and I think separating this, will make this more easy to understand.

ui/src/application/Applications.tsx
```diff
@@ -196,13 +227,14 @@ const Applications = observer(() => {
 
 interface IRowProps {
     app: IApplication;
+    fRekey?: VoidFunction;
```
jmattheis:
```suggestion
    fRekey: VoidFunction;
```
The row component is only used here, and always filled with a callback. the check for fRekey to be set below can be removed too.

jmattheis (changes requested):

api/application_test.go
```diff
@@ -143,6 +143,50 @@ func (s *ApplicationSuite) Test_CreateApplication_ignoresReadOnlyPropertiesInPar
 	}
 }
 
+func (s *ApplicationSuite) Test_UpdateApplicationSecurity_regenerateToken() {
```
jmattheis, line 146: Slightly off-topic: Can you add a small description of the breaking api changes from your previous PR to https://gotify.net/docs/migrate-to-3#api-changes ?

ui/src/application/AddApplicationDialog.tsx
eternal-flame-AD: My reasoning was to generalize the dialog into a wizard - when you create application it's "enter details" -> "get token", when you regenerate token it goes straight to "get token". The intended benefit was to enable more complex workflows (like conditional account elevation, etc) without having to bubble the state into parent components, but you are right that's not how it's currently done (there is also UpdateApplicationDiaglog)

ui/src/tests/application.test.ts
```diff
@@ -61,12 +63,13 @@ const createApp =
...
         await page.click($dialog.button('.create'));
-        await page.waitForSelector($dialog.button('.finish'));
-        await page.waitForSelector($dialog.p('.token'));
-        const token = await innerText(page, $dialog.p('.token'));
-        expect(token.startsWith('gtfya.')).toBeTruthy();
-        await page.click($dialog.button('.finish'));
         await waitToDisappear(page, $dialog.selector());
+        await page.waitForSelector($tokenDialog.p('.token'));
+        const token = await innerText(page, $tokenDialog.p('.token'));
+        console.log('token', token);
```
jmattheis: Remove.

ui/src/client/Clients.tsx
```diff
@@ -80,9 +82,15 @@ const Clients = observer(() => {
...
             </Grid>
+            {toShowToken && (
+                <TokenConfirmDialog token={toShowToken} fClose={() => setToShowToken('')} />
+            )}
             {createDialog && (
                 <AddClientDialog
-                    fClose={() => setCreateDialog(false)}
+                    fClose={(token) => {
+                        setCreateDialog(false);
+                        setToShowToken(token || '');
```
jmattheis: Use `??` instead of `||`. It's normally the better choice: https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or

jmattheis (dismissed):

jmattheis (approved):

PR #988, opened 2026-07-04
fixup! feat(security): application token refresh

eternal-flame-AD:
Sorry I shouldn't have missed this important check. This adds back a missing userid check that prevents users from changing the application token for another.

Credits to @kaimandalic for catching this before release and sending in a report in private.

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -335,6 +335,11 @@ func (a *ApplicationAPI) UpdateApplicationSecurity(ctx *gin.Context) {
 		if success := successOrAbort(ctx, 500, err); !success {
 			return
 		}
+		if app == nil || app.UserID != auth.GetUserID(ctx) {
+			ctx.AbortWithError(404, fmt.Errorf("app with id %d doesn't exists", id))
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


api/application.go
```diff
@@ -335,20 +335,24 @@ func (a *ApplicationAPI) UpdateApplicationSecurity(ctx *gin.Context) {
...
 		}
-		action := model.SecurityUpdateAction{}
-		response := model.SecurityUpdateActionResponse{}
-		if err := ctx.Bind(&action); err == nil {
-			if action.RegenerateToken {
-				tokenPublic, tokenPrivate := generateApplicationToken()
-				app.Token = tokenPublic
-				response.RegenerateToken = &model.RegenerateTokenResponse{
-					Token: tokenPrivate,
+		if app != nil && app.UserID == auth.GetUserID(ctx) {
```
jmattheis: I'll change this to an early return.

> Sorry I shouldn't have missed this important check

No problem, I missed it too in the review :D.

jmattheis (approved):

PR #996, opened 2026-07-09
fix(api): keep reverse-proxy path prefix in paging next url

### LINKED ISSUE #664: include back path prefix
**Can the issue be reproduced with the latest available release? y**

- [x] Docker
- [ ] Linux machine
- [ ] Windows machine

```

```
</details>

- [ ] None
- [ ] Nginx
- [ ] Apache
- [x] Caddy

```
[…]
  route /gotify/* {
    uri strip_prefix /gotify
    # Set the port to the one you are using in gotify
    reverse_proxy localhost:1245
  }
[…]
```
</details>

- [ ] WebUI
- [ ] gotify-cli
- [ ] Android Client 

**What did you do?**

```
curl -H "x-gotify-key: $token" 'https://example.com/gotify/message?limit=100&since=xxx'
```

In the resulting JSON, `next` key points to `https://example.com/gotify/message?limit=100&since=yyy`

Instead it points to `https://example.com/message?limit=100&since=yyy` losing the `/gotify/` prefix.

The reverse-proxy configuration strips the prefix, so it's rather expected that gotify *receives* URLs without the prefix. So gotify should have an option to include that prefix when it *emits* URLs, like `next`.

TowyTowy:
Fixes #664

When gotify runs behind a reverse proxy that strips a subpath before forwarding (e.g. Caddy `uri strip_prefix /gotify`), the incoming request path no longer contains the prefix, so the `next` paging URL emitted by `GetMessages` dropped it — returning `https://example.com/message?...` instead of `https://example.com/gotify/message?...`.

As you noted in the issue, this case was missed in #127 (which fixed subpath handling for the WebUI/images by moving to relative URLs, but left the API's absolute `next` URL untouched).

This honors the standard `X-Forwarded-Prefix` header — which such proxies set to announce the stripped prefix — when building the `next` URL, consistent with how gotify already trusts `X-Forwarded-Proto`/`X-Forwarded-Host` for scheme and host. Behavior is unchanged when the header is absent, so it is fully backward compatible.

Proxy config example (Caddy): add `header_up X-Forwarded-Prefix /gotify` alongside `uri strip_prefix /gotify`.

Added `Test_GetMessages_WithLimit_BehindProxyPrefix_ReturnsNextWithPrefix`; `go test ./api/` passes.

On direction: you mentioned possibly deprecating `next` in favour of clients building URLs manually. This is meant as the minimal, backward-compatible fix that makes `next` correct today for existing clients, and it does not preclude that deprecation later — but if you would rather go straight to deprecating it, happy to close this. Your call.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -105,7 +105,7 @@ func buildWithPaging(ctx *gin.Context, paging *pagingParams, messages []*model.M
 		useMessages = messages[:len(messages)-1]
 		since = useMessages[len(useMessages)-1].ID
 		url := location.Get(ctx)
-		url.Path = ctx.Request.URL.Path
+		url.Path = withForwardedPrefix(ctx, ctx.Request.URL.Path)
 		query := url.Query()
 		query.Add("limit", strconv.Itoa(paging.Limit))
 		query.Add("since", strconv.FormatUint(uint64(since), 10))
@@ -116,6 +116,20 @@ func buildWithPaging(ctx *gin.Context, paging *pagingParams, messages []*model.M
 		Paging:   model.Paging{Size: len(useMessages), Limit: paging.Limit, Next: next, Since: since},
 		Messages: toExternalMessages(useMessages),
 	}
+}
+
+// withForwardedPrefix prepends the reverse-proxy path prefix to the given
+// request path. Reverse proxies that strip a subpath before forwarding the
+// request (e.g. Caddy's `uri strip_prefix /gotify`) can announce the stripped
+// prefix via the X-Forwarded-Prefix header, so that URLs emitted by gotify —
+// like the paging `next` link — keep pointing at the externally reachable path
+// instead of dropping the prefix. See #664.
+func withForwardedPrefix(ctx *gin.Context, path string) string {
+	prefix := strings.Trim(ctx.GetHeader("X-Forwarded-Prefix"), "/")
+	if prefix == "" {
+		return path
+	}
+	return "/" + prefix + "/" + strings.TrimPrefix(path, "/")
 }
 
 func withPaging(ctx *gin.Context, f func(pagingParams *pagingParams)) {
--- a/api/message_test.go
+++ b/api/message_test.go
@@ -106,6 +106,33 @@ func (s *MessageSuite) Test_GetMessages_WithLimit_ReturnsNext() {
 	test.BodyEquals(s.T(), expected, s.recorder)
 }
 
+func (s *MessageSuite) Test_GetMessages_WithLimit_BehindProxyPrefix_ReturnsNextWithPrefix() {
+	user := s.db.User(5)
+	app1 := user.App(1)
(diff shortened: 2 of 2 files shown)


eternal-flame-AD:
Hi, thanks for the contribution.

I think in general XF headers should be a gated opt-in feature due to security and misconfiguration risk. Maybe this time there isn't an immediate vulnerability, in the future we may want to introduce rate limiting and such and I hope there is a master control for all/each of these magic headers.

Also, given that XFP is so obscure it isn't even on MDN, maybe an alternative solution is just add a config option to override the public URL of the instance?

TowyTowy:
Thanks for taking a look, and fair point on the security framing.

For context on where things stand today: gotify already trusts `X-Forwarded-For`, but gated — `router.go` sets `RemoteIPHeaders = ["X-Forwarded-For"]` + `SetTrustedProxies(conf.Server.TrustedProxies)`, so gin only honors it when the request comes from a configured trusted proxy. The gap you're pointing at is real: this PR reads `X-Forwarded-Prefix` unconditionally, so unlike XFF it isn't covered by that `GOTIFY_SERVER_TRUSTEDPROXIES` master control.

Two ways I'm happy to take it, whichever you prefer:

1. Gate the `X-Forwarded-Prefix` read behind a trusted-proxy check that reuses the existing `TrustedProxies` config, so it's only honored from a trusted proxy and inherits the same master control — no new config, consistent with how XFF is already handled.
2. Your suggestion — add a config option for the instance's public base URL — and build the paging `next` URL from that instead of any forwarded header.

(1) keeps things working with zero extra config behind a correctly-configured proxy; (2) is more explicit and sidesteps the magic headers entirely. I lean slightly toward (1) since it reuses infrastructure you already have and needs no user action, but the design direction is your call — tell me which you'd rather and I'll rework it accordingly.

api/message.go
```diff
@@ -105,7 +105,7 @@ func buildWithPaging(ctx *gin.Context, paging *pagingParams, messages []*model.M
 		useMessages = messages[:len(messages)-1]
 		since = useMessages[len(useMessages)-1].ID
 		url := location.Get(ctx)
-		url.Path = ctx.Request.URL.Path
+		url.Path = withForwardedPrefix(ctx, ctx.Request.URL.Path)
```
jmattheis, line 108: I prefer if this would be changed to be consistent with the changes done in #127, so that a relative path is returned like it's now with the application images.

```json
{"id": 2, "image": "static/defaultapp.png", ...},
```
Example in the message api:
```
"paging": {
    "limit": 100,
    "next": "/message?limit=100&since=23",
    "since": 23,
    "size": 100
}
```
or
```
"paging": {
    "limit": 100,
    "next": "/application/3/message?limit=100&since=23",
    "since": 23,
    "size": 100
}
```

jmattheis: See https://github.com/gotify/server/pull/1000

PR #998, opened 2026-07-12
fix: create internal application when plugin adds Messenger after init

### LINKED ISSUE #653: Plugins messages not displayed upon refresh of the message page, nor can they be deleted
Yes

- [x] Docker
- [ ] Linux machine
- [ ] Windows machine

```
Standard config file using sqlite
```
</details>

- [x] None
- [ ] Nginx
- [ ] Apache
- [ ] Caddy

- [x] WebUI
- [ ] gotify-cli
- [ ] Android Client 

**What did you do?**

Yesterday I was trying to write a simple plugin for Gotify and I noticed a weird behaviour. Take the following plugin code:
<details>

<summary>The code</summary>

```go
package main

import (
	"time"

	"github.com/gin-gonic/gin"
	"github.com/gotify/plugin-api"
)

// GetGotifyPluginInfo returns gotify plugin info.
func GetGotifyPluginInfo() plugin.Info {
	return plugin.Info{
		ModulePath:  "github.com/thomas-maurice/plugin",
		Version:     "0.0.1",
		Author:      "Thomas Maurice",
		Website:     "https://gotify.net/docs/plugin",
		Description: "Plugin",
		License:     "MIT",
		Name:        "thomas-maurice/plugin",
	}
}

// Plugin is the gotify plugin instance.
type Plugin struct {
	msgHandler plugin.MessageHandler
}

// SetMessageHandler implements plugin.Messenger
// Invoked during initialization
func (c *Plugin) SetMessageHandler(h plugin.MessageHandler) {
	c.msgHandler = h
}

func (c *Plugin) Enable() error {
	go func() {
		time.Sleep(5 * time.Second)
		c.msgHandler.SendMessage(plugin.Message{
			Message: "The plugin has been enabled for 5 seconds.",
		})
	}()
	return nil
}

// Disable disables the plugin.
func (c *Plugin) Disable() error {
	return nil
}

// RegisterWebhook implements plugin.Webhooker.
func (c *Plugin) RegisterWebhook(basePath string, g *gin.RouterGroup) {
}

// NewGotifyPluginInstance creates a plugin instance for a user context.
func NewGotifyPluginInstance(ctx plugin.UserContext) plugin.Plugin {
	return &Plugin{}
}

func main() {
	panic("this should be built as go plugin")
}

```

</details>

I compiled it and ran it as per the instructions of the plugin template repo, it compiles and loads properly, so far so good. However when it starts to send messages the messages show up on the web ui, but for instance if I try to delete them it fails, and I see this error in the logs:

```
2024-05-15T09:31:45Z | 404 |     289.835µs |      172.17.0.1 | DELETE   "/message/49"
Error #01: message does not exist
```

Reloading the messages page also shows none of the messages sent by the plugin.

I tested with a standard app, and the messages sent via apps do survive a refresh and behave as i would expect.

This is very odd because if I inspect the database itself the messages are there:
```
sqlite> select * from messages where id=49;
49|0|The plugin has been enabled for 5 seconds.|foo|5||2024-05-15 09:31:42.899755887+00:00
```

I expect to see the plugin messages show up on the message page even after a refresh

The messages disappeared after a page restart, and I cannot seem to delete them from the UI when they show up (immediately after a publish)

TowyTowy:
Fixes #653

### Problem

When a plugin implements the `plugin.Messenger` interface *after* it has already been initialized for a user, its `PluginConf` already exists with no associated internal application (`ApplicationID == 0`). On the next startup, `initializeSingleUserPlugin` reuses that conf and wires the message handler with `ApplicationID: 0`, so every message the plugin sends is stored with `application_id = 0`. Those messages are orphaned: they don't appear after a page refresh and can't be deleted (DELETE returns 404). This matches the maintainer's diagnosis in #653.

### Fix

On initialization, if a plugin supports `Messenger` but has no internal application yet (`ApplicationID == 0`), create one and persist it on the conf — mirroring what already happens for plugins that support `Messenger` from the start. The internal-application creation is extracted into a shared helper. The back-fill is idempotent: plugins that already have an application are untouched.

### Testing

Added `TestNewManager_MessengerAddedAfterInit_createsApplication`, which seeds a conf without an application and asserts an internal application is created after re-initialization with a Messenger-capable plugin. It fails on `master` (ApplicationID stays 0) and passes with this change. `go test ./plugin/...`, `gofmt`/`gofumpt`, and `go vet` are clean.

---
Disclosure: this change was prepared with assistance from Claude (an AI tool); all code was reviewed and verified by me.

### DIFF
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -323,6 +323,20 @@ func (m *Manager) initializeSingleUserPlugin(userCtx compat.UserContext, p compa
 	m.instances[pluginConf.ID] = instance
 
 	if compat.HasSupport(instance, compat.Messenger) {
+		if pluginConf.ApplicationID == 0 {
+			// The Messenger capability was added after this plugin was first
+			// initialized for the user, so no internal application exists yet.
+			// Create one now, otherwise messages would be stored with
+			// application_id = 0 and become orphaned (not shown, not deletable).
+			app, err := m.createInternalApplication(info, userID)
+			if err != nil {
+				return err
+			}
+			pluginConf.ApplicationID = app.ID
+			if err := m.db.UpdatePluginConf(pluginConf); err != nil {
+				return err
+			}
+		}
 		instance.SetMessageHandler(redirectToChannel{
 			ApplicationID: pluginConf.ApplicationID,
 			UserID:        pluginConf.UserID,
@@ -399,15 +413,8 @@ func (m *Manager) createPluginConf(instance compat.PluginInstance, info compat.I
 		pluginConf.Config, _ = yaml.Marshal(instance.DefaultConfig())
 	}
 	if compat.HasSupport(instance, compat.Messenger) {
-		tokenPublic, _ := auth.GenerateApplicationToken()
-		app := &model.Application{
-			Token:       tokenPublic,
-			Name:        info.String(),
-			UserID:      userID,
-			Internal:    true,
-			Description: fmt.Sprintf("auto generated application for %s", info.ModulePath),
-		}
-		if err := m.db.CreateApplication(app); err != nil {
+		app, err := m.createInternalApplication(info, userID)
+		if err != nil {
 			return nil, err
 		}
 		pluginConf.ApplicationID = app.ID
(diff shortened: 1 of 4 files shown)


eternal-flame-AD (commented):
looks good to me, just one thing: can you add a double check in messagehandler.go to do a final check that the application ID is not zero? thanks. I wish we have foreign keys working.

TowyTowy:
Done. Added the safety net in `messagehandler.go`: `SendMessage` now returns an error if `ApplicationID == 0` rather than queueing a message that would be stored with `application_id = 0` and orphaned. Also added tests covering it and the internal-application back-fill error paths, which should clear the codecov patch/project checks. Agreed on the foreign keys wish.

(Prepared with AI assistance, reviewed and verified by me.)

eternal-flame-AD (approved):

PR #1000, opened 2026-07-15
fix: don't include scheme://host inside next

### LINKED ISSUE #664: include back path prefix

jmattheis:
Closes #996 
Fixes #664

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -3,13 +3,13 @@ package api
 import (
 	"encoding/json"
 	"errors"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
-	"github.com/gotify/location"
 	"github.com/gotify/server/v2/auth"
 	"github.com/gotify/server/v2/model"
 )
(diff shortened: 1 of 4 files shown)


eternal-flame-AD:
Did you forget to update the swagger spec?

jmattheis:
Good catch, edited the spec.

eternal-flame-AD (approved):

PR #1001, opened 2026-07-15
fix: error when legacy config exists

jmattheis:


config/file.go
```diff
@@ -11,7 +11,20 @@ import (
 
 var osStat = os.Stat
 
+func checkLegacyConfigFiles() error {
+	for _, file := range []string{"config.yml", "/etc/gotify/config.yml"} {
```
eternal-flame-AD, line 15: I don't think this is a great idea as it will make it impossible to start the server if there happens to be a config in /etc you are not authorized to delete. 

Config.yml also could be for anything. I would say this should be a warning or warning unless no environment variable that start with GOTIFY_ exists.

jmattheis: Fair. I wanted a way to prevent a user from just auto-updating, therefore the hard crash. But yeah, your right, I'll just close this PR.

PR #1012, opened 2026-07-26
fix: use makeObservable, remove decorators

jmattheis:
Closes #980


eternal-flame-AD:
If this is only blocking a dependency bump can we wait for a sec and see whether this can get fixed upstream? Decorators is already the recommended way for Mobx so I think it would be great if we follow that.

jmattheis:
Alright, we can wait. I'll close this for now.

PR #1014, opened 2026-07-27
fix: remove panic() and relay password hashing errors to UI

### LINKED ISSUE #1013: Can't create a new user, got a 500 error
- [x] Docker (Podman with NixOS)
- [ ] Linux machine
- [ ] Windows machine

```Nix
{
  virtualisation.oci-containers.containers.gotify = {
    image = "gotify/server:3.0.0";

    environmentFiles = [
      "/mnt/stormwind/containers/envs/commun.env"
      "/mnt/stormwind/containers/envs/gotify.env"
    ];

    extraOptions = [
      "--uidmap=0:1000:1"
      "--gidmap=0:100:1"
    ];

    hostname = "gotify";
    networks = ["repiproxy:ip=10.89.0.23"];

    volumes = [
      "/mnt/stormwind/containers/volumes/gotify-data:/app/data:U"

      "/home/admin/.dotfiles/data/root-leafayon_lan.crt:/app/cert/root-leafayon_lan.crt"
    ];
  };
}
```
</details>

- [ ] None
- [ ] Nginx
- [ ] Apache
- [x] Caddy

```Caddy
(gotify) {
  reverse_proxy 10.89.0.23:80
}

*.leafayon.lan leafayon.lan {
  tls internal

  # Networking
  #...

  # Medias
  #...

  # Services/Applications
  @gotify host gotify.leafayon.lan
  handle @gotify {
    import gotify
  }
}
```
</details>

- [x] WebUI
- [ ] gotify-cli
- [ ] Android Client 

**What did you do?**
I wanted the create a new user in the WebUI, I juste enter an username and a password with 128 charaters
But if I enter a simple password like 123456789 it works, but not a complex one

The new user created

An "undefined:undefined" error in the WebUI and an HTTP 500 error in the Firefox Web Deb Tools

Image

Image

eternal-flame-AD:
Fixes #1013

### DIFF
--- a/api/session_test.go
+++ b/api/session_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/gotify/server/v2/model"
 	"github.com/gotify/server/v2/test/testdb"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
@@ -39,9 +40,12 @@ func (s *SessionSuite) BeforeTest(suiteName, testName string) {
 	s.notified = false
 	s.a = &SessionAPI{DB: s.db, NotifyDeleted: s.notify}
 
+	pw, err := password.CreatePassword("testpass", 5)
+	require.NoError(s.T(), err)
+
 	s.db.CreateUser(&model.User{
 		Name: "testuser",
-		Pass: password.CreatePassword("testpass", 5),
+		Pass: pw,
 	})
 }
 
--- a/auth/password/password.go
+++ b/auth/password/password.go
@@ -1,14 +1,25 @@
 package password
 
-import "golang.org/x/crypto/bcrypt"
+import (
+	"errors"
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+func ValidateNewPassword(pw string) error {
+	if pw == "" {
+		return errors.New("password must not be empty")
+	}
+	if len([]byte(pw)) > 72 {
+		return bcrypt.ErrPasswordTooLong
+	}
+	return nil
+}
 
 // CreatePassword returns a hashed version of the given password.
-func CreatePassword(pw string, strength int) []byte {
+func CreatePassword(pw string, strength int) ([]byte, error) {
 	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(pw), strength)
-	if err != nil {
-		panic(err)
-	}
-	return hashedPassword
+	return hashedPassword, err
 }
 
 // ComparePassword compares a hashed password with its possible plaintext equivalent.
--- a/api/user.go
+++ b/api/user.go
@@ -188,10 +188,19 @@ func (a *UserAPI) GetCurrentUser(ctx *gin.Context) {
 func (a *UserAPI) CreateUser(ctx *gin.Context) {
 	user := model.CreateUserExternal{}
 	if err := ctx.Bind(&user); err == nil {
+		if err := password.ValidateNewPassword(user.Pass); err != nil {
+			ctx.AbortWithError(http.StatusBadRequest, err)
+			return
+		}
+		pw, err := password.CreatePassword(user.Pass, a.PasswordStrength)
+		if err != nil {
+			ctx.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to prepare password: %s", err))
+			return
+		}
 		internal := &model.User{
 			Name:  user.Name,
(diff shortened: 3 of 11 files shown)


api/session_test.go
```diff
@@ -39,9 +39,14 @@ func (s *SessionSuite) BeforeTest(suiteName, testName string) {
 	s.notified = false
 	s.a = &SessionAPI{DB: s.db, NotifyDeleted: s.notify}
```
jmattheis, line 41: Have you thought about adding validation to the model, and letting the password handling as is? E.g.
```
	Pass string `json:"pass,omitempty" form:"pass" query:"pass" binding:"required,max=72"`
```
This likely makes the error message more explicit. I'm okay with the current solution, but this may be a better solution.

api/session_test.go
```diff
@@ -39,9 +39,14 @@ func (s *SessionSuite) BeforeTest(suiteName, testName string) {
...
 
+	pw, err := password.CreatePassword("testpass", 5)
+	if err != nil {
+		s.T().Fatalf("Failed to create password: %v", err)
+	}
```
jmattheis:
```suggestion
	require.NoError(t, err)
```
(multiple occurences)

jmattheis (changes requested):

api/session_test.go
eternal-flame-AD: We will have to hardcode this number. I can see in the future users might request we migrate to other password hashes so I didn't go that route.

jmattheis: Yeah, but the error message will be better, currently it's a 500 internal server error for something that the user provided an invalid value for. How about we do both? Your change and the extra validation?

auth/password/password.go
```diff
@@ -1,14 +1,20 @@
...
+var ErrUnexpectedError = errors.New("unexpected error")
 
 // CreatePassword returns a hashed version of the given password.
-func CreatePassword(pw string, strength int) []byte {
+func CreatePassword(pw string, strength int) ([]byte, error) {
 	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(pw), strength)
-	if err != nil {
-		panic(err)
+	if err != nil && err != bcrypt.ErrPasswordTooLong {
+		err = ErrUnexpectedError
```
jmattheis: If we mask the error here, we don't get the initial error anywhere, meaning if a user misconfigures the passstrength, it will always return unexpected error without any log about the actual error. I think we can pass the full error without masking into the 500 internal server error, and with the extra validation the length check is done before and results in a 400.

auth/password/password.go
```diff
@@ -1,14 +1,30 @@
...
+import (
+	"errors"
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+var ErrUnexpectedError = errors.New("unexpected error")
+
+func ValidateNewPassword(pw string) error {
+	if pw == "" {
```
jmattheis, line 10: This is already checked by the binding on the property (`binding:"required"`). Please add the 72 char validation there too. Then we don't need this method.

eternal-flame-AD: - I actually didn't figure out how to make sure it's no more than 72 bytes long using validator.. do you know how?

jmattheis: Ahh true, it bytes and not len on string. Then nevermind about the validation on the binding.

Can we do this as extra check for the create password method in the
```
		pw, err := password.CreatePassword(user.Pass, a.PasswordStrength)
		if err != nil {
			status := http.StatusInternalServerError
			if errors.Is(err, bcrypt.ErrPasswordTooLong) {
				status = http.StatusBadRequest
			}
			ctx.AbortWithError(status, fmt.Errorf("failed to prepare password: %s", err))
			return
		}
```

and remove the masking described in https://github.com/gotify/server/pull/1014/changes#r3666911979.

eternal-flame-AD: I actually like a separate preflight check better because it makes it easier to implement things like live password strength validation in the future (which can be hard to do using only the frontend due to dictionaries, platform parity, etc). wdyt?

jmattheis: I prefer the direct approach because then the check is only done once in the bcrypt function. Now it looks like the done a second time for getting a different status code. Regarding future features for passwords: I can see this, but I'd prefer this to be kept simple, so this can be implemented when it's needed not as a precaution.

We can keep the method, I don't think it makes to much of a difference, but the error masking should be fixed.

jmattheis (approved):

Issue #1007, opened 2026-07-22
Option to disable local login when OIDC is enabled


MichielMak:
**Is your feature request related to a problem? Please describe.**
When OIDC SSO is enabled, the local username/password login form is still shown and the `/auth/local/login` endpoint stays active. In setups where all users should authenticate through the identity provider (centralized auth, MFA/policy enforced at the IdP, deprovisioning handled centrally), operators want to fully turn off local login so it can't be used as a bypass.

**Describe the solution you'd like**
A new boolean config option, e.g. `GOTIFY_OIDC_DISABLELOCALLOGIN` (default `false`). When it is `true` **and** OIDC is enabled:
- Backend rejects `POST /auth/local/login` with `403` and a clear error instead of `404`.
- UI hides the username/password form (and the register button), leaving only the "Login with OIDC" button.
- `/gotifyinfo` reports local-login availability so clients can adapt.

**Critical safety behavior:** the flag must ONLY take effect when OIDC is actually enabled. If `DISABLELOCALLOGIN=true` but OIDC is disabled, the flag is ignored (local login stays working) and a warning is logged, so an operator can never lock themselves out.

**Describe alternatives you've considered**
- Reverse-proxy rules blocking `/auth/local/login`: works but is external, easy to misconfigure, and doesn't hide the UI form.
- Removing local accounts entirely: not viable, a break-glass admin is often still wanted.

**Additional context**
We already have a working implementation on a local branch and are happy to open a PR if the maintainers are open to it. Rough scope:
- `config/keys.go`, `config/config.go`: new env var + `OIDC.DisableLocalLogin` field.
- `router/router.go`: conditional `/auth/local/login` handler (403 when disabled) + ignore-with-warning guard.
- `model/gotifyinfo.go`, `ui/serve.go`: expose `localLogin` to clients/UI.
- `ui/src/config.ts`, `ui/src/user/Login.tsx`: hide form/register, keep OIDC button.
- Docs (`gotify-server.env.example`), swagger spec, and tests.

Happy to align with your preferred naming/behavior before submitting.

jmattheis:
Yeah, you can submit this, with these changes:
* Besides /auth/local/login the handleUser path inside auth/authentication.go has to be disabled too, otherwise the local login credentials can still be used via basic auth on other api endpoints.
* Setting should be called GOTIFY_LOCALAUTH_ENABELED with default `true`
[...]

PR #1020, opened 2026-08-04
feat: #1007 - Option to disable local login when OIDC is enabled  

DerDummePunkt:
This PR attempts to implement #1007 

Adds configurable local authentication through GOTIFY_LOCALAUTH_ENABLED, defaults to true to preserve existing behavior.

When local authentication is disabled:
- The server does not register POST /auth/local/login.
- Basic Auth with local Gotify user credentials is rejected by authentication middleware.
- The check for a local admin account and subsequent creation is skipped.
- The UI hides the local username/password login form, registration button, and password-based elevation form.
- The UI receives the public localauth flag through injected window.config.
- GET /gotifyinfo exposes the same flag as localauth.

The server exits with a fatal error when both local authentication and OIDC are disabled, preventing startup without an available login method.

### DIFF
--- a/app.go
+++ b/app.go
@@ -106,7 +106,7 @@ func serve(vInfo *model.VersionInfo) int {
 		return 1
 	}
 
-	db, err := database.New(conf.Database.Dialect, conf.Database.Connection, conf.DefaultUser.Name, conf.DefaultUser.Pass, conf.PassStrength, true, time.Now)
+	db, err := database.New(conf.Database.Dialect, conf.Database.Connection, conf.DefaultUser.Name, conf.DefaultUser.Pass, conf.PassStrength, conf.LocalAuthEnabled, time.Now)
 	if err != nil {
 		log.Error().Err(err).Msg("Cannot initialize database")
 		return 1
--- a/auth/authentication.go
+++ b/auth/authentication.go
@@ -18,6 +18,7 @@ const (
 	authStateForbidden
 	authStateNotElevated
 	authStateOk
+	authStateLocalAuthDisabled
 )
 
 const (
@@ -39,9 +40,10 @@ type Database interface {
 
 // Auth is the provider for authentication middleware.
 type Auth struct {
-	DB           Database
-	SecureCookie bool
-	CrossOrigin  *http.CrossOriginProtection
+	DB               Database
+	SecureCookie     bool
+	LocalAuthEnabled bool
+	CrossOrigin      *http.CrossOriginProtection
 }
 
 // RequireAdmin requires an elevated client token or basic auth, the user must be an admin.
@@ -109,6 +111,9 @@ func (a *Auth) evaluate(ctx *gin.Context, funcs ...func(ctx *gin.Context) (authS
 		case authStateNotElevated:
 			ctx.AbortWithError(403, errors.New("session not elevated, use basic auth or call /client:elevate"))
 			return true
+		case authStateLocalAuthDisabled:
+			ctx.AbortWithError(403, errors.New("local authentication is disabled"))
+			return true
 		case authStateOk:
 			ctx.Next()
 			return true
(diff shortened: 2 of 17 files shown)


auth/authentication.go
```diff
@@ -146,6 +147,9 @@ func (a *Auth) rejectForeignOrigin(ctx *gin.Context) bool {
 
 func (a *Auth) handleUser(checks ...func(*model.User) (authState, error)) func(ctx *gin.Context) (authState, error) {
 	return func(ctx *gin.Context) (authState, error) {
+		if !a.LocalAuthEnabled {
+			return authStateSkip, nil
+		}
```
jmattheis: Instead of skipping this should error and abort.
* Create a new authState named authStateAuthDisabled
* Return 403 with the message "local authentication is disabled" in the evaluate method.

router/router.go
```diff
@@ -158,7 +159,9 @@ func Create(db *database.GormDatabase, vInfo *model.VersionInfo, conf *config.Co
 
 	g.Group("/user").Use(authentication.Optional).POST("", userHandler.CreateUser)
 
-	g.POST("/auth/local/login", sessionHandler.Login)
```
jmattheis, line 162: This endpoint should still be registered. 404 seems like a weird response. It should instead return 403 with message "local authentication is disabled"

app.go
```diff
@@ -85,6 +85,9 @@ func serve(vInfo *model.VersionInfo) int {
 	conf, futureLogs := config.Get()
 	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: time.RFC3339, NoColor: noColor(conf.NoColor)}).Level(zerolog.Level(conf.LogLevel))
 	log.Info().Str("version", vInfo.Version).Str("build_date", BuildDate).Msg("Gotify")
+	if !conf.LocalAuthEnabled && !conf.OIDC.Enabled {
+		log.Fatal().Msg("either local authentication or OIDC must be enabled")
+	}
```
jmattheis: Can you move this check to the end of config.Get, so it's included in the futureLogs that is used in this method?

model/gotifyinfo.go
```diff
@@ -14,6 +14,11 @@ type GotifyInfo struct {
 	// required: true
 	// example: true
 	Register bool `json:"register"`
+	// If local authentication is enabled.
+	//
+	// required: true
+	// example: true
+	LocalAuth bool `json:"localauth"`
```
jmattheis:
```suggestion
	LocalAuth bool `json:"localAuth"`
```
Case this with camel case:
```
$ rg localauth
docs/spec.json
2943:        "localauth",
2947:        "localauth": {

model/gotifyinfo.go
21:     LocalAuth bool `json:"localauth"`

ui/src/user/Login.tsx
20:    const localAuthEnabled = config.get('localauth');

ui/src/common/ElevationForm.tsx
18:    const localAuthEnabled = config.get('localauth');

ui/src/config.ts
8:    localauth: boolean;
22:    localauth: true,

ui/serve.go
21:     LocalAuth bool              `json:"localauth"`

```

jmattheis (changes requested):
Looks good so far. I've some remarks and haven't tested it manually.

router/router.go
DerDummePunkt: Im not 100% sure how to implement that.

For now, im wiring the `LocalAuthEnabled` config flag into the `SessionAPI` struct, so i can inject the 403 into the context in its `Login` Handler function

I could also add a new Middleware function to the authentication layer for that and instruct the router to use that.

app.go
DerDummePunkt: Im not sure i understand that.
What method are you referring to? Do you mean the `serve` function?

jmattheis: Yes, I mean the serve method. Basically I don't want to have config based validation in a serve method, so everything config related is handled in config#get, and serve just handles the futureLogs returned by config#get. You changed it how I meant it.

router/router.go
jmattheis: Adding it to the session api is good enough for now.

jmattheis (commented):
I've made some small fixes.

I'll test this some more and will merge it around end of next week. (For security related stuff I want to take my time and I'm currently a bit busy).

DerDummePunkt:
No worries, take your time, there's no rush

jmattheis (approved):
Thanks

Issue #991, opened 2026-07-07
OIDC: UX improvements


jmattheis:
From #433 

* Auto redirect to IdP when unauthorized in the WebUI.
* Allow customizing the "Login with OIDC" label in the UI.

DerDummePunkt:
Hey @jmattheis,
not quite sure if an issue you yourself create counts as an _approved_ feature according to [the contribution guidelines](https://github.com/gotify/server/blob/master/CONTRIBUTING.md), but i guess so.

Mind if i give this a shot? I just had a look at the server repos code and that doesnt seem to difficult to implement,
i've been using gotify since the release of 3.0 and i really like it, the lack of exacly this feature was pretty much the first thing i noted once i had my instance up and running and i would love to see this implemented.

jmattheis:
@DerDummePunkt Sure, you can give it a shot. I've removed one point from this ticket, as we already have a separate ticket for this => https://github.com/gotify/server/issues/1007.

DerDummePunkt:
Hmm, i didnt see #1007, disabling local login when OIDC is enabled was actually the most important point for me here.

is there anyone actually working on #1007?
If not, i might tackle both, i got some free time on my hands.

After having read #1007, there are some concerns listed in it that might also come into play here:
* automatically redirecting to the IdP only makes sense to me when local login is disabled
* when local login is still enabled, should the UI still automatically redirect to the IdP? Imho it shouldnt
* alternatively, we could add yet another env variable to control the behaviour when both Login variants are enabled

jmattheis:
It should be behind a config flag, even if localauth is disabled. GOTIFY_OIDC_AUTO_REDIRECT=false. With default = false. I'm not sure if someone is working on 1007, but you can probably give it a try.

PR #1022, opened 2026-08-05
#991 ux improvements: OIDC Login Label

DerDummePunkt:
Implements part of #991

adds env var GOTIFY_OIDC_IDP_NAME with default "OIDC"

variable is used in the UI for button labels at login and elevation forms

Few notes here:
* I decided to name the new env var `GOTIFY_OIDC_IDP_NAME`, since the value is used for both the login button label as well as the label of the elevation form button. Let me know if you want that variable name changed.

* I wasnt really sure what tests to add here. I could have written something like
   ```go
   func TestOIDCIDPName(t *testing.T) {
    t.Setenv("GOTIFY_OIDC_IDP_NAME", "Authelia")

    conf, _ := Get()

    assert.Equal(t, "Authelia", conf.OIDC.IDPName)
    }
   ```
  in `config/config_test.go`, but all that would guard against is that the env variable name `GOTIFY_OIDC_IDP_NAME` doesnt change in the `keys.go`, pretty useless imho

* I havent added any length limit/check yet, I say, if some poor soul wants to put the combined works of Williams Shakespear into that button, let them. Values longer than a few chars cause a line break in the button, personally I think that still looks fine, still, if you want some arbitrary limit on that, let me know.
image


jmattheis:
> I wasnt really sure what tests to add here.

Yeah, no need for tests here :+1:.

> I havent added any length limit/check yet

Sounds good

jmattheis (approved):
Thanks!

PR #1025, opened 2026-08-08
feat(ui): highlight current client session

### LINKED ISSUE #677: Do not allow deleting current client from GUI

You could accidentally delete your current client from within the GUI.

There should be an indication that its your current client, and delete/hide the delete button.

None

**Additional context**

Concept:

image

SulimanAbdulrazzaq:
Fixes #677

The clients table now highlights the client associated with the authenticated session. The current row also exposes `aria-current=true` so the indication is available to assistive technology.

A browser regression assertion verifies that exactly one client row is marked current and that it matches the login client. Existing deletion behavior is unchanged, following the maintainer’s guidance.

## Validation

- Prettier 3.6.2 check
- TypeScript 6.0.3 TSX parse
- `git diff --check`

The full UI suite, build, and runtime browser test could not be run locally because the exact frozen-lockfile installation timed out while fetching `@mui/icons-material` from the Yarn registry. No dependency files were changed.

### DIFF
--- a/ui/src/client/Clients.tsx
+++ b/ui/src/client/Clients.tsx
@@ -26,7 +26,7 @@ import {useStores} from '../stores';
 import {TokenConfirmDialog} from '../common/TokenConfirmDialog';
 
 const Clients = observer(() => {
-    const {clientStore} = useStores();
+    const {clientStore, currentUser} = useStores();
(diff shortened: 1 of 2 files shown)


jmattheis (approved):
Thanks!

PR #1027, opened 2026-08-11
Add GOTIFY_LOCALAUTH_ENABLED to disable local authentication

### LINKED ISSUE #1007: Option to disable local login when OIDC is enabled

MichielMak:
Closes #1007

Adds an option to turn off Gotify's built-in username/password authentication, so an instance can rely on OIDC only. Implemented following @jmattheis' design in #1007.

## The option

`GOTIFY_LOCALAUTH_ENABLED` (`localauth.enabled`), **default `true`**. Existing installations are unaffected.

When it is `false`:

- **HTTP basic auth with a username/password is rejected on every endpoint.** Client tokens, application tokens and OIDC sessions keep working.
- `POST /auth/local/login` returns `403` with a clear error.
- The default admin user is not created on startup.
- The WebUI hides the login form and the register button, leaving only the OIDC button.

Gotify refuses to start when both local auth and OIDC are disabled, since that would leave no way to log in.

## Why basic auth had to change

Blocking the login endpoint alone is not enough — Gotify also accepts a username and password directly via HTTP basic auth on the regular API endpoints. If only the login route were blocked, someone could keep using local credentials there and the option would be cosmetic.

The fix is a single guard in `auth/authentication.go`. `handleUser()` is the only place a password is checked for basic auth, and all six middlewares (`RequireAdmin`, `RequireClient`, `RequireElevatedClient`, `RequireApplicationToken`, `RequireApplicationOrClient`, `Optional`) go through it, so one guard covers them all. `handleClient` and `handleApplication` are untouched, which is why token auth is unaffected.

Only two places in the codebase verify a password: this one, and `SessionAPI.Login` behind the now-403 route. So with the option off there is no remaining password login path. There is a test (`TestLocalAuthDisabledRejectsBasicAuth`) that fails if the guard is removed.

## `/gotifyinfo`

Gained an additive `localauth` boolean so the UI knows whether to show the login form. Clients that ignore it are unaffected. `docs/spec.json` is regenerated with `make update-swagger`.

## Two notes for review

**1. Refusing to start.** You asked for `log.Fatal()`. I used the existing `futureFatal` mechanism in `config.Get()` instead, which is how other config errors are reported — same result (fatal message, exit 1), but it stays unit-testable, whereas `log.Fatal()` calls `os.Exit`. Happy to switch to a literal `log.Fatal()` if you prefer.

**2. Deliberately left alone.** Two things are out of scope here, tell me if you want either included:

- With `GOTIFY_REGISTRATION=true` and local auth off, `POST /user` still creates accounts with a password that can never be used. Arguably it should be refused for the same reason the default admin is skipped.
- An OIDC user can still change their password via the API. It is a dead write today, but it may be intentional so an operator can re-enable local auth later.

The docs repo probably needs a matching entry for the new option.

## Testing

`go build ./...` and `go test ./...` pass, plus `tsc`, `eslint` and `prettier` on the UI. New tests cover the config parsing and start-up refusal, the `403` on the login route, basic auth rejection, and that no default user is created.

### DIFF
--- a/app.go
+++ b/app.go
@@ -106,7 +106,7 @@ func serve(vInfo *model.VersionInfo) int {
 		return 1
 	}
 
-	db, err := database.New(conf.Database.Dialect, conf.Database.Connection, conf.DefaultUser.Name, conf.DefaultUser.Pass, conf.PassStrength, true, time.Now)
+	db, err := database.New(conf.Database.Dialect, conf.Database.Connection, conf.DefaultUser.Name, conf.DefaultUser.Pass, conf.PassStrength, conf.LocalAuth.Enabled, time.Now)
 	if err != nil {
 		log.Error().Err(err).Msg("Cannot initialize database")
 		return 1
--- a/auth/authentication.go
+++ b/auth/authentication.go
@@ -42,6 +42,9 @@ type Auth struct {
 	DB           Database
 	SecureCookie bool
 	CrossOrigin  *http.CrossOriginProtection
+	// LocalAuthEnabled controls whether username/password (basic auth) credentials
+	// are accepted. When false, only token based auth and OIDC sessions work.
+	LocalAuthEnabled bool
 }
 
 // RequireAdmin requires an elevated client token or basic auth, the user must be an admin.
@@ -146,6 +149,9 @@ func (a *Auth) rejectForeignOrigin(ctx *gin.Context) bool {
 
 func (a *Auth) handleUser(checks ...func(*model.User) (authState, error)) func(ctx *gin.Context) (authState, error) {
 	return func(ctx *gin.Context) (authState, error) {
+		if !a.LocalAuthEnabled {
+			return authStateSkip, nil
+		}
 		if name, pass, ok := ctx.Request.BasicAuth(); ok {
 			if user, err := a.DB.GetUserByName(name); err != nil {
 				return authStateSkip, err
--- a/auth/authentication_test.go
+++ b/auth/authentication_test.go
@@ -29,7 +29,7 @@ type AuthenticationSuite struct {
 func (s *AuthenticationSuite) SetupSuite() {
 	mode.Set(mode.TestDev)
 	s.DB = testdb.NewDB(s.T())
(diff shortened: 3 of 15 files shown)


jmattheis:
Hey thanks for the contribution. Another user created a PR for this feature: https://github.com/gotify/server/pull/1020, and this PR is already reviewed and nearly merged. Therefore, I'll close this one.

PR #1028, opened 2026-08-12
feat(oidc): allow disabling client_secret on external flow

gekmihesg:
Add GOTIFY_OIDC_EXTERNALSECRET to allow omitting the client_secret from OIDC token exchange for the external (mobile app) flow.

This is required to get OIDC to work with Microsoft Entra that rejects requests containing a client_secret for mobile and desktop applications as they are marked as public client (gotify://oidc/callback).

The error I got when trying to use OIDC from the mobile app: `errors="Error #01: token exchange failed: oauth2: "invalid_client" "AADSTS700025: Client is public so neither 'client_assertion' nor 'client_secret' should be presented.`
In the browser, it worked fine. With this patch, OIDC works as expected for the mobile app as well.

### DIFF
--- a/api/oidc.go
+++ b/api/oidc.go
@@ -57,9 +57,26 @@ func NewOIDC(conf *config.Configuration, db *database.GormDatabase, userChangeNo
 		log.Fatal().Err(err).Msg("failed to initialize OIDC provider")
 	}
 
+	externalProvider := provider
+	if !conf.OIDC.ExternalSecret {
+		externalProvider, err = rp.NewRelyingPartyOIDC(
+			context.Background(),
+			conf.OIDC.Issuer,
+			conf.OIDC.ClientID,
+			"",  // no secret, PKCE only
+			conf.OIDC.RedirectURL,
+			conf.OIDC.Scopes,
+			opts...,
+		)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


api/oidc.go
```diff
@@ -57,9 +57,26 @@ func NewOIDC(conf *config.Configuration, db *database.GormDatabase, userChangeNo
 		log.Fatal().Err(err).Msg("failed to initialize OIDC provider")
 	}
 
+	externalProvider := provider
+	if !conf.OIDC.ExternalSecret {
+		externalProvider, err = rp.NewRelyingPartyOIDC(
+			context.Background(),
+			conf.OIDC.Issuer,
+			conf.OIDC.ClientID,
+			"",  // no secret, PKCE only
```
jmattheis, line 66: Can't you use a public client for the web ui too? Keep `GOTIFY_OIDC_CLIENTSECRET` unset or set to an empty string. Then we don't need this workaround for Entra.

gekmihesg: True, that works as well. The change took some iterations and had more workarounds added but they all proved to be unnecessary and this was the only part left. So, nevermind then, feel free to close this PR.

PR #1029, opened 2026-08-15
feat: auto-redirect to OIDC IdP (#991)

DerDummePunkt:
This PR attempts to implement the missing part of #991

Add GOTIFY_OIDC_AUTO_REDIRECT (with default `false`) to skip the login page and redirect straight to the configured OIDC provider. Only takes effect when local auth is disabled, since local login would otherwise be unreachable.

Add GOTIFY_OIDC_AUTO_REDIRECT_REQUIRE_REAUTH (default false) to send prompt=login on that redirect, so logging out of Gotify doesn't silently log the user back in via an existing IdP session. Does not end that IdP session, so other apps using it are unaffected.

Wire both flags through gotifyinfo/injected UI config and the WebUI login page, which now redirects instead of showing the OIDC button when enabled.

Added helperfunction to emit warnigns via FutureLog in config builder/parser


gotify-server.env.example
```diff
@@ -224,6 +224,23 @@
...
+# GOTIFY_OIDC_AUTO_REDIRECT=false
+
+# Send prompt=login to the OIDC provider on auto-redirect, so that logging
+# out of Gotify does not silently and invisibly log the user back in via an
+# existing IdP session. This does not end that IdP session, so other
+# applications using it are unaffected. Only takes effect if
+# GOTIFY_OIDC_AUTO_REDIRECT is also in effect.
+#
+# Type: boolean
+# GOTIFY_OIDC_AUTO_REDIRECT_REQUIRE_REAUTH=false
```
jmattheis: Can you rename this to GOTIFY_OIDC_PROMPT with type text and default `login`. This should be used for both login and elevate. And be independent of the auto redirect feature.

gotify-server.env.example
```diff
@@ -224,6 +224,23 @@
 # Type: text-list
 # GOTIFY_OIDC_SCOPES=openid,profile,email
 
+# Automatically redirect to the OIDC identity provider instead of showing the
+# login page. Only takes effect if GOTIFY_OIDC_ENABLED is true and
+# GOTIFY_LOCALAUTH_ENABLED is false, since local login would otherwise be
+# unreachable.
```
jmattheis: This should also have effect when local auth is enabled. Users probably have a default admin user, and everything else is using oidc.

The UI should support disabling the redirect, so admins can still use the local user. E.g. http://localhost:8080/#/login?redirect=false

ui/src/user/Login.tsx
```diff
@@ -21,11 +22,24 @@ const Login = observer(() => {
...
+        'auth/oidc/login?name=' +
+        encodeURIComponent(currentUser.createClientName());
     React.useEffect(() => {
         if (currentUser.loggedIn) {
             navigate('/');
+            return;
         }
-    }, [currentUser.loggedIn]);
+        if (!currentUser.authenticating && oidcAutoRedirect) {
+            window.location.href = oidcLoginUrl;
```
jmattheis, line 40: Is it possible to use `navigate(oidcLoginUrl)` here?

model/gotifyinfo.go
```diff
@@ -29,4 +29,11 @@ type GotifyInfo struct {
 	// required: true
 	// example: OIDC
 	OIDCIDPName string `json:"oidcIdpName"`
+	// If the WebUI should automatically redirect to the OIDC identity
+	// provider instead of showing the login page. Always false while local
+	// authentication is enabled.
+	//
```
jmattheis, line 34:
```suggestion
	// If the WebUI should automatically redirect to the OIDC identity
	// provider instead of showing the login page.
	//
```
No need to expose this in the docs.

config/config.go
```diff
@@ -194,6 +198,20 @@ func Get() (*Configuration, []FutureLog) {
...
+		))
+	}
+	c.OIDC.AutoRedirect = c.OIDC.AutoRedirect && !c.LocalAuthEnabled
+
+	if c.OIDC.RequireReauth && !c.OIDC.AutoRedirect {
+		logs = append(logs, futureWarn(
+			"GOTIFY_OIDC_AUTO_REDIRECT_REQUIRE_REAUTH has no effect unless GOTIFY_OIDC_AUTO_REDIRECT is also in effect",
+		))
+	}
+	c.OIDC.RequireReauth = c.OIDC.RequireReauth && c.OIDC.AutoRedirect
```
jmattheis: I guess these aren't needed anymore, after autoredirect has effect when local auth is enabled. and reauth is migrated to GOTIFY_OIDC_PROMPT and doesn't require auto redirect.

jmattheis (changes requested):
Thanks for the contribution.

ui/src/user/Login.tsx
DerDummePunkt: whoops, thats a leftover from when i experimented wiht some additional url param to enable the brwosers full history and back buttons

gotify-server.env.example
DerDummePunkt: yeah, i was thinking about that when i was trying to figure out how to best make the login UI still reachable for local admin users, but i decided against it eventually.

alright, i'll add it back, its not that big of a change.

ui/src/user/Login.tsx
DerDummePunkt: on second thought, can navigate even do that properly?
isnt that meant for routes within the app?

Believe it or not, so far i always avoided working with react in a professional capacity, so im not that familiar with it.

Edit: ok, so i had a look at what [useNaviagte](https://reactrouter.com/api/hooks/useNavigate#summary) actually does.
as far as i understand it, if we use HashRouter, calling `naviagate(oidcLoginUrl)` would force an update of the hashrouter, effectively creating a route like `#/auth/oidc/login`, which is not what we want, we want the browser to follow the regular redirect flow to iniate the OIDC flow.

so, no, i wouldnt use `navigate` here, i think `window.location` is correct here.
we could use `window.location.assign(oidcLoginUrl)`, but i dont think thats functionally any different from just setting the `window.location.href`

gotify-server.env.example
DerDummePunkt: Makes sense, but if we do that, since we then pass whatever is the value of that env var to the IdP as a url param, I would suggest we also check the value provided against the allowed values according to [the oidc spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)
I really dont want to pass any unvalidated data to a login system

ui/src/user/Login.tsx
jmattheis: Okay, I thought that maybe it would just do normal redirect when an absolute url is provided. It can stay like this.

gotify-server.env.example
jmattheis: I think it's okay to do this unvalidated as this setting set by administrators. If the spec changes we'd have to adjust the validation, so I don't think there is much benefit in validating it.

DerDummePunkt:
Im pretty confident the prompt params are correct and to spec, but I'm having trouble testing value combinations, it seems I have discovered a bug in authelia :man_shrugging: 

I will test mutliple prompt values against one of my dev keycloak instances later, but configuring keycloak is always such a pain in the a**, I doint have time for that right now

gotify-server.env.example
DerDummePunkt: Well, yes, if a protocol you use changes, you have might have to adjust the code.

If i remember correctly, the OIDC spec follows semnatic versioning, that means future minor version bumps shouldnt remove existing params or values, only add to it.

I already added validation as `func parseOIDCPrompt` to `config/parse.go`

```golang
func parseOIDCPrompt(target *string, env string) error {
	raw, ok, err := lookupEnv(env)
	if err != nil {
		return err
	}
	if !ok {
		return nil
	}
	values := strings.Fields(raw)
	hasNone := false
	for _, value := range values {
		if !validOIDCPromptValues[value] {
			return fmt.Errorf(
				"invalid value for %s (%q): must be a space-delimited combination of none, login, consent, select_account",
				env, raw,
			)
		}
		hasNone = hasNone || value == "none"
	}
	if hasNone && len(values) > 1 {
		return fmt.Errorf("invalid value for %s (%q): none must not be combined with other values", env, raw)
	}
	*target = raw
	return nil
}
```

I think its worth doing, it protects gotify admins from faulty configs and protects OIDC IdP system from some injection vectors.

jmattheis: I've decided against including this, as the existing list already missed the `create` prompt type.

jmattheis:
I've made some changes, notably:
* Made the prompt a text-list
* Added the prompt parameter to the external auth flow
* The login form is always shown (no login spinner), and no redirect is done when oidc is disabled or there is a connection error

I'll test this some more this week.

jmattheis:
I've tested with dex, authelia, keycloak and authentik. Authentik didn't seem to use the prompt parameter, but didn't error when it was set. I tried using multiple prompt values like `login consent`, but this didn't seem to make any difference on the listed IdPs. authelia prompt for consent anyway when using prompt=login.

jmattheis (approved):

PR #1030, opened 2026-08-16
chore: update go module path to github.com/gotify/server/v3

eternal-flame-AD:
Updates go import paths to V3.

Sorry I missed this in the review. We need to update the go module path for any part of the V3 tree to be visible to the go toolchain. We would also have to bump the version once more to make plugins buildable again against v3.


api/stream/client.go
```diff
@@ -4,7 +4,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
-	"github.com/gotify/server/v2/model"
+	"github.com/gotify/server/v3/model"
```
jmattheis, line 7: The plugin builds don't depend on gotify/server, why does the wrong module path block the plugin builds?

---

The build is failing because of formatting, otherwise this LGTM.

eternal-flame-AD: Because it needs a phantom dependency to gotify/server/v3 to ensure it gets built with compatible dependencies. If the import path is wrong it wont be able to reference it.

jmattheis (approved):
LGTM. I'll create a release later today or tomorrow (after merging #1029).

PR #1033, opened 2026-08-20
feat: add support for mapping user admin status from OIDC claims

### LINKED ISSUE #957: OIDC: Map group claims to admin user flag
Allow configuring group claims that map to the user admin status. From https://github.com/gotify/server/issues/433#issuecomment-4322820716

UiP9AV6Y:
an optional new configuration setting allows operators to define a JMESPath expression to calculate the admin status of a user based on their OIDC claims. this feature can also be used to deny users access alltogether, by mapping their claims to a special role value.

closes #957


api/oidc.go
```diff
@@ -57,11 +58,20 @@ func NewOIDC(conf *config.Configuration, db *database.GormDatabase, userChangeNo
 		log.Fatal().Err(err).Msg("failed to initialize OIDC provider")
 	}
 
+	var roleExpression jmespath.JMESPath
+	if conf.OIDC.RoleExpression != "" {
+		roleExpression, err = jmespath.Compile(conf.OIDC.RoleExpression)
```
jmattheis: I don't think we need a full expression language here. Instead I propose three settings like this:
```
# OIDC ID-token claim containing the user's group memberships as a list of
# strings.
#
# Type: text
# GOTIFY_OIDC_GROUPS_CLAIM=groups

# Groups allowed to log in via OIDC. When set, users whose groups claim
# matches neither USER_GROUPS nor ADMIN_GROUPS are rejected. When empty,
# every user authenticated by the identity provider may log in.
#
# Type: text-list
# Example: messenger,gotify
# GOTIFY_OIDC_USER_GROUPS=

# OIDC groups whose members are granted admin rights.
# Membership is evaluated after each login.
#
# Type: text-list
# Example: admins,gotify-admins
# GOTIFY_OIDC_ADMIN_GROUPS=
```

UiP9AV6Y: > I don't think we need a full expression language here.

does that mean you are waiting for a second opinion or is that decision final? i am fine with the proposed changes, but wanted to clarify before i implement them.

UiP9AV6Y: > GOTIFY_OIDC_USER_GROUPS
> GOTIFY_OIDC_ADMIN_GROUPS

shouldn't that be `GOTIFY_OIDC_GROUPS_USER_ROLE` and `GOTIFY_OIDC_GROUPS_ADMIN_ROLE` to follow the nested hierarchy nature of the config keys?

in v2 this would have looked like this:

```yaml
oidc:
  groups:
    claim: groups
    user_role: [messenger, gotify]
    admin_role: [admins, gotify-admins]
```

jmattheis: > does that mean you are waiting for a second opinion or is that decision final? i am fine with the proposed changes, but wanted to clarify before i implement them.

It's not final, you can tell your opinion, and give good points why a expression language is needed here. But if the three-setting solution is good enough that I'd prefer this.

> shouldn't that be GOTIFY_OIDC_GROUPS_USER_ROLE and GOTIFY_OIDC_GROUPS_ADMIN_ROLE to follow the nested hierarchy nature of the config keys?

Good point, but I dislike the group and role in one name, as it basically describe the same thing. How about
```
GOTIFY_OIDC_GROUPS_CLAIM
GOTIFY_OIDC_GROUPS_USER
GOTIFY_OIDC_GROUPS_ADMIN
```
?

UiP9AV6Y: i have no strong opinion about using an expression languange. i have simply chosen this approach, as Grafana does the same. my personal deployment would also work with the simplified approach. this approach is also what other applications (e.g. ArgoCD) use for their mapping.

i have pushed a new commit with the refactored system. test and documentation have also been updated.

api/rbac.go
```diff
@@ -0,0 +1,104 @@
...
+// with the list will yield [RoleUser], otherwise [RoleNone] is returned.
+func (r *SimpleRBACResolver) ResolveRole(subject any) (model.Role, error) {
+	if subject == nil {
+		return r.resolveSlice(nil), nil
+	}
+
+	switch roles := subject.(type) {
+	case string:
+		return r.resolveSlice([]string{roles}), nil
+	case []string:
```
jmattheis: This doesn't work. the groups are provided here as `[]any`. (I've already adjusted this).

api/rbac.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"fmt"
+	"slices"
+
+	"github.com/gotify/server/v3/model"
+)
+
+// RBACResolver allows for arbitrary input to be mapped to roles for
+// use in a permission system.
+type RBACResolver interface {
```
jmattheis: I feel like this is pretty overengineered. All this matching and conversion should be done in a single function. This makes this IMO easier to understand. (I've already adjusted this)

jmattheis:
@eternal-flame-AD Could you take a look at these changes? I've made some larger adjustments and would like a second pair of eyes (:

---

I've tested this with authentik and Authelia with different combinations of settings and groups.

eternal-flame-AD:
sure, will do tonight :)

api/oidc.go
```diff
@@ -436,11 +444,25 @@ func (a *OIDCAPI) resolveUser(issuer string, info *oidc.UserInfo) (*model.User,
...
 	}
+
+	admin, status, err := a.resolvePermission(idToken.Claims, info.Claims)
+	if err != nil {
+		log.Err(err).Str("oidc_id", oidcID).Interface("idTokenClaims", idToken.Claims).Interface("userinfoClaims", info.Claims).Msg("OIDC: resolve permission")
+		return nil, status, err
+	}
+
 	if user != nil {
+		if admin != nil && user.Admin != *admin {
```
eternal-flame-AD: The status of being _not_ an admin is simply the lack of the admin role in the group claim, there isn't really a third "empty" state here IMO. Every case where there is not an admin role present is just not an admin.

Why not just make admin a boolean instead of tristate , and update the admin flag if GOTIFY_OIDC_GROUPS_ADMIN is defined. Ideally also check this configuration and prohibit manual admin changes in UpdateUserById()

The main risk I am worried about is accidentally omitting a code path where the admin status isn't stripped when it should be. I don't think there is really a use case where you might have this situation where you have some users have the GroupClaim but other don't.

gotify-server.env.example
```diff
@@ -224,6 +224,30 @@
 # Example: email
 # GOTIFY_OIDC_USERNAMECLAIM=preferred_username
 
+# OIDC ID-token claim containing the group memberships as a list of strings.
+# The claim must be part of the requested scopes (GOTIFY_OIDC_SCOPES). When
```
eternal-flame-AD, line 228: Should we check this or maybe just silently add it if it's not present?

api/oidc.go
jmattheis: > Why not just make admin a boolean instead of tristate , and update the admin flag if GOTIFY_OIDC_GROUPS_ADMIN is defined.

Okay

> Ideally also check this configuration and prohibit manual admin changes in UpdateUserById()

I think I skip this for now, as we still allow logging in with user/pass for oidc accounts. I create a new ticket for this, as this is likely not that straightforward. => https://github.com/gotify/server/issues/1037

gotify-server.env.example
jmattheis: I'm not sure if we can. Afaik, this isn't standardized (https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.5.4) , so the claim-key could be something like "http://example.info/claims/groups" and the scope maybe too. Most IdPs use `groups`, but I rather not force this and instead document this per IdP in our OIDC documentation.

api/oidc.go
UiP9AV6Y: there is actually a third state: the case where the user is not allowed to login at all b/c they are not a member of any USER or ADMIN group. that is why i initially implemented an enum to cover all cases in a descriptive manner.

gotify-server.env.example
UiP9AV6Y: as @jmattheis said already, you cannot infer the correct scope configuration from the used claims.

for example: gitlab offers [multiple](https://docs.gitlab.com/integration/openid_connect_provider/#shared-information) group-like claims, all of which are provided as part of the **profile** scope.

api/oidc.go
eternal-flame-AD: > there is actually a third state

True, but it's entirely handled in a self contained switch statement around L581 in resolvePermission. The code did not use the nil value directly to decide the user should not be allowed.

The return of the tri-state symbol is propagated in a way that fallbacks to a no-op  (L455 of oidc.go) creating a semantics of "whatever, leave the admin state alone" instead of "I don't know this user at all." Should the independent check be removed or bypassed by new callers in the future this could be a silent failure mode.

eternal-flame-AD (approved):
LGTM!

PR #1042, opened 2026-09-02
enhance: serialize user update methods

eternal-flame-AD:
Serializes user update actions to prevent race conditions leading to unexpected results.

I removed the 'Test_UpdateUserByID_EmptyPassword_Expect400' test as it seemed to be a mistake - it should return 200, it returned 400 in the test because there wasn't a second admin.

### DIFF
--- a/api/user.go
+++ b/api/user.go
@@ -9,19 +9,11 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/gotify/server/v3/auth"
 	"github.com/gotify/server/v3/auth/password"
+	"github.com/gotify/server/v3/database"
 	"github.com/gotify/server/v3/model"
 )
 
-// The UserDatabase interface for encapsulating database access.
-type UserDatabase interface {
-	GetUsers() ([]*model.User, error)
-	GetUserByID(id uint) (*model.User, error)
-	GetUserByName(name string) (*model.User, error)
-	DeleteUserByID(id uint) error
-	UpdateUser(user *model.User) error
-	CreateUser(user *model.User) error
-	CountUser(condition ...any) (int64, error)
-}
+var errCannotDeleteLastAdmin = errors.New("cannot delete last admin")
 
 // UserChangeNotifier notifies listeners for user changes.
 type UserChangeNotifier struct {
@@ -59,7 +51,7 @@ func (c *UserChangeNotifier) fireUserAdded(uid uint) error {
 
 // The UserAPI provides handlers for managing users.
 type UserAPI struct {
-	DB                 UserDatabase
+	DB                 *database.GormDatabase
 	PasswordStrength   int
 	UserChangeNotifier *UserChangeNotifier
 	Registration       bool
@@ -343,19 +335,26 @@ func (a *UserAPI) DeleteUserByID(ctx *gin.Context) {
 			return
 		}
 		if user != nil {
-			adminCount, err := a.DB.CountUser(&model.User{Admin: true})
-			if success := successOrAbort(ctx, 500, err); !success {
-				return
+			for range 3 {
+				err = a.DB.Txn(func(txdb *database.GormDatabase) error {
+					if err := txdb.DeleteUserByID(id); err != nil {
+						return err
+					}
+					anotherAdmin, err := txdb.GetUsers(&model.User{Admin: true})
+					if err != nil {
+						return err
+					}
+					if user.Admin && len(anotherAdmin) == 0 {
+						ctx.AbortWithError(400, errCannotDeleteLastAdmin)
+						return errCannotDeleteLastAdmin
+					}
+					return a.UserChangeNotifier.fireUserDeleted(id)
+				})
+				if err == nil || ctx.IsAborted() {
+					return
+				}
 			}
-			if user.Admin && adminCount == 1 {
-				ctx.AbortWithError(400, errors.New("cannot delete last admin"))
-				return
-			}
-			if err := a.UserChangeNotifier.fireUserDeleted(id); err != nil {
-				ctx.AbortWithError(500, err)
-				return
-			}
-			successOrAbort(ctx, 500, a.DB.DeleteUserByID(id))
+			successOrAbort(ctx, 500, err)
 		} else {
 			ctx.AbortWithError(404, errors.New("user does not exist"))
 		}
@@ -470,15 +469,7 @@ func (a *UserAPI) UpdateUserByID(ctx *gin.Context) {
 				return
 			}
 			if dbUser != nil {
-				adminCount, err := a.DB.CountUser(&model.User{Admin: true})
-				if success := successOrAbort(ctx, 500, err); !success {
-					return
-				}
-				if !updatedUser.Admin && dbUser.Admin && adminCount == 1 {
-					ctx.AbortWithError(400, errors.New("cannot delete last admin"))
-					return
-				}
-
+				dbUserWasAdmin := dbUser.Admin
 				dbUser.Name = updatedUser.Name
 				dbUser.Admin = updatedUser.Admin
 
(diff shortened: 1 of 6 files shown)


api/user.go
```diff
@@ -58,8 +60,8 @@ func (c *UserChangeNotifier) fireUserAdded(uid uint) error {
 }
 
 // The UserAPI provides handlers for managing users.
-type UserAPI struct {
-	DB                 UserDatabase
+type UserAPI[T UserDatabase[T]] struct {
+	DB                 T
```
jmattheis, line 64: Can we use database.GormDatabase directly here? I don't think we need the interface here. I think I previously added this when the DB was mocked, but this isn't done anymore.

api/user.go
```diff
@@ -336,26 +338,41 @@ func (a *UserAPI) GetUserByID(ctx *gin.Context) {
...
-			}
-			if err := a.UserChangeNotifier.fireUserDeleted(id); err != nil {
-				ctx.AbortWithError(500, err)
-				return
+			for range 3 {
+				commitError := false
+				err = a.DB.Txn(func(txdb T) error {
+					if success := successOrAbort(ctx, 500, txdb.DeleteUserByID(id)); !success {
+						return err
+					}
```
jmattheis, line 353: After the transaction commits/aborts there is another ctx.AbortWithError(500, err). successOrAbort already Aborts the ctx. The context should only be aborted once, as I think otherwise multiple errors are printed to the body.

api/user.go
```diff
@@ -336,26 +338,41 @@ func (a *UserAPI) GetUserByID(ctx *gin.Context) {
...
+					if success := successOrAbort(ctx, 500, a.UserChangeNotifier.fireUserDeleted(id)); !success {
+						return err
+					}
+					commitError = true
+					return nil
+				})
+				if !commitError || err == nil {
+					break
+				}
```
jmattheis, line 370: It took me a while to understand this case, maybe it could be done like this with explicit definition what the boolean is used for?
```go
func (a *UserAPI[T]) DeleteUserByID(ctx *gin.Context) {
	withID(ctx, "id", func(id uint) {
		user, err := a.DB.GetUserByID(id)
		if success := successOrAbort(ctx, 500, err); !success {
			return
		}
		if user == nil {
			ctx.AbortWithError(404, errors.New("user does not exist"))
			return
		}

		for range 3 {
			retryable := true
			err = a.DB.Txn(func(txdb T) error {
				if err := txdb.DeleteUserByID(id); err != nil {
					return err
				}
				anotherAdmin, err := txdb.GetUsers(&model.User{Admin: true})
				if err != nil {
					return err
				}
				if user.Admin && len(anotherAdmin) == 0 {
					retryable = false
					return errCannotDeleteLastAdmin
				}
				if err := a.UserChangeNotifier.fireUserDeleted(id); err != nil {
					retryable = false
					return err
				}
				return nil
			})
			if err == nil {
				// user deleted successfully
				ctx.Status(200)
				return
			}
			if retryable {
				continue
			}
			if err != nil {
				status := 500
				if errors.Is(err, errCannotDeleteLastAdmin) {
					status = 400
				}
				ctx.AbortWithError(status, err)
				return
			}
		}
	})
}
```

jmattheis (changes requested):

api/user.go
eternal-flame-AD: I think only commit errors should be retryable (temporary serialization failures). If regular statements returned errors it means there is something wrong with the database connection or underlying data, we should just return immediately.

jmattheis: Okay, then only retry then. Can it use the adjusted statements so than err == null is a separate statement and commitError == true does an explicit continue? (I think I still prefer calling it retryable) I had trouble understanding the retry condition.

api/user.go
eternal-flame-AD: I didn't get what you meant, isn't user_test.go using the testdb wrapper and therefore this has to be generic?

api/user.go
eternal-flame-AD: I replaced with a much simpler logic. I think this is better and doesn't assume the serialization error must happen on COMMIT

api/user.go
jmattheis: You can get the underlying GormDatabase on the testdb.Database via .GormDatabase. So no generic needed.

api/user.go
```diff
@@ -494,10 +496,30 @@ func (a *UserAPI) UpdateUserByID(ctx *gin.Context) {
...
+						}
+
+						return nil
+					})
+
+					if err == nil || ctx.IsAborted() {
+						return
+					}
 				}
-				ctx.JSON(200, toExternalUser(dbUser))
```
jmattheis, line 500: This changes the behavior of the api, previously, the updated record is returned, now it's 200 without any body:
```
# OLD
$ http -a admin:admin POST :8080/user/11 name==update
HTTP/1.1 200 OK
Content-Length: 89
Date: Sun, 06 Sep 2026 10:14:23 GMT

{
    "admin": false,
    "createdAt": "2026-09-06T12:13:42.095298193+02:00",
    "id": 11,
    "name": "update"
}

# NEW
http -a admin:admin POST :8080/user/11 name==update-branch
HTTP/1.1 200 OK
Content-Length: 0
Date: Sun, 06 Sep 2026 10:14:36 GMT
```

Probably needs something like this:
```
				if success := successOrAbort(ctx, 500, err); !success {
					return
				}
				ctx.JSON(200, toExternalUser(dbUser))
```

api/user.go
```diff
@@ -59,7 +62,7 @@ func (c *UserChangeNotifier) fireUserAdded(uid uint) error {
 
 // The UserAPI provides handlers for managing users.
 type UserAPI struct {
-	DB                 UserDatabase
+	DB                 database.GormDatabase
```
jmattheis: Can you change this to a pointer, and remove the UserDatabase interface?

jmattheis (changes requested):

api/user.go
eternal-flame-AD: I'm not super sure about this and I'm think we should back off to the original solution (make UserAPI a generic). Should we stop using interfaces for DB abstractions? It seems like a much bigger change than just changing this one

jmattheis: I'm okay with changing this only for the UserAPI in this PR, and change the rest, when there would be the need to adjust the interface. E.g. the oidc handler already directly uses the GormDatabase.

api/user.go
jmattheis: Seems like my suggestion didn't fix it because of the early return if err == nil || ctx.Aborted(). Should probably only return when it's aborted, and break when "err == nil".

api/user.go
```diff
@@ -343,19 +346,26 @@ func (a *UserAPI) DeleteUserByID(ctx *gin.Context) {
...
+					}
+					anotherAdmin, err := txdb.GetUsers(&model.User{Admin: true})
+					if err != nil {
+						return err
+					}
+					if user.Admin && len(anotherAdmin) == 0 {
+						ctx.AbortWithError(400, errCannotDeleteLastAdmin)
+						return errCannotDeleteLastAdmin
+					}
+					return a.UserChangeNotifier.fireUserDeleted(id)
```
jmattheis, line 362: This should probably be moved outside of the transaction and be only invoked when the deleted succeeded. Otherwise the transaction could fail, and we've already deleted all plugin related resources.

eternal-flame-AD: Hmmm, this looks about right to me? If we failed to clean up all resources the transaction should be rolled back so they can try again.

There is the possibility of a partial cleanup I guess but I think it's more intuitive to keep the user record itself intact instead of completely deleting the user record despite leaving residual dependencies not deleted.

jmattheis: Sorry missed the notification; I mean it more the other way around, the deletion from the plugin manager succeeds, but the committing of the transaction fails.

Then the users isn't deleted, but the plugin state is already cleaned up.

PR #1046, opened 2026-09-08
fix: don't panic when a Let's Encrypt request fails

NotAFlightRisk:
A refused connection to the ACME server can take the whole process down, not just the renewal that hit it.

`LoggingRoundTripper.RoundTrip` reads `resp.StatusCode` before it looks at `err`. A dial failure, a DNS miss, a cancelled context or the clients own 60s timeout all come back as a nil response with an error, so that first line panics and the `else if err != nil` branch below it never gets its turn.

Point an `acme.Client` at a directory URL that refuses the connection, through that transport, and you get:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10]

github.com/gotify/server/v3/runner.(*LoggingRoundTripper).RoundTrip
	runner/runner.go:111
```

With the branches swapped it logs the line that branch was written for and hands the error back:

```
{"level":"warn","client":"Let's Encrypt","error":"dial tcp 127.0.0.1:1: connect: connection refused","method":"GET","url":"https://127.0.0.1:1/directory","message":"Request failed"}
```

Which of those you get depends on who made the request. During a handshake `net/http` recovers it, so you lose that connection and a stack trace lands in the log. Renewal is the bad one: autocert schedules it with `time.AfterFunc`, so the callback is a bare goroutine with nothing to recover it, and the server exits.

The diff just moves the error branch to the front, all three log lines are unchaged. There was no test file under `runner/` at all, so theres one now covering the nil-response case.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -108,7 +108,14 @@ type LoggingRoundTripper struct {
 
 func (l *LoggingRoundTripper) RoundTrip(r *http.Request) (resp *http.Response, err error) {
 	resp, err = l.RoundTripper.RoundTrip(r)
-	if resp.StatusCode == 429 {
+	if err != nil {
+		log.Warn().
+			Str("client", l.Name).
+			Err(err).
+			Str("method", r.Method).
+			Str("url", r.URL.String()).
+			Msg("Request failed")
+	} else if resp.StatusCode == 429 {
 		log.Warn().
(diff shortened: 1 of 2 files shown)


jmattheis (approved):
Thanks!

